PR #477, opened 2015-01-16
add length property for function objects

AlexanderBrevig:
Currently this does not work:

```js
var fx = function(a,b,c){};
console.log(fx.length);
```

It seems to be standard since [JavaScript 1.1](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/length).

This path adds support for function.length with the result of
```js
console.log(function(){}.length);
console.log(function(a){}.length);
console.log(function(a,b){}.length);
//prints 
//0
//1
//2
```

### DIFF
--- a/src/jsvar.c
+++ b/src/jsvar.c
@@ -1497,6 +1497,27 @@ JsVar *jsvArrayBufferGetFromName(JsVar *name) {
   return value;
 }
 
+
+JsVar *jsvGetFunctionArgumentLength(JsVar *functionScope) {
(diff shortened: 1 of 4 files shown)


gfwilliams:
That look perfect. Thanks! I'll just tweak the comments on jsvGetFunctionArgumentLength though :)

AlexanderBrevig:
I wrote a test for it and it seemed to have a mem leak. I just sent you an additional PR with test and fix.

PR #478, opened 2015-01-16
and make tests pass...

AlexanderBrevig:
I forgot to add a test and the test revealed a mem leak. Should be fixed now?


gfwilliams:
Thanks! Should have spotted that :)

AlexanderBrevig:
And I should have a made a test before submitting the first PR ;) Starting to get my head around the code now. It's good fun!

PR #483, opened 2015-01-26
Mapleret6

jalawson:
A few cleanups in the MAPLERET6 definition file.
Comment out SDCARD pin defs since this board does not have an SDCARD built in.
Changed MAPLERET6 STLIB entry in Makefile from STM32F10X_XL to STM32F10X_HD


gfwilliams:
Thanks!

PR #487, opened 2015-01-29
Fixes E.variance(arr, mean) by implementing the final averaging (divide ...

jalawson:
...by N-1).


jalawson:
I thought of adding a E.mean(arr) function, incorporating that in E.variance(arr, mean) and 
changing the second parameter of E.variance to select between a sample (n-1) or population (n)
final average.
Thoughts?

gfwilliams:
I think the variance should actually be `/ arraySize` not `/ (arraySize-1)`?

I don't think this should change though - I know there are people using this as-is, and there are some good reasons why you might just want to sum the squares and not divide. The documentation (in the header) explains exactly what it does too: http://www.espruino.com/Reference#l_E_variance

I guess it might be worth making it even clearer though - and perhaps even changing the name to something more descriptive?

jalawson:
The / arraySize vs / (arraySize - 1) choice depends (statistically at least) on whether you are working with the entire population or just a sample of the population of interest.

Either way I get your point. That particular construct (sums of squared differences) does show up all over the place.

Perhaps a module is a more useful implementation. Thanks.

PR #498, opened 2015-03-02
add vagrant setup + documentation

boneskull:
In the current `README.md`, it's recommended to install Virtual Box and create an Ubuntu VM, etc., if you need to build on a non-Linux system.

The same thing can be accomplished in fewer steps with [Vagrant](http://vagrantup.com).  

In addition, this gives the project the ability to make changes to the provisioning of the virtual machine as necessary.  For example, if another compiler needed to be chosen, or perhaps Ubuntu wasn't cutting it anymore--just modify `Vagrantfile` and `scripts/provision.sh`, and that's that.  You could also go as far as *recommending* that the system be built from within a Vagrant machine, since it's a controlled environment--and you control it.

Anyway, I've built the firmware using this method.  Attaching the Espruino to the Vagrant box can probably be accomplished via the Vagrantfile as well (for flashing), but I haven't tried it--otherwise, you can simply load the VM in the VirtualBox GUI and do it as documented.

If I have any suggestions from here, looking for a smaller box on [Atlas](http://atlas.hashicorp.com) may be a good idea, because it will reduce the download size.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,19 @@ To cross compile,
 
 This will create a file called ```functions.html```
 
+### Building under Windows/MacOS with a VM (Vagrant)
+
+* Clone this repository.
+* Download and install the correct [VirtualBox](https://www.virtualbox.org/) for your platform. eg. If you have Windows, download 'VirtualBox for Windows Hosts'.
+* Download and install the correct [Vagrant](https://www.vagrantup.com/downloads.html) for your platform.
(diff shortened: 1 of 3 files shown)


gfwilliams:
Looks great - thanks! Seems a lot easier than other options.

Just quickly, please could you just add a comment to the top of `Vagrantfile` to explain what it is? If people see it there and aren't willing to grep then they may wonder why it exists :)

PR #500, opened 2015-03-04
add documentation for http.request's options.headers field

AlexanderBrevig:
add documentation for headers field for option object used by http.request


gfwilliams:
Thanks!

PR #508, opened 2015-03-18
Update stm32f401_af.csv

the1laz:
Removed SPI4 as it was removed from the pins file and was causing issues compiling.


gfwilliams:
Are you talking about compiling for the STM32F401 here?

Strange... Did SPI4 actually get removed or was it never in there? I would have thought it'd be better to add support for SPI4 than to just disable it?

the1laz:
Yes,  the 401. I had to remove SPI4 to compile for the F401 discovery. It looks like it was removed from stm32f401.csv in commit f70cd978bcd588a95be9d46c5eb35f52546f3ff7, but I'm not sure why.
I think the discovery board is the only F401 that has SPI4 as I don't think the others are in the 100 pin packages. I suppose it would be better to get SPI4 working, but I thought that for now at least people would be able to compile it.

gfwilliams:
Ahh - thanks. Yes, I guess it'll be because of the bigger package. I guess I accidentally re-committed the SPI4 when I came up with a new af file that had the AFs in the right places.

I'll merge this in and file a bug for adding SPI4 support.

Thanks!

the1laz:
I'll try and have a look at SPI4 once I get up and running with my discovery board. Thanks for including these other boards/chips in espruino!

PR #558, opened 2015-08-06
Update documentation

bartmichu:
Add a note about `Pin.read` and `digitalRead` resetting pin's state.


gfwilliams:
Great - thanks! I'll copy/paste that into digitalWrite/Read/others as well

PR #560, opened 2015-08-11
Porting Espruino to the nRF52832 Preview Development Kit

mjdietzx:
The nRF52832 is a BLE enabled arm-cortex M4f SoC with near field communication capabilities. I only implemented the bare bone functions i.e (non blocking uart to talk to terminal, blinking LEDs). I'm currently working on sending javascript commands to the Espruino interpreter over BLE instead of USB. (My goal is to program these chips from a smartphone over Nordic's BLE uart service). Also want to add more functionality in the future like timers, spi, etc...

Compiles and runs on nRF52 Preview DK. Shouldnt have any affect on other targets. I tested for STM32F4DISCOVERY and that is still working as expected. I included the entire Nordic SDK so it is easier for developers to build on features.

-Mike


gfwilliams:
Wow, that's awesome! I just pushed this : https://github.com/espruino/Espruino/commit/235f881c91945c6bdd657c83fb905dfb4bc5d88b

I'm also trying to support the BBC micro:bit, which is NRF51822 (although I've been looking at doing it via mbed) - and that's what killed the 'load from flash' bit in jsinteractive.c.

Before I merge, do you think you could:

* try removing your changes to `jsinteractive.c` and see if it still works? Should be tidier.
* revert `build_linker.py`, and instead just set `LINKER_FILE = targetlibs/nrf52/linker.ld` in the `Makefile`? I think it'd be a bit tidier as you're not actually custom-generating any link script? (I guess I should probably make `build_linker.py` STM32-only)
* just change the `foo` bit at the end of `Makefile` to `debuginfo` or something?

I don't mind doing that, but at least if you do it you can check it hasn't broken anything.

Having NFC too sounds really cool - if *that* could be used for sending JS code it'd be awesome... I've got a bunch of hardware with NFC but no BLE - also requiring the proximity goes some way to helping with security issues ;)

gfwilliams:
Just to add - while I've been doing stuff with mbed, it looks like nordic have kept their APIs very similar between nrf51 and nrf52 - and it's definitely more efficient than the mbed libs. It feels like your work could be quite easily ported to nrf51 as well.

mjdietzx:
yup. ill get those changes done. and Ill look into porting it to the nrf51 -shouldnt have to do to much there.

mjdietzx:
OK, i added the changes you wanted. however my code breaks if i remove my code from jsinteractive.c. Right now I've been working on sending the javascript commands from my smartphone over BLE instead of from the espruino chrome ide. Im hoping to make a pull request for that soon as well. This is done with nordics BLE UART service and using their UART app in nRF toolbox. Working pretty good so far until I try to write over 20 bytes but hoping to fix that soon!

Also will port to nRF51 once i get this BLE done but I have some other projects coming up so not sure when ill get around to it.

-Mike

gfwilliams:
Thanks! Looks like I actually misread where your change was in `jsinteractive.c`.

What usually happens is for the first second, all console output goes to EV_LIMBO, then about 1 sec after that via jsiOneSecAfterStartup (when USB has had a chance to pick up), it switches to either USB or Serial, depending on whether USB has worked. 

In Espruino, jsiOneSecAfterStartup is called in an IRQ, so if a print statement has blocked because output buffer got full while writing to EV_LIMBO it'll resume... But if you were calling jsiOneSecAfterStartup in jshIdle (like I think I probably did in the Linux version) then that won't happen.

... but the NRF52 doesn't have USB? I should probably just `ifdef USB` all the LIMBO stuff in `jsinteractive.c` and I think it'd be fine.

gfwilliams:
Maybe check out what I just pushed? Hopefully that'll still work for you, but it should be a bit more flexible.

I think as it was, even something like:

```
function onInit() {
  for (var i=0;i<100;i++) console.log("Hello World");
}
```

would have caused it to lock up.

PR #563, opened 2015-08-19
Mapleret6

jalawson:
Fixes the pin names by adding MAPLERET6 to the Olimexino pin renumbering condition in
build_pininfo.py.
Adds more features to the relevant Makefile section since the RET6 has enough memory.


gfwilliams:
That's great, thanks! If there's enough space, you might find that `USE_FILESYSTEM=1` in the Makefile is handy (even if there's no SD card, you can wire one up to *any* pins and use it).

jalawson:
Thanks Gordon. 
I wasn't sure if USE_FILESYSTEM=1 should be the default if the board does not normally come with an SD interface. 
Personally I do include it. 

----- Original Message -----

PR #568, opened 2015-08-27
Change print "XXX" to print("XXX")

nkolban:
Change the python scripts to add parenthesis to print statements.


gfwilliams:
Awesome. thanks!

PR #570, opened 2015-08-29
Update Vagrantfile

nihaopaul:
added require 'vagrant-auto_network' to vagrantfile as it was throwing an error due to not having vagrant-auto_network installed..

`vagrant plugin install vagrant-auto_network`

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,3 +1,5 @@
+require 'vagrant-auto_network'
+
 Vagrant.configure(2) do |config|
 
   # Ubuntu Server


gfwilliams:
Thanks!

luciodj:
I have been trying to build on a Mac (Yosemite) with vagrant 1.7.4 (obtained via brew cask.. as suggested)  and this very line (`require vagrant-auto_network ...`) was rejected as a "likely syntax error in line 1 of Vagrantfile"!?.
I had to google it a bit and found out that the `vagrant-auto_network` was a plug in that required separate installation, so I ended up installing the plug in manually anyway... sort of defeating the purpose I suppose....

`vagrant plugin install vagrant-auto_network`

All worked after that...

gfwilliams:
Ok, so if this line wasn't in there, but you had the plugin installed, then it would have just worked anyway?

luciodj:
I suppose so, I cannot prove it now as I did eventually install the plugin...
Perhaps a note (adding a step in the README_Building.md ) to alert users to install the extra plugin would do a better service... 
or ( I am not familiar with homebrew enough to say) perhaps there is a way to get the plugin with vagrant at the same time

luciodj:
Ok, I did check. After uninstalling it all, I brewed vagrant from scratch but this time added the plugin right away. No error message was produced.

gfwilliams:
Wow, thanks! So without the `vagrant plugin install vagrant-auto_network` line it's fine as well? So I could just revert this and stick something in the README?

luciodj:
Actually I would keep the require... after all it helped to point me in the right direction rather than finding other more indirect errors later.
I would definitely add the plugin install instructions (single liner)  in the README!

gfwilliams:
Ok, so literally just `vagrant plugin install vagrant-auto_network`?

I use Linux so haven't ever used vagrant.

luciodj:
yes, that did it for me!
BTW my first time with vagrant, you forced me to look into something potentially very powerful/useful, THANKS

PR #575, opened 2015-09-21
Definitions for __WORDSIZE not guarded

nkolban:
Found some additional definitions for __WORDSIZE that weren't guarded.


gfwilliams:
Thanks - not sure how it's happened, but your editor has changed the file type - so it's converted every line from Linux formatting to Windows formatting. I'll close this pull request and change those 2 lines manually

gfwilliams:
Ok, just done!

nkolban:
Thank you Gordon.  I found the problem with LF vs CRLF ... that should not happen again.  Many thanks for being patient with me.  I hope not to make the same mistakes twice.

PR #576, opened 2015-09-22
nrf52 functionality, libraries: nrf51 platform support - working

mjdietzx:
nRF52:
-added more functionality. 
-cleaned up code
-added temperature library (mainly for test) and almost working bluetooth library (sending strings from terminal to phone get messed up because jswrap wont let me use char as input type).

nRF51:
-added support for nrf51. seems to work fine. flash is too small - no room for softdevice. microbit is in the makefile but i didn't make a boards.py or anything else for it.

todo:
- clean up code more, figure out a way not to duplicate anything between nRF5x platforms. add nrf52 functionality and libraries.figure out flash issues for nrf51 and how to get ble running.


gfwilliams:
Awesome - thanks! By the way, there's already `E.getTemperature()` that it'd be worth using eventually :)

mjdietzx:
hey np. ok, im going to remove that class. i added it for testing out the libs and have been using it as an example to show people so ill remove it once i have some other better libraries :)

mjdietzx:
you can actually remove it from your branch if you want.

PR #578, opened 2015-09-27
Commit of ESP8266 board specific source files.

nkolban:
This is a pull request for inclusion of ESP8266 board specific source files.  This pull request does NOT influence the build systems or other non-ESP8266 specific areas.  As such the risk is extremely low that the merging of this pull request will regress the existing Espruino project.  The end result of a merge here will be NEW files with ZERO updates to existing Espruino files.


gfwilliams:
That's awesome - thanks!

PR #580, opened 2015-09-27
Build system changes for the ESP8266 board.

nkolban:
This is the second Pull Request for the ESP8266 board integration with the Espruino primary code base.  This one includes changes that are needed for a build environment.  I consider this change the highest risk possible as it changes the Makefile which is core and key to all Espruino builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,7 @@
 # MINISTM32_STRIVE=1
 # MINISTM32_ANGLED_VE=1
 # MINISTM32_ANGLED_VG=1
+# ESP8266_BOARD=1                  # ESP8266 from Espressif
 # Or nothing for standard linux compile
 #
 # Also:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -777,6 +795,14 @@ libs/network/js/network_js.c
  libs/network/wiznet/Ethernet/socket.c \
  libs/network/wiznet/W5500/w5500.c
  endif
+
```
gfwilliams: This might be left over from the other ESP8266 code that I pulled out in a commit a few days ago? Merging it anyway, but could be worth checking?

gfwilliams:
Thanks! That's great!

Just to add, if you see something called 'Travis', it's a continuous integration tool - if it detects that something that's committed won't build, it'll flag it up.

... this builds fine though :)

PR #583, opened 2015-09-28
fix esp8266 to compile with gcc4 on linux

tve:
The makefile got a bit of an overhaul to bring the esp8266 stuff more in line with other architectures and there was some major fixing necessary in order to make link time optimizations with inlining work. One part was to remove -std=gnu99 as compiler option, which required to change some variable definitions.
It would be good to wait for nkolban to verify that this still builds on windows with gcc5 before merging


tve:
With the additional changes it looks like Kolban is successful in compiling. I had to turn link-time-optimizations off 'cause gcc5.1/windows crashes with an internal compiler error. I'm continuing to push ahead on linux using the new ESP8266_4MB target.

gfwilliams:
Thanks, this looks good!

The `$(PROJ_NAME).elf` stuff looks to be basically the same for both targets at the moment - is there any chance it could be kept as one?

tve:
the non-512KB target is work in progress, it will be significantly different when done. I'm almost there, but it's time for bed and work tomorrow. Kolban is unblocked as is anyone else who want to play with the esp8266, so that's good and the next PR will be better finished. Thanks for humoring us...

PR #587, opened 2015-09-28
Changes to support clean compile for os_printf

nkolban:
The ESP8266 os_printf function requires that any data of format "%d" must be of type "int".  Explicit casts to "int" to remove compiler warnings.


gfwilliams:
Thanks - if you're adding other print statements, it's probably easiest to use `jsiConsolePrintf` - it'll work in all platforms and while not full printf, it handles stuff like `%v` for JsVars - making it seriously useful for debugging

PR #597, opened 2015-09-29
Adding Espruino for DPT-Board (OpenWRT MIPS and uClibc)

daanpape:
This is a pull request to add DPT-Board to the supported devices. I have updated the makefile to integrate nicely with the OpenWRT build system. I'm also applying the fixes from the pull request from vshymanskyy

I made a small mistake in the first commit, so therefore there are two commits.


gfwilliams:
Thanks!

Issue #596, opened 2015-09-29
Don't reuse socket ids for ESP8266


nkolban:
When a user writes a JS application that works against sockets (directly or indirectly through HTTP), the call to connect returns a socket id which is an integer that is the key to the socket data structure maintained by the environment.   Todays implementation maintains an internal array of MAX_SOCKETS of structures and the key that is returned is merely the index into the array.  This is not satisfactory ... consider the following:

1. Create a socket
2. id = 3 is returned
3. Close the socket
4. Create a socket
5. id = 3 is returned

And now we see that socket ids (as returned from the ESP8266 board handlers) are now re-used.  This opens up a potential issue that a program that "thinks" it has a good socket id may now accidentally use a socket that was closed and re-opened by other logic.

This is trivially fixable by making each socket id returned from creating a socket unique (simplest solution is to monotonically increment socket ids).

This work item asks to address that.

nkolban:
Starting to think in detail about the line item.  The affected area is all in the source file called `libs\network\esp8266\network_esp8266.c`.  Today, there is an array of sockets.  The size of the array is a constant (`MAX_SOCKETS`).  The name of the array is `socketData`.

The new design envisages a global called `int nextSocket` which increments each time a new socket is requested.  Each socket data structure (`struct socketData`) already has a field called `socketId` which holds the socket number.

Today we have a function called `getNextFreeSocket()` which returns an array index into the socket structure array.  This will be changed to be called `allocateNewSocket` which will now return a socketData structure instance (from the array) which the `socketId` sent correctly.   On no sockets available, NULL will be returned.

We also have a second function called `getSocketData()` which already returns a socketData pointer when passed in a socket id.  This will be changed to use the socketId property to find the correct socket.

The design notes at the top of the file will also need to be changed.

The `tcpArray` which holds the array of `esp_tcp` structures will be removed and the allocation of a new socket will `os_malloc()` one while the release of a socket will `os_free()` one.  The same with the `espconnArray`.

While I consider these changes "radical" to the internals, the complexity is relatively low.  In addition, since the socket constructs are fully encapsulated within this one source file, no coordination with other functions will be needed.

For scheduling of the work, I recommend a solid 4 hour session be allocated for the effort to include detailed design, coding and testing.   The impact areas will be all HTTP requests and all raw sockets.

nkolban:
Work starting ...

nkolban:
Pull request #608 has been merged and that concludes this work item.

PR #608, opened 2015-10-03
Implementation of #596 - unique socket ids

nkolban:
Issue number #596 looks to create unique socket ids that are never reused.  Prior to this, the actual socket id value was re-used when a socket was released and then a new socket claimed.  This design causes a new socket id to always be created and never re-used.  This reduces the possibility that an app may be using a socket id that it was closed and re-opened elsewhere.

In addition, this pull request includes in-line code commentary (without code logic changes) to some other unrelated files.


gfwilliams:
Thanks!

PR #611, opened 2015-10-04
Multiple new esp8266 features

tve:
This PR contains:
- support for flash save area (`save()` and `load()`)
- support for time-of-day (`setTime()` and `getTime()`)
- preliminary support for utilityTimers, needs clarifications
- split of 512KB vs 1MB..4MB boards due to feature differences
- wifi status improvements
*Note*: I added comments with `???` into `src/jsHardware.h`, please review all comments added for correctness and answer questions :-).


Makefile
```diff
@@ -1513,7 +1516,7 @@ proj: $(PROJ_NAME).elf $(PROJ_NAME)_0x00000.bin $(PROJ_NAME)_0x10000.bin
 $(PROJ_NAME).elf: $(OBJS) $(LINKER_FILE)
 	$(Q)$(LD) $(OPTIMIZEFLAGS) -nostdlib -Wl,--no-check-sections -Wl,-static -r -o partial.o $(OBJS)
 	$(Q)$(OBJCOPY) --rename-section .text=.irom0.text --rename-section .literal=.irom0.literal partial.o
-	$(Q)$(LD) $(LDFLAGS) -o $@ partial.o -Wl,--start-group $(LIBS) -Wl,--end-group
+	$(Q)$(LD) $(LDFLAGS) -Ttargets/esp8266/eagle.app.v6.0x10000.ld -o $@ partial.o -Wl,--start-group $(LIBS) -Wl,--end-group
```
gfwilliams: Any reason you can't just use the `LINKER_FILE=foo.ld` that we have for the NRF targets? I think it'd tidy things up.

gfwilliams:
Thanks! Will update jshardware.h in a bit.

Makefile
tve: Yes, I should have used the LINKER_FILE definition instead of having the filename in plain there, but I need two linker files 'cause I need to produce two binaries for OTA stuff. So while right now the $(LINKER_FILE) as a dependency is bogus and should be fixed, it's not going to simplify the Makefile overall.

PR #616, opened 2015-10-05
flash working 

mjdietzx:
Added functionality for flash and cleaned up code a bit. Flash module works now and save() and load() seem to work.


gfwilliams:
Looks great, thanks!

Issue #632, opened 2015-10-08
Get licence into ESP8266 files


gfwilliams:
Hi Neil,

Assuming you're happy with the license, please can you add the following header to the ESP8266 files you've made?

The rest of Espruino files have it, and I'm supposed to have it for MPL.

```
/*
 * This file is part of Espruino, a JavaScript interpreter for Microcontrollers
 *
 * Copyright (C) 2015 Neil Kolban
 *
 * This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/.
 *
 * ----------------------------------------------------------------------------
 * This file is designed to be parsed during the build process
    ^^^^ include this in jswrap_*.c
 *
 * A description of the file
 * ----------------------------------------------------------------------------
 */
```

nkolban:
Absolutely ... I have no need to have my name attached to the source.  Would it be ok with you if I changed my name to yours in the source files?  You should be the owner of all this stuff.

gfwilliams:
Wow, thanks - that'd be great. Maybe put your name down as well though?

nkolban:
The changes have been made to the source files and will be included in a commit and subsequent pull request.  There were two files that I couldn't change ... those were `targets\esp8266\hw_timer.c` and `targets\esp8266\hw_timer.h`.  They were introduced by @tve and I don't know their ancestry.   Suggest that we ask @tve to comment upon them and then you can decide what if anything needs to be done to change comments or register a dependency/inclusion.

MaBecker:
Hi @gfwilliams  as the pull request is merged, I suggest to close this.

gfwilliams:
Thanks! good plan :)

Issue #603, opened 2015-09-30
Error while connecting to TCP on second go


samehhady:
When uploading this sample sketch everything works fine on the first time, but when i re-upload it again I get this error and the ESP restarts.

```html
ASSERT(pSocketData->state != SOCKET_STATE_UNUSED) FAILED AT libs/network/esp8266/network_esp8266.c:781
  #1[r1,l2] Object {
    #2[r1,l2] Name String [1 blocks] "ÿ"      #3[r1,l2] Object {
        #6[r1,l2] Name String [2 blocks] "timers"          #8[r2,l1] Array(0) [ ]
        #9[r1,l2] Name String [2 blocks] "watches"          #11[r2,l1] Array(0) [ ]
        #14[r1,l2] Name String [1 blocks] "net"          #219[r1,l3] String [2 blocks] "\x04\x00\x00\x00\x00ÿÿÿ"
        #15[r1,l2] Name String [2 blocks] "history"          #19[r1,l1] Array(8) [
            #20[r1,l2] Name Integer 0              #16[r1,l1] String [2 blocks] "echo(0);"
            #25[r1,l2] Name Integer 1              #29[r1,l1] String [9 blocks] "console.log(\"<<\"+\"<<<\"+JSON.stringify(process.env)+\">>>\"+\">>\");"
            #24[r1,l2] Name Integer 2              #39[r1,l1] String [2 blocks] "echo(1);"
            #153[r1,l2] Name Integer 4              #150[r1,l1] String [4 blocks] "var net = require(\"net\");"
            #166[r1,l2] Name Integer 5              #102[r1,l1] String [42 blocks] "function startTCP(){\n  net.connect({host:\"54.68.0.122\", port: 8080}, function(socket) {\n    console.log(\"Connected to TCP\");\n    socket.write(\"GET / HTTP/1.1\\r\\n\");\n    socket.on('data', function(data) {\n      console.log(\"Got: \" + data);\n    });\n    socket.on('close', function(data) {\n    console.log(\"closed\");\n    });\n});\n}"
            #183[r1,l2] Name Integer 6              #180[r1,l1] String [18 blocks] "ESP8266WiFi.connect(\"Sam WIFI\", \"sameh.hady\", function(response){\n console.log(\"connected ;)\"); // Now should be connected\n  startTCP();\n});"
            #157[r1,l2] Name Integer 7              #226[r1,l1] String [3 blocks] "ESP8266WiFi.init();"
          ]
[...]

PR #633, opened 2015-10-09
Pull request for 2015-10-08

nkolban:
Work on #603 and #632


libs/network/esp8266/jswrap_esp8266.c
```diff
@@ -354,10 +411,10 @@ void jswrap_ESP8266WiFi_onWiFiEvent(
   ) {
   // If the callback is null
   if (callback == NULL || jsvIsNull(callback)) {
-    if (jsWiFiEventCallback != NULL) {
-      jsvUnLock(jsWiFiEventCallback);
+    if (g_jsWiFiEventCallback != NULL) {
+      jsvUnLock(g_jsWiFiEventCallback);
```
gfwilliams: jsvUnLock is null-safe, so here (and in other places) you don't actually need the explicit pointer check - not that it does much harm though

gfwilliams:
Thanks!

MaBecker:
@nkolban can this be closed ?

nkolban:
I don't think "pull requests" are closed ... they are just "merged" and thats the end of the path?

MaBecker:
hmm - as I am not an expert on this I checked file jswrap_esp8266_network.c which has the changes you made.

gfwilliams has merged into espruino:master on 0 Oct 2015 - so this look complete to me.

What do you think ?

nkolban:
I think it is complete ... its status says that it is "Merged" which in the set of "pull requests" is the equivalent in issues of it being closed?

PR #647, opened 2015-10-12
OTA update of firmware for esp8266

tve:
This PR adds a tiny web server to support OTA updates via port 88. The host-side support script is scripts/wiflash and how to use it is documented in README_Building.md

I implemented this in an esp8266 native manner because it's a rather small module that is completely independent of Espruino and is active even if Espruino is messed-up (as long as there's wifi connectivity, of course). Using the sockets library also would have required more code and more memory for buffers.

Note that this PR also reduces the number of JsVars from 1023 to 800 for the "OTA" versions of the esp8266 because there's not enough memory for 1023 IMHO. With 1023 vars there's about 5KB left after start-up, and that causes out of memory errors if more than one socket is active, or if one has more than 2 network buffers (1460 bytes ea) anywhere.


gfwilliams:
Wow, that's great - thanks! Yes, something separate for updates sounds like a good plan. Is the server always active, or only at bootup?

So you're saying that every time you open a new socket, ESP8266 itself allocates ~1.5kB?

PR #650, opened 2015-10-13
Olimexino with a STM32F103RET6 replacing the RBT6

jalawson:
Increases the memory of the Olimexino_STM32 board
to 512K FLASH and 64K SRAM


gfwilliams:
Thanks! Is there actually an RE Olimexino available? I couldn't find one on Olimex's site

jalawson:
No, you are correct, this isn't an actual Olimex product. It is one of the regular Olimexino-STM32 boards with the 
microcontroller swapped out. 
I also have one with a RG version (1M FLASH, 96K SRAM). 

Is there a better way of implementing slight variations of a board? 

I assume (understandably) that you don't want to litter the build with a vast number of very similar boards. 

Jamie 
----- Original Message -----

gfwilliams:
Hmm - yes, it's difficult. I think ideally the Makefile would be modified such that all the build info came from the `BOARD.py` file, then it wouldn't be a big deal having a bunch of files in there.

Thanks for the PR though, I'll merge it in.

Issue #643, opened 2015-10-10
Implement Espruino setWatch on ESP8266


samehhady:
I've been trying the `IRReceiver` module on ESP8266 and found that it is using `setWatch`
Currently this functionality is not implemented yet on ESP.

Talking to @nkolban he mentioned that at a minimum, it looks like ESP8266 has to implement `jshPinWatch` and `jshCanWatch` which are stubbed to be "NONE" and "false" right now.

nkolban:
Today, `jshCanWatch` for ESP8266 returns false for every pin indicating that none of the pins are able to support interrupt driven change detection.  Further, `jshPinWatch` is stubbed to do nothing.

soswow:
Oh, finally I found this issue after an hour debugging and understanding why library http://www.espruino.com/DHT22 doesn't work =) (it was using setWatch)

gfwilliams:
Yes, I think people are going to end up a bit disappointed when they try the really realtime stuff lime WS2811 and DHT11/IR/radio/etc

I'm tempted to put a bit of a note after the Espruino banner that mentions it's not 100%? it would suck if people got the wrong impression about Espruino itself because of the ESP8266 port

soswow:
@gfwilliams I was aware it's a Beta, and I know it was brought in very lately from separate repo, so I am not surprised or disappointed much. That's BETA is for. But maybe having it said explicitly is good enough.

nkolban:
But the GREAT news is that ESP8266 does indeed has interrupt handling for GPIOs ... it simply hasn't been implemented in the ESP8266 port yet.  Weve been studying the Espressif SDK and it doesn't look like rocket science to add it.  As such, that would then tell me that the speed of running Espruino on the ESP8266 should be no worse than other boards that are also capable of running at 160MHz.

gfwilliams:
Yes, really good news about the IRQs. And @tve implemented the utility timer - does that run off a proper IRQ as well?

> the speed of running Espruino on the ESP8266 should be no worse than other boards that are also capable of running at 160MHz.

Unfortunately that's probably not entirely right - the ESP8266 is having to load a lot of program code out of SPI flash in order to run it (as far as I'm aware?) so I think realistically it's going to end up slower than even the 72Mhz parts... Not that it actually matters much though.

nkolban:
Now it is time to write a use case that should work ... and then we can test against it.

gfwilliams:
> the next step will be linking these into the Espruino event processing logic. ....

[...]

Issue #642, opened 2015-10-10
Add ESP8266 support for mDNS


nkolban:
The ESP8266 SDK provides support for mDNS (Multicast Dynamic Name Server).  This issue asks for and tracks development towards mDNS support.

nkolban:
Test code was built which introduces a function called `ESP8266WiFi.mdnsInit()` which hard codes a call to the ESP SDK `espconn_mdns_init`.  This will **not** be exposed in the end ... however, it is useful at this early level of testing.

nkolban:
The test of `ESP8266WiFi.mdnsInit()` has been included in a Pull Request.  Further design and implementation of remaining functions still required.

tve:
I'm commenting out the mDNS support in PR #666 as a temporary move. I'm not sure what to do. mDNS is very helpful, but I find the esp8266 port to be really tight on memory. I think we need to set-up a little benchmark that streams some data and the see whether we can make it survive after adding mDNS... Unfortunately with/without mDNS is a compile-time decision, not a run-time one.

Bluebie:
Multicast DNS runs atop UDP. Why does this need to be in the core firmware? Shouldn't this be something people implement in Javascript as needed, or pull in from NPM?

nkolban:
The ESP8266 provides an implementation mDNS as part of its built in capabilities.  As such, this work item merely exposes the inherent capability of the ESP8266 itself.  Nothing special was needed to be added into the Espruino implementation other than the ability to "switch it on or off".  NPM (as I understand it) is a Node.JS concept and separate from the concept of JavaScript as a programming language.

Bluebie:
Maybe I'm misinterpreting this, but it sounded like the mDNS functionality in the ESP8266 SDK wasn't being compiled in until the functions were referenced, and that was blowing up the size of the firmware? The "1200 bytes of storage" statement is a bit ambiguous. If it's not using any extra resources when the user doesn't initialize the feature then this change makes a lot of sense to me. :)

Regarding NPM: The Espruino IDE includes an experimental feature to import code from NPM repositories in order to satisfy code dependancies, which would allow users to publish modules that provide extra functionality. It should also allow users to make use of pure javascript libraries that don't depend on unimplemented NodeJS APIs, like pure javascript crypto libraries. NPM isn't just used for NodeJS - it also distributes javascript code used in web apps, and there's no reason embedded targets can't represent another platform.

nkolban:
[...]

PR #651, opened 2015-10-13
Implementation of #642 and #643

nkolban:
Initial start on mDNS support and full implementation of interrupt driven GPIO for `setWatch`.

### DIFF
--- a/src/jsdevices.c
+++ b/src/jsdevices.c
@@ -270,9 +270,18 @@ void jshPushIOCharEvent(
   ioHead = nextHead;
 }
 
-void jshPushIOWatchEvent(IOEventFlags channel) {
+/**
+ * Signal an IO watch event as having happened.
+ */
+void jshPushIOWatchEvent(
+    IOEventFlags channel //!< The channel on which the IO watch event has happened.
+  ) {
+  assert(channel >= EV_EXTI0 && channel <= EV_EXTI_MAX);
+
   bool state = jshGetWatchedPinState(channel);
 
+  // If there is a callback associated with this GPIO event then invoke
+  // it and we are done.
   if (jshEventCallbacks[channel-EV_EXTI0]) {
     jshEventCallbacks[channel-EV_EXTI0](state);
     return;
@@ -289,7 +298,13 @@ void jshPushIOWatchEvent(IOEventFlags channel) {
   jshPushIOEvent(channel | (state?EV_EXTI_IS_HIGH:0), time);
 }
 
-void jshPushIOEvent(IOEventFlags channel, JsSysTime time) {
+/**
+ * Add this IO event to the IO event queue.
+ */
+void jshPushIOEvent(
+    IOEventFlags channel, //!< The event to add to the queue.
+    JsSysTime time        //!< The time that the event is thought to have happened.
+  ) {
   unsigned char nextHead = (unsigned char)((ioHead+1) & IOBUFFERMASK);
   if (ioTail == nextHead) {
     jshIOEventOverflowed();
@@ -515,7 +530,11 @@ void jshSetFlowControlEnabled(IOEventFlags device, bool xOnXOff) {
 }
 
 /// Set a callback function to be called when an event occurs
-void jshSetEventCallback(IOEventFlags channel, JshEventCallbackCallback callback) {
+void jshSetEventCallback(
+    IOEventFlags channel,             //!< The event that fires the callback.
+    JshEventCallbackCallback callback //!< The callback to be invoked.
+  ) {
+  // Save the callback function for this event channel.
   assert(channel>=EV_EXTI0 && channel<=EV_EXTI_MAX);
   jshEventCallbacks[channel-EV_EXTI0] = callback;
 }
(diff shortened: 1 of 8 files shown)


src/jsdevices.c
```diff
@@ -270,9 +270,18 @@ void jshPushIOCharEvent(
...
 }
 
-void jshPushIOWatchEvent(IOEventFlags channel) {
+/**
+ * Signal an IO watch event as having happened.
+ */
+void jshPushIOWatchEvent(
+    IOEventFlags channel //!< The channel on which the IO watch event has happened.
+  ) {
+  assert(channel >= EV_EXTI0 && channel <= EV_EXTI_MAX);
```
gfwilliams: Are you sure this is really the case? I can't remember for sure

src/jsdevices.h
```diff
@@ -23,22 +23,22 @@ void jshInitDevices(); // called from jshInit
...
-  EV_EXTI7,
-  EV_EXTI8,
-  EV_EXTI9,
-  EV_EXTI10,
-  EV_EXTI11,
-  EV_EXTI12,
-  EV_EXTI13,
-  EV_EXTI14,
-  EV_EXTI15,
+  EV_EXTI0,  // External Interrupt 0
```
gfwilliams: Do you need `//<` for these to register with Doxygen?

src/jspin.c
```diff
@@ -30,6 +30,10 @@
  */
 bool jshIsPinValid(Pin pin) {
   // Note, PIN_UNDEFINED is always > JSH_PIN_COUNT
+  if ((pin < JSH_PIN_COUNT && pinInfo[pin].port != JSH_PORT_NONE) == 0) {
```
gfwilliams: This doesn't seem right at all. `jshIsPinValid` is a check - it shouldn't produce an error message, and it's used in IRQs where writing to the console could potentially lock the system up. You could create a second function that did produce an error message if you needed to?

src/jswrap_io.c
```diff
@@ -576,7 +572,11 @@ and there will be no debouncing.
 watch two pins with the same number - eg `A0` and `B0`.
 
  */
-JsVar *jswrap_interface_setWatch(JsVar *func, Pin pin, JsVar *repeatOrObject) {
+JsVar *jswrap_interface_setWatch(
+    JsVar *func,           //!< A callback function to be invoked when the pin state changes.
```
gfwilliams: You know these are documented in the JSON docs right above - and in more detail? This would be one of those times that having some JSON doc->doxygen formatter for Doxygen would be a huge help

targets/esp8266/jshardware.c
```diff
@@ -85,17 +95,65 @@ static void utilTimerInit(void);
...
  * Reset the Espruino environment.
  */
 void jshReset() {
   //system_restart();
   os_printf("> jshReset\n");
+  // Set all GPIO pins to be input.
+  /*
+  int i;
+  for (int i=0; i<16; i++) {
+    jshPinSetState(i, JSHPINSTATE_GPIO_IN);
```
gfwilliams: There's a #define around - if you look at the STM32 implementation - that is something like 'is_pin_used' - it stops pins such as the serial IO getting set to inputs

gfwilliams:
Hi, Sorry for all the comments on this :) 

Please can you take the print out of `jshIsPinValid` and I can merge this? `IsPinValid` is just a check - if the pin isn't valid it's not an error condition.

In future, please can you try not to refactor the function calls to add comments  - especially `jswrap_` ones where there is already good documentation for the arguments, just in the JSON above? *We definitely don't want duplicate documentation  - especially where the two comments are different!* I think far from being helpful, that just confuses anyone looking at the code.

... when you do update the docs, please can you update the header file too? That's where I usually put documentation as it's easier to search through if you're not using Doxygen - having two different sets of documentation there is pretty confusing too.

Finally - and sorry, this is a real pet hate of mine so it's a bit of a rant :)

```
 /**
  * Set the value of the corresponding pin.
  */
void jshPinSetValue(
    Pin pin,   //!< The pin to have its value changed.
    bool value //!< The new value of the pin.
```

You've got to ask yourself - is ANY of that needed? If someone can't look at `jshPinSetValue(pin,value)` and work out exactly what's in that documentation then they really shouldn't be hacking on Espruino, or anything, really :) The arguments are obvious, and (IMO) we shouldn't waste space describing them (in this case).

Now if it said "change the output data register of the GPIO pin, without changing the pin's state to an output" it would be really handy, but as it is (IMO) it's just wasting space.

**What's worse,  in the header file it describes it properly.** So you're overwriting documentation that was really helpful (and would have been the same across all platforms) with something that doesn't help at all!

```
void jshPinSetValue(Pin pin, bool value); ///< Set a digital output to 1 or 0. DOES NOT change pin state OR CHECK PIN VALIDITY
```

So please try and avoid doing that - as I understand it Doxygen will pull descriptions from the header file OR the source, so please try and keep docs in the header (where possible), or at the very least in sync.

nkolban:
Removed the print from `jshIsPinValid`.

gfwilliams:
Thanks!

Issue #655, opened 2015-10-13
Add ability to disable ESP8266 development debug logs


nkolban:
Within the ESP8266 specific code there are copious amounts of debug statements that log to the second ESP8266 UART known as UART1.   These logging statements have proven invaluable on the ESP8266 board as it lacks any richer ability to debug on the board beyond writing logs.  The debugs are written by ESP8266 API called `os_printf`.   The ESP8266 specific code is littered with such statements.

This issue asks for an exposed mechanism to switch on or off these debug steps.  It is believed that we do not want to simply "remove them" from the code base but rather selectively enable/disable.  We have choices on whether a build should contain debug or not and we will assume that a "RELEASE" build will not have them.  However at this stage, all builds are DEBUG ... and we still need a way to disable them.   Thankfully, the ESP8266 seems to have exactly an API for doing just that and what we need now is an exposed API similar to `ESP8266WiFi.logDebug(onOff)` that will enable/disable debug logging.

nkolban:
From the ESP8266 SDK, the API to enable/disable debug output is `system_set_os_print`.  This takes a value of 0 to disable and a value of 1 to enable.

PR #657, opened 2015-10-13
Implementation of #655

nkolban:
Issue #655 adds the ability to disable ESP8266 debugging to UART1.  The new JS level API called `ESP8266WiFi.logDebug(onOff)` has been added.


gfwilliams:
Thanks - looks great!

gfwilliams:
Quick question - is the debugging on by default?

IMO it should probably be off, as most users won't know/care about it.

nkolban:
During the ESP8266 board port beta, the debugging is indeed on by default but needs to be default of off when the beta ends.

Issue #660, opened 2015-10-14
Allow for pinMode input_pullup in ESP8266


rakeshpai:
It appears that `input_pullup` isn't currently implemented. I'm using the following code:
```
var button = new Pin(14);

pinMode(button, "input_pullup");

setWatch(function() {
	print("Button clicked");
}, button, {repeat: true});
```

I have a button connected between GPIO14 and ground. When I press the button the first time, the `print` gets called correctly. However, subsequent button presses don't print anything.

If I then change the `pinMode` to `input`, and add a 10k pull up resistor between GPIO14 and 3.3v, the print gets called as expected upon every button press.

My guess is that that in the first case, the internal pullup resistors in the esp8266 aren't actually enabled, so the pin floats near ground after the first button press, never rising again.

nkolban:
There appears to be ESP8266 SDK functions called `PIN_PULLUP_EN` and `PIN_PULLUP_DIS`.  Not yet sure if those are leveraged in the Espruino code base yet, but it seems that we have a pull-up capability in ESP8266 which is a good start.  Not seeing a "pull-down" ability though.

opichals:
https://github.com/nekromant/esp8266-frankenstein/blob/master/src/cmd_ds18b20.c#L176

rakeshpai:
Just a heads-up - @tve mentioned something in Gitter today about this already implemented in his fork. We should check with him before duplicating effort.

nkolban:
I found the source of the error.  The Espruino code does indeed honor the pull-up ability ... however, in the implementation of `jshPinWatch` for ESP8266 we currently set the state of the pin to be `input` before we register an interrupt handler for it.  The thinking here is that there is no point in registering an interrupt handler for an `output` pin.  However, the flag for setting it as input disregards the concept of `input with pullup` and hence we force it to be input only.   The design change I propose is that when jshPinWatch is called it first checks to see if the pin is in any input mode ... if it is, leave it alone.  If it isn't set it to input (simple).  That way, a JS programmer can pre configure the pin to be input or input with pull-up and whatever its current state is when jshPinWatch is caller, will still be honored.

gfwilliams:
@nkolban there's something called `jshGetPinStateIsManual` (or similar) which is used in the STM32 implementation. It sets the pin to an input if the state wasn't set, but if it was it leaves it as-is.

nkolban:
WOW!!!   Perfect ... exactly what I needed.

nkolban:
[...]

PR #663, opened 2015-10-14
Implement of #660.

nkolban:
This issue corrects the handling of a pin defined as input with pullup
as related to `setWatch()`.


src/jspin.c
```diff
@@ -212,10 +212,16 @@ Pin jshGetPinFromVarAndUnLock(JsVar *pinv) {
   // Whether a pin's state has been set manually or not
 BITFIELD_DECL(jshPinStateIsManual, JSH_PIN_COUNT); // TODO: This should be set to all 0
 
+/**
+ * Determine if the state of a given pin has been defined as being manually set.
```
gfwilliams: Please can you either take this out, use the existing docs that are in the header file, or update the docs in the header file to say the same thing?

This function should already be documented in Doxygen (and Eclipse's tooltips) because of the comments in the header file - it's crazy having *two different sets of comments* - especially when the existing ones (that will presumably get overwritten) elaborate more on what 'manually set' is:

https://github.com/espruino/Espruino/blob/master/src/jspin.h#L207

Sorry to keep banging on about this, but it's wasting everyone's time and is actually making the source **harder to understand**.

nkolban:
I have removed the bad comments.  As always, I'd much rather be told that I was making mistakes so that I can fix them than have resentment.  I am delighted to make the changes.  There can only be one leader ... and I vote for you.  I'll try harder and harder to get the comments right.

gfwilliams:
Great, thanks!

And yeah, sorry about being picky about this. Having docs at source level is great, but I think it's best to have them at header file level *first* and then duplicate them at source level if needed.

In a lot of editors if you ctrl-click on a function/variable name that you see, it'll go direct to the header file so having docs there is really handy (having it go to source requires the editor to have looked at all files, rather than just your current file and the `#include`s).

Issue #649, opened 2015-10-12
Add I2C master support to esp8266


tve:
The initial proposal is to use a slightly modified version of the Espressif i2c library (there is no i2c hardware in the esp8266).

tve:
Q: What is `jshI2CSetup` supposed to do if the parameters passed to it are invalid/impossible? For example, if it's asked to be a slave and it can't do that?

gfwilliams:
I think it just calls jsError - generally it outputs a list of pins that would be allowed for that peripheral (I'd see the STM32 implementation). Swapping to return true/false too would be cool though.

Do you think it'd be possible to actually implement Software I2C ourselves? It's already done for SPI.

There's a bug for it here: https://github.com/espruino/Espruino/issues/549

It looks pretty trivial to make our own implementation - looking at the Arduino one anyway.

That way it'd add functionality to normal Espruino boards as well.

tve:
My current plan is to slap in the esp8266 specific sw i2c. I'd like to see that working first. But I agree that the right way would be a hardware independent implementation using gpio and perhaps timers. I need to read up on how all that hardware timer stuff works in espruino first...

gfwilliams:
With I2C, timing isn't generally a big issue (as far as I can sell) so it'd really be a matter of just bashing it out as fast as possible.

That link above mentioned Serial, and timers would be handy there - especially on the receive side of things.

tve:
This is implemented. Please reopen if there are issues.

Issue #598, opened 2015-09-29
Get the Pin mapping correct for the ESP8266


nkolban:
In the configuration file called `boards\ESP8266_BOARD.py` there are definitions of the pins that are available to be used.   At this point, the definitions are basically a place holder.  Work needs to be done to understand the format of this definition file and design and implementation made to set the definitions correctly based upon what is found.

nkolban:
@tve has made progress on this item in #611.

Issue #549, opened 2015-06-30
Software Serial


gfwilliams:
Shouldn't be too difficult, especially for lower baud rates. Software SPI is already there, and we can execute a user-defined function from inside a watch IRQ.

DrAzzy on the forums definitely wants it anyway.

gfwilliams:
Also: this could be a way to get Serial that works from within deep sleep.

gfwilliams:
Software I2C looks [trivial](https://github.com/landis/arduino/blob/master/libraries/I2cMaster/I2cMaster.cpp), and could be implemented in the same way as software SPI.

HyGy:
This UART will be good to implement to support http://wiki.iteadstudio.com/Nextion_HMI_Solution can somebody implement please?

@tve maybe? :)

gfwilliams:
I think we ought to have a bounty system. How much do you want it... enough to pay for the time we spend on it? :)

HyGy:
Do it need to implement in javascript? If yes, then can u please help me to start it? (I promiss nothing, but maybe a try... I don't know how good now in low level things, this is why I like espruino couse its js :) and higher level language )

profra:
... bounty system... It is not bad idea... I am ready to contribute... You can try to begin the discussion in the News... but I have no idea about the total price for such task...
I have asked for SW UART already in the times of Kolban...
...I have been just playing with Lua and WiFiMCU (EMW3165) and there is possible to emulate UART on whatever two pins (TX, RX)...but it is clear, it is another HW platform where you have sufficient quantity of timers for easy implementation and other sources...I am not able to review the possibility to implement it on ESP8266... maybe @tve
.some inspiration for SW UART you can find in the source of WiFiMCU... here..
..https://github.com/loboris/MICO/blob/master/Projects/WiFiMCU/LUA/lua/exlibs/uart.c...
...IMO it has to be written in C

gfwilliams:
Well, to do it properly you'll need to do it in C. However if you just want to send a few characters you can use JavaScript and digitalPulse, at least for stuff at slow-ish speeds like 9600 baud.

All you need to do is keep the output at 1 usually, then you can use `digitalPulse(pin, 0, array)` to create an RS232 waveform for each byte.

HyGy:
I think C is hard for me now.

Here is some js implementation but it does not work yet. Thanks for yerpj user from gitter.

Now I just send 'a' ascii char what i think is 01100001.

[...]

PR #666, opened 2015-10-16
Esp8266 I2C support and pin refactoring

tve:
This pull request refactors the way pins are handled in order to correctly support pull-up and open-drain outputs in particular. It also adds software I2C support. The I2C support is esp8266 specific using a module from Espressif's SDK. In the end it could easily be rewritten in a device independent manner using general jshPin functions (maybe with some inlining) but right now I'm happy I have something. (Maybe a short discussion on where to implement it in a device independent manner would be helpful.)

Some other changes: I reverted back to 1023 JSvars because not even the relatively small BMP085 module would load otherwise and I "compensated" by switching to the 536 byte TCP MSS LwIP library. I also reduced the buffer sizes required by the OTA flashing. I tested what I could.
One known issue is that the I2C implementation doesn't support clock stretching. I don't know how important that is in practice.
NB: apologies for all the whitespace changes in `jshardware.h`, my editor automatically removes spaces at the end of lines. There's only one real change on line 317 to ensure that slaveAddr is signed.

This PR addresses: #649 #598 #549


gfwilliams:
Thanks - looks great!

Issue #670, opened 2015-10-16
Memory Leak when auto-connecting to a non-existent AP


markusgritsch:
There seems to be a memory leak somewhere when the ESP8266 tries to auto-connect to a not-present AP.   Every 1-2 seconds a connection is attempted and 6 JSvars get periodically comsumed without any user code running.

After about 5 minutes, all 1023 JSvars are consumed, and the board resets.

The leaking is stopped, when I call ESP8266WiFi.disconnect().

nkolban:
Starting to examine now.

nkolban:
Fixes included in the master repository.  Latest firmware builds now include the fixes.  Waiting for @markusgritsch to validate.

markusgritsch:
Thanks. Works fine.

ESP8266WiFi.init(); still looses two JsVars on each call though.

PR #671, opened 2015-10-16
Fixes for #670.

nkolban:
Issue #670 showed that we weren't releasing all the JsVars that we should.  This Pull Request includes fixes for that problem.


gfwilliams:
thanks!

Issue #630, opened 2015-10-08
Creation of a new static object specifically for ESP8266 functions.


nkolban:
We currently have all the ESP8266 functions hung off a badly named class called `ESP8266WiFi`.  There is a work item in the queue to design a generic WiFi class ... see issue #589.   When that work comes to fruition, the majority of methods will be moved to the new WiFi class/library ... however, there will be methods that will be ESP8266 specific such as:

* updateCPUFreq
* getRstInfo
* restart
* debugLog

These need to be moved into a new class that is static that will be ESP8266 specific.  This work item is to track this work and should be coupled with issue #589.

nkolban:
An initial start has been made on this issue.  The selected name of the module is `ESP8266` resulting in user coding of:

    var ESP8266 = require("ESP8266");

The function called `getAddressAsString()` has been implemented.

Issue #589, opened 2015-09-28
WiFi generic object interface


nkolban:
Background: [Forum article post](http://forum.espruino.com/conversations/275060/#comment12536663)

As more board types start to offer WiFi services such as:
* connect as station
* list access points 
* be an access point

Unless we do something, we will see fragmented implementations and different styles of interaction.  This work item will include the requirements gathering, design, documentation and implementation of a general purpose interface that is board agnostic.

One of the early implementors of the interface will be the ESP8266 port.

A [WiKi page](https://github.com/espruino/Espruino/wiki/Generic-WiFi-Design) has been created to capture discussion items for all on this work item.

nkolban:
This work item needs coupled with #630.

nkolban:
Initial lower level design work has started but we immediately ran into a major design question.  Should the `wifi` object be a `class` or a `library`?  Can we store `state` of the `wifi` object in this class or library?  For example, if one calls the `connect()` method and passes in a callback function, can we store a reference to the callback function off some JS object instance that represents the `wifi` object?  And if so, how?

gfwilliams:
I'd a library. I know you hate me saying 'look at the code', but:

JSNetwork stores callbacks in a hidden `net` variable: 

https://github.com/espruino/Espruino/blob/master/libs/network/js/network_js.c

And the other network stuff stores data (pins, etc) in JsNetworkData, which it accesses like:

```
  JsNetwork net;
  if (!networkGetFromVar(&net)) return;

  // do stuff
 
  // if it changed, then
  networkSet(&net);

  networkFree(&net);
```

* https://github.com/espruino/Espruino/blob/master/libs/network/cc3000/jswrap_cc3000.c
* https://github.com/espruino/Espruino/blob/master/libs/network/wiznet/jswrap_wiznet.c

I'd be tempted to stick with the hidden `net` variable though, because it'll be more resilient to `save()`/etc.

If you were to store all data in a single user-accessible class then it's possible someone might do something like create another, or clone the existing and it'd really screw you up.

nkolban:
Howdy @gfwilliams I looked at the code of cc3000 and of wiznet ... and think I realize that their is a misunderstanding on what I am thinking.  I see that the `JsNetwork` object contains the callbacks to the generic TCP/IP functions such as `createsocket`, `closesocket`, `accept` etc etc and the ESP8266 falls in line with that architecture today.

[...]

PR #678, opened 2015-10-25
Implementations of #589 and #630.

nkolban:
This pull request sees a significant dent in the work items required for #589 and #630.  While the work is likely not 100% done ... the code is a consistent state with no obvious regressions from what was there before.  One should study the Issues for #589 and #630 for full details.


libs/network/network.c
```diff
@@ -103,7 +114,17 @@ JsVar *networkGetAddressAsString(unsigned char *ip, int nBytes, unsigned int bas
   return jsvNewFromString(data);
 }
 
-void networkPutAddressAsString(JsVar *object, const char *name,  unsigned char *ip, int nBytes, unsigned int base, char separator) {
+/**
+ * Put an IP address as a child property of a JsVar.
```
gfwilliams: Just to note that this (and others) aren't just for IP addresses. They're used for MACs too...

libs/network/network.c
```diff
@@ -148,7 +169,9 @@ void networkGetHostByName(
 }
 
 
-
+/**
+ * Create a network.
```
gfwilliams: You've done it again ;) https://github.com/espruino/Espruino/blob/master/libs/network/network.h#L77

I know the original wasn't Doxygen-formatted, and it is small - but it does provide more info...

gfwilliams:
Thanks! Merged :) When you get a second, please could you try and tidy up the documentation in `network.c`? There was already some documentation in `network.h` that contained more information about how the functions worked: https://github.com/espruino/Espruino/blob/master/libs/network/network.h#L77

PR #680, opened 2015-10-27
Function for #589 and comment removals

nkolban:
Issue #589 adds a generic WiFi interface.  This pull-request adds support for a wifi.connect() option called dnsServers.  When we connect to the WiFi access point, we can supply a property called "dnsServers" which is an array of up to two IP addresses which will be the DNS servers to use as opposed to being DHCP satisfied.

This Pull Request also undoes some comments added to network.c.


gfwilliams:
Thanks! With the comments, I'm not against you adding them at all - it's really good. I just want to make sure that if there are comments in the header file and the C file, that they're both saying the same thing.

PR #686, opened 2015-10-30
EMW3165 support without Wifi

tve:
This PR has support for the emw3165, including a board file and Makefile changes. I set the EMW3165 to use STM32F411xx and had to add #ifdefs for this chip series to various stm32 files. Dunno whether that was the right thing to do or not, but it seems to work. The problem I'm running into is that the stm32 files have trailing spaces on many lines and vim (or my vim settings) remove those automatically. So the diffs are littered with whitespace changes. I see two options: review the PR by adding `&w=1` to the github URL, which gives a diff without the whitespace changes; or I have to go back and figure out how to apply the changes to the original files without affecting the whitespace. You tell me :-).


tve:
This is the link to see the changes without the whitespace changes: https://github.com/espruino/Espruino/pull/686/files?w=1

tve:
Mhh, I'm not sure I understand why the travis builds are all broken...

tve:
OK, travis builds fixed

gfwilliams:
Thanks - looks great. Shame about whitespace but I'm not sure what we can do about that.

I did just re-add the 2 files that got deleted (`gen/README` and `.gdbinit`). With that you don't need the Makefile changes either so I took them out too

Issue #606, opened 2015-10-01
Test and resolve any issues with ESP8266 hardware SPI


nkolban:
We now have basic testing of software SPI tested and working.  It is now time to create a work item for the hardware SPI built into the ESP8266.  This work item will track the design and efforts necessary to perform that task.

nkolban:
Here is the link to the ESP8266 API docs ... http://bbs.espressif.com/download/file.php?id=571

nkolban:
After studying the SPI hardware support docs from Espressif ... I am unable to understand how to use these functions.  Looking now for someone who understands ESP8266 SPI details to provide guidance and consultation.

nkolban:
User @jumjum123 is looking for this ability to drive WS2812 devices.  This should indicate to us that the relative priority of this issue should increase.

nkolban:
Started examination of what is involved in integrating hardware SPI support.  This starts with the examination of the MetalPhreak/ESP8266_SPI_Driver project.

I was able to compile this project cleanly with no issues.

Started a new Wiki page called `SPI-Architecture.md` to document our understanding of how Espruino handles SPI processing.

Changed the `ESP8266_BOARD.py` definition to set the number of hardware SPI's to `1`.  This may later turn out to be need to be set to `2` but we'll get one working first before looking at the second SPI.

nkolban:
With the [MetalPhreak/ESP8266_SPI_Driver](https://github.com/MetalPhreak/ESP8266_SPI_Driver) included in the build and the framework for SPI hardware support examined, changes were made to `jshardware.c` to link Espruino to the library.  It compiled cleanly and SPI JS calls don't cause any fatal exceptions.  What remains now is to test this basic support with a logic analyzer and see what we see.

nkolban:
Integration of Espruino with MetalPhreak library happened exactly as desired with no hiccups.  We now have a hardware SPI known as `SPI1`.  Changes committed and integrated with a pull-request from yesterday.  This Issue will be left open pending user acceptance testing and problem resolution resulting from feedback.

markusgritsch:
Hmm, I tried
```javascript
SPI1.setup({baud:3200000});
SPI1.send4bit([255,0,0], 0b0001, 0b0011);
```
which gives me just this output
![shot](https://cloud.githubusercontent.com/assets/2684183/10962958/93689cf6-839c-11e5-9a8a-200169fdbceb.png)

markusgritsch:
Another test:
```javascript
[...]

Issue #675, opened 2015-10-19
Add ESP8266 board to the Web IDE module lookup system


nkolban:
When we use the Web IDE, modules that are referenced by `require` are showing up as not found.  It is believed that this is caused by the espruino.com web site not having a board definition for the ESP8266 in its configuration.   This issue asks for examination and resolution of that issue to prevent warnings from appearing within the Web IDE when being used with an ESP8266 module.

gfwilliams:
Which specific modules? If the built-in ones are creating errors, that'd definitely be the problem - if it's others then it'll be something else.

nkolban:
I believe the one reported to me was a `require('http')`.

gfwilliams:
Ok, yes - so that'd be it then...

nkolban:
With the work being done on #589 and #630 this issue is becoming more pressing as use of ESP8266 board now always needs "unknown" `require` modules.

gfwilliams:
Will see what I can do. I'm out again this week though - conference season :)

nkolban:
From my perspective, this is relatively low priority as everything is functioning even with these warning messages.  However, I believe that "impression" is as much quality as function and some users are expressing "query" over the messages in Web IDE.

nkolban:
User @Ducky is reporting the warnings about modules within the Web IDE.  See:

http://forum.espruino.com/conversations/266886/?offset=125#comment12608682

gfwilliams:
We now have ESP8266_BOARD and ESP8266_12 boards. Is this going to stay the case? There was some talk on the forum about just having a single binary for both boards?

nkolban:
This is where the community should come together and make a decision.  @tve did a lot of work on the OTA flashing and it would be great to involve him for sure.   There is also a work item on a re-design of the Makefile/build system.  I think any significant changes to the story on boards should probably await the decisions on the outcome of a reworked Makefile/build system as changes there would affect the design of boards vs board options.

I am also guessing that if we polled the user community, the majority of users who are reporting the problems are using the firmware images supplied at EspruinoBuilds ... and if that is the case, those are being compiled today using "ESP8266_BOARD".  If there is validity in that notion, then accommodating "ESP8266_BOARD" as a known board today (even if not strategically where we might want to go in the future) might satisfy the majority of users sooner than later.

gfwilliams:
Just done. Added the docs at http://www.espruino.com/EspruinoESP8266

[...]

PR #692, opened 2015-11-04
#606. Initial framework for hardware SPI for ESP8266.

nkolban:
The ESP8266 jshardware source file needs prepared for more detailed implementation of SPI hardware support.  This commit includes the framework for implementation of those detailed functions.

The hardware SPI source files are also introduced in this commit with the corresponding Makefile change to include them in ESP8266 compilation.


Makefile
```diff
@@ -1493,7 +1493,8 @@ LDFLAGS += -L$(ESP8266_SDK_ROOT)/lib \
 
 # Extra source files specific to the ESP8266
 SOURCES += targets/esp8266/uart.c \
-	   targets/esp8266/user_main.c \
+	targets/esp8266/spi.c \
+       targets/esp8266/user_main.c \
```
tve: looks like a space/tab mismatch?

boards/.gitignore
```diff
@@ -0,0 +1 @@
+/__pycache__/
```
tve: ?

boards/ESP8266_BOARD.py
```diff
@@ -28,7 +28,7 @@
   'flash'   : 512,
   'speed'   : 80,
   'usart'   : 1,
-  'spi'     : 0,
+  'spi'     : 1,
```
tve: Would be nice to also add it to the ESP8266_12.py board...

targets/esp8266/jshardware.c
```diff
@@ -588,15 +593,47 @@ void jshUSARTKick(
...
+ * Initialize the hardware SPI device.
+ * On the ESP8266, hardware SPI is implemented via a set of pins defined
+ * as follows:
  *
+ * | GPIO   | NodeMCU | Name  | Function |
+ * |--------|---------|-------|----------|
+ * | GPIO12 | D6      | HMISO | MISO     |
+ * | GPIO13 | D7      | HMOSI | MOSI     |
+ * | GPIO14 | D5      | HSCLK | CLK      |
+ * | GPIO15 | D8      | HCS   | CS       |
```
tve: Is there a way to have more than one device on the SPI bus, i.e., by specifying a second chip select pin?

targets/esp8266/jshardware.c
```diff
@@ -588,15 +593,47 @@ void jshUSARTKick(
...
+  case EV_SPI3:
+    os_printf(" - Device is SPI3\n");
+    break;
+  default:
+    os_printf(" - Device is Unknown!!\n");
+    break;
+  }
+  if (inf != NULL) {
+    os_printf("baudRate=%d, baudRateSpec=%d, pinSCK=%d, pinMISO=%d, pinMOSI=%d, spiMode=%d, spiMSB=%d\n",
+        inf->baudRate, inf->baudRateSpec, inf->pinSCK, inf->pinMISO, inf->pinMOSI, inf->spiMode, inf->spiMSB);
```
tve: It seems it would be good to return an error if the user specifies parameters that the implementation doesn't support.

Makefile
nkolban: Yup ... sorry about that.  Have corrected now.

boards/.gitignore
nkolban: In my build system I found that during build, my Python environment was writing python cache information.  I can't imagine that we would want that in our build system so added a gitignore for it.  If that is wrong, we can remove it.

boards/ESP8266_BOARD.py
nkolban: I agree ... there are a couple of issues floating around just now ...  see Issue #675 for the one that is most relevant.

targets/esp8266/jshardware.c
nkolban: That's a good question.  This table and documentation comes from ESP8266 "stuff".   I think that when one "drives" the ESP8266 hardware SPI, the SPI hardware knows to drive CS.  I suspect that if we wanted to drive alternate devices, we would ignore the hardware CS and dedicate alternate pins to CS.

targets/esp8266/jshardware.c
nkolban: These guards are an over-abundance of caution.  The identies of available SPI interfaces are not user configurable but instead defined in the ESP8266_xxx.py files.  Today we only have SPI1 but it is possible to have SPI2 and SPI3 as well.  However, nothing a user can code at their JS level could result in triggering these alternate code paths.

targets/esp8266/jshardware.c
```diff
@@ -618,9 +657,10 @@ void jshSPISend16(
     IOEventFlags device, //!< Unknown
     int data             //!< Unknown
   ) {
-  os_printf("ESP8266: jshSPISend16\n");
+  os_printf("> jshSPISend16 - device=%d, data=%x\n", device, data);
```
gfwilliams: These won't be helping with the speed *at all*, and you have `spi_tx16` available - so use that instead. It should make a big difference.

nkolban:
Howdy ... is there something that needs attention with this Pull Request before it can be merged?   Want to make sure that it isn't waiting on something I need to attend to.

gfwilliams:
Hi, sorry for the delay... Looks good - please could you just:

* tweak the indenting in the Makefile to match?
* add `spi:1` to `ESP8266_12.py` - I don't see any reason not to at the moment? Anything we can do to keep them in sync for now is good.
* Swap the licence text in `targets/esp8266/spi.h/c` to MPLv2? You said MetalPhreak was fine with that?

nkolban:
Requested changes made and committed.

gfwilliams:
Thanks, I'll merge - but you know you committed quite a lot of extra stuff this time as well? Or was adding NeoPixel intentional? 

https://github.com/nkolban/Espruino/commit/71d19752eea5941db68f8945d23969ba04da64c7

The AES branch is now merged too, so you could massively simplify your NeoPixel code.Literally just the following will do:

```
/*JSON{
 "type"     : "staticmethod",
 "class"    : "ESP8266",
 "name"     : "neopixelWrite",
 "generate" : "jswrap_ESP8266_neopixelWrite",
 "params"   : [
   ["pin", "pin", "Pin for output signal."],
   ["arrayOfData", "JsVar", "Array of LED data."]
 ]
}*/
void jswrap_ESP8266_neopixelWrite(Pin pin, JsVar *jsArrayOfData) {
  if (!jshIsPinValid(pin)) {
    jsExceptionHere(JSET_ERROR, "Pin is not valid.");
    return;
  }
  JSV_GET_AS_CHAR_ARRAY(dataPtr, dataLen, jsArrayOfData);
  if (!dataPtr) return 0;
 // ...
```

Note also that you can set the variable type in the JSON to `pin` and the conversion is done for you. Same for ints and floats too.

nkolban:
Pure gold!!!  I had no idea that the params section of JSON was more than documentation ... and had been filling it in as a good citizen by accident.  I didn't know it could dynamically cast parameters ... VERY nice.

The JSV_GET_AS_CHAR_ARRAY is also nice however I had to convince myself that I could understand it.  It seems to create new variables which I found an unusual side effect for a macro.  I can't say it is wrong ... just unusual (to me).

I have accommodated both changes and after testing, will commit them with the next batch.

gfwilliams:
Thanks! 

Yes, I was unsure about that behaviour, but the variables always have to be the same time so it seemed to make sense to avoid duplication. Maybe not...

For the JSON, it's used for the reference, to define how the function gets called, *and* for autocomplete in the Web IDE.

Some info here, maybe it needs clearing up a bit though:

* https://github.com/espruino/Espruino/blob/master/README_BuildProcess.md#wrapper-files-
* https://github.com/espruino/Espruino/blob/master/scripts/common.py#L52

... by making all the type conversions automatic, Espruino cuts out on all the duplicated code that handles that so it doesn't use up much Flash. It's also removing another potential source of bugs :)

PR #694, opened 2015-11-05
fix: typo in sleep pin indicator dump

nailxx:
Clear copy-paste bug that lead to error while `save()` if busy pin is set and sleep pin is not.

Also, note `jspin.c:187`. Should we write `UNKNOWN` there? Maybe `undefined` is a better idea? In current state, following scenario is possible:

```
Uncaught ReferenceError: "UNKNOWN" is not defined
 at line 1 col 19
setSleepIndicator(UNKNOWN);
                  ^
>dump()
setSleepIndicator(UNKNOWN);
```

Use of `undefined` instead of `UNKNOWN` would solve the problem.


gfwilliams:
Good spot - thanks! Presumably this is only if you set DEFAULT_SLEEP_PIN_INDICATOR in the build itself? Good point about the pin dumping, I'll do that

nailxx:
Yes, it's definitelly an issue if I set `DEFAULT_BUSY_PIN_INDICATOR` (busy) in the build.

Issue #702, opened 2015-11-11
Honor the default baud rate defined in ESP8266_BOARD.py


nkolban:
The default baud rate for the ESP8266 port is hard coded within the ESP8266 port code.  This should be changed to the default baud rate defined in the ESP8266_BOARD.py definition.  See the following thread for details:

http://forum.espruino.com/conversations/277346/#comment12620504

nkolban:
Design and coding complete.  Commiting the changes.

nkolban:
Now included in master build.

PR #703, opened 2015-11-11
Implementation of #702

nkolban:
Implementation of issue #702 which is about the default baud rate of the consoles.


targets/esp8266/jswrap_esp8266.c
```diff
@@ -87,13 +82,8 @@ __attribute__((section(".force.text"))) void jswrap_ESP8266_neopixelWrite(JsVar
...
-  uint8_t *pixels = (uint8 *)os_malloc(dataLength);
-  if (pixels == NULL) {
-    jsExceptionHere(JSET_ERROR, "Out of memory.");
-    return;
-  }
-  int i;
-  for (int i=0; i<dataLength; i++) {
+
+  for (size_t i=0; i<dataLength; i++) {
     JsVar *jsItem = jsvGetArrayItem(jsArrayOfData, i);
```
gfwilliams: You don't need any of this now... `JSV_GET_AS_CHAR_ARRAY` gets the char array but *also makes sure it's filled with the correct data*

Issue #595, opened 2015-09-29
Handle ESP8266 reconnectCB events


nkolban:
The ESP8266 reconnectCB event which is called when the ESP8266 detects a TCP/IP level error currently does no more than log a message.  This will result in a leak of socket descriptors or stalled code.  This work item asks for the design and implementation of handling the reconnectCB properly.

It is imagined that it will not be too complex ... and that a reconnectCB will "merely" set the state of the socket into its error state which will then result in a clean up in the future.   However since this event can almost literally occur at any time, analysis needs to be done on the state machine to ensure that the state handling is correct.

nkolban:
Examining the code, we find that reconnectCB does nothing.  Here are the recommended changes.

First within the reconnectCB, we call `setSocketInError` to flag the socket as now in the error state.  And then we can return from the callback.   Now we can figure out how to handle the error in the Espruino callback architecture.

For a recv ... we will say that if we have no data to return and SOCKET_STATE_IN_ERROR then return -1 which should flag upstream that we can close the socket.

For a send ... if we are in SOCKET_STATE_IN_ERROR then return -1 which should flag upstream that we can close the socket.

For doClose() ... things get a little bit more interesting.  If we have received an async error indication, should we call espcon_disconnect() and if so, will we ever get a callback?  This is research that must be completed before attempting the further design and implementation.  It may turn out that how we handle this will be a function of the error type and not just that an error was caught.

nkolban:
A question has been created on the ESP8266 forum to see if we can figure out what might happen to the state of the socket if a reconnectCB happens.

See: http://www.esp8266.com/viewtopic.php?f=9&t=5864

nkolban:
While waiting for the above response, we have implemented some basic flagging logic as we are finding that the absence of a handler is more troublesome than getting it right first time.

tve:
When you get a reconcb the connection has been closed. It's equivalent to a disconnect callback, except that it tells you that an error has occurred, i.e. it's not a clean remote-party initiated connection close.

nkolban:
Thanks @tve  ... so the notion then is that if one gets a reconncb there is no subsequent need to perform an explicit espconn_disconnect()?

tve:
Yup. You will get an error if you call disconnect.

nkolban:
The understanding that a reconnectCB will **not** be followed by a disconnectCB needs to be handled correctly in the environment.

tve:
[...]

Issue #698, opened 2015-11-09
Dedicated module for NeoPixel support


nkolban:
The NeoPixel (aka WS2811, WS2812 and others) is a smart LED that contains an embedded controller.  It can take a one wire stream of data and change its color based on this while at the same time allowing other NeoPixels to be serially connected in a chain.   In order to drive these NeoPixels, sub microsecond timings are required in the GPIO signals in conjunction with no-latency control (there is no clock signal).  This is extremely hard to do even in native C and believed to be unlikely to be achieved in a native JS app.  Possibilities for implementing these functions include hardware SPI and other techniques.

For some boards however, (eg. ESP8266) it may make sense to provide a native C/Assembler implementation that can then be wrapped in a JS module built into the base Espruino.  Proto types of this have been tested and seem to work.   The amount of compiled code appears to be on the order of 100-200 bytes with no apparent static storage being required and hence can live exclusively in flash memory without run-time impact on precious RAM.

This Issue tracks any work items needed to provide this function as well as discuss other ramifications such as design choices for the module and "correctness" of adding ostensibly non-JavaScript function into a JavaScript runtime environment.

nkolban:
Prior to work being performed on this issue, it is vital to read and study this whole thread ...

http://forum.espruino.com/conversations/276945

There appears to be some gold nuggets kept within that should be mined, considered, designed in and utilized.

nkolban:
During a commit today, some early work on this item has "leaked" into the master build.  Thankfully it is all consistent so no harm has been done.  The file called `targets/esp8266/jswrap_esp8266.c` exposes a static method called `neopixelWrite` on the module called `ESP8266`.  The signature of this function is:

    neopixelWrite(Pin, Array)

Where Pin is the Pin to use for data transmission and Array is an array of bytes that should be supplied as triples for red, green and blue.  It is unlikely that this will be the final selection for the interface but it gives us a starting point.

gfwilliams:
Can you look at the message I added to your accidental commit? There's some info in there that would really help to clean that function uo

nkolban:
While trying to implement the recipe supplied by @gfwilliams I mistakenly left in logic that was not needed.  The logic that we wanted to achieve was the use of `JSV_GET_AS_CHAR_ARRAY` to retrieve the data.  This was indeed used... however the logic to then walk the array and retrieve the pixels one at a time was not removed.  This has now been corrected.

nkolban:
[...]

PR #705, opened 2015-11-12
#595 #698

nkolban:
Work performed against issues #595 and #698.


gfwilliams:
Looks good, thanks! You also need to do `recv`, because if Espruino doesn't have anything left to send it won't call `send` and it won't see the error

gfwilliams:
also - you know if you put `fix #595` in the commit message, GitHub magically closes the relevant issue?

Issue #711, opened 2015-11-16
Would like to add STM32F103TB ...


jcw:
I'm trying to add a definition for this [HY-TinySTM103T](http://www.hotmcu.com/stm32f103tb-arm-cortex-m3-development-board-p-222.html) board, which is an STM32F103xB in a 36-pin package. The package type I'm using in the board .py file is VFQFPN36.

First off - I get a working build when using a serial port (yeay!). Now I'm trying to get USB working. The board has a discovery pin enabling circuit on A0, so I adjusted that, as well as the LED (and no button).

With USB, the binary size is 123088 (I'm building on Mac OSX, the size check in the makefile fails with "gdu not found").

But there's something odd with the auto-generated `jspininfo` files:

```
#define JSH_PIN_COUNT 80

#define JSH_PORTA_COUNT 16
#define JSH_PORTB_COUNT 16
#define JSH_PORTC_COUNT 16
#define JSH_PORTD_COUNT 16
#define JSH_PORTE_COUNT 16
#define JSH_PORTF_COUNT 0
```

Not sure this is related, but for some reason I can't get the USB device to appear. The LED (which is on A1), *does* light up as expected on reset (it's active low - I was able to toggle in in the USART build).

I can't quite figure out what's going on - some help would be appreciated :)

-jcw

gfwilliams:
It should be doable - it's STM32 ARM - so will be making the most of all available memory. The 128kB boards have some stuff cut out, but the Olimexino has SD card support, and this won't need that.

I'll have more of a think, but stuff to check out for now:

* The package name matches a column in the CSV file referenced from the BOARD.py - that might be why you have more pins than expected
* Check your GCC version. I've seen some issues with 4.9 and USB. Try 4.8 for now and see if it makes any difference...

jcw:
Yes! - this one works (from the Arduino IDE 1.6.5's Due package):
gcc version 4.8.3 20140228 (release) [ARM/embedded-4_8-branch revision 208322]

... while this one fails (installed from https://launchpad.net/gcc-arm-embedded):
gcc version 4.9.3 20150529 (release) [ARM/embedded-4_9-branch revision 227977]

Thank you Gordon, that was a great tip. I'll investigate the package / pin mixup.

gfwilliams:
Great! Glad you got it sorted :)

Maybe we'll close this when the package/pin thing gets sorted?

PR #712, opened 2015-11-16
Add definition for HY-TinySTM103T

jcw:
This board has an STM32F103TB, i.e. 128K flash, 20K RAM - no file system support enabled, so still a little room left in flash. As with the Olimexino-STM32, this defines 1020 vars.

There's still a quirk with the generated pin file, see #711.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -217,6 +217,14 @@ STLIB=STM32F10X_MD
 PRECOMPILED_OBJS+=$(ROOT)/targetlibs/stm32f1/lib/startup_stm32f10x_md.o
 OPTIMIZEFLAGS+=-Os # short on program memory
 
+else ifdef HYTINY_STM103T
+EMBEDDED=1
+SAVE_ON_FLASH=1
+BOARD=HYTINY_STM103T
+STLIB=STM32F10X_MD
+PRECOMPILED_OBJS+=$(ROOT)/targetlibs/stm32f1/lib/startup_stm32f10x_md.o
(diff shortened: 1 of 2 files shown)


gfwilliams:
Great, thanks! Any chance you could get a top-down image and have a fiddle with the CSS so that we can have a nice page that describes the board? A bit like  http://www.espruino.com/ReferenceSTM32VLDISCOVERY

You'd need to run `scripts/builkd_board_docs.py HYTINY_STM103T`

jcw:
Ok, I've got something - how do you spread out the pins (or reduce the image scale?

screen shot 2015-11-18 at 11 46 34

jcw:
Suggestion: support a simpler notation as well, i.e.

    'bottom' : [ '3.3','A9','A10','SWDIO','SWCLK','GND' ],

could also be written as follows, if you split on spaces when the entry is a string?

    'bottom' : '3.3 A9 A10 SWDIO SWCLK GND',

Gets rid of some of that line noise ...

gfwilliams:
Looks great! Add:

```
.leftpin { height: 32px; }
.rightpin { height: 32px; }
```

to the CSS - you probably want something a bit less than 32 though!

I think those lines are read in a few different places so it makes sense not to have to get each place to do it. I guess you could write `' .... '.split(' ')` - but it's not too hard to enter it.

Worst case, highlight the line and do a replace of ` ` with `','` :)

PR #713, opened 2015-11-17
Fix VFQFPN-36 package

### LINKED ISSUE #711: Would like to add STM32F103TB ...

jcw:
closes #711


jcw:
Actually, that missing column should be called "LQFP100", according to the datasheet.

gfwilliams:
Ok, thanks! I'll add that then. Might be useful :)

PR #716, opened 2015-11-18
Make size detection cross-platform

jcw:
this gets rid of the "gdu" dependency, and simplifies the logic


gfwilliams:
Neat - thanks!

PR #717, opened 2015-11-18
Add graphics to HY-Tiny board

jcw:
It barely fits, 122540 of 124928 bytes now used, but still makes sense
because of the SPI-LCD connector on the bottom of this board (confirmed
to work, using OLED via I2C).


gfwilliams:
Thanks!

PR #718, opened 2015-11-19
Clear graphics screen on startup

jcw:
The HY-STM32-3.2" was starting up with garbage on-screen. There may be a better way to clear it (I think it's a ILI9331, but I don't know anything about the init sequence).


jcw:
No idea why the Pico build became too large and failed. I only added a graphics "clear" call, which is present in the build anyway, AFAICT.

gfwilliams:
It was my fault - it'll be fine now :)

PR #719, opened 2015-11-21
Unify the two esp8266 boards and add to travis builds

tve:
This PR has a bunch of changes:
- unifies the two esp8266 boards (I have not checked how the EspruinoDocs site comes out)
- minor OneWire fix to provide active power to parasitic devices
- adds esp8266 to travis builds and supports uploads to S3 (conditional, so won't affect "official" builds)


Makefile
```diff
@@ -103,12 +94,18 @@ ifdef RELEASE
 DEFINES += -DNO_ASSERT -DRELEASE
 endif
 
-LATEST_RELEASE=$(shell git tag | grep RELEASE_ | sort | tail -1)
-# use egrep to count lines instead of wc to avoid whitespace error on Mac
-COMMITS_SINCE_RELEASE=$(shell git log --oneline $(LATEST_RELEASE)..HEAD | egrep -c .)
-ifneq ($(COMMITS_SINCE_RELEASE),0)
-DEFINES += -DBUILDNUMBER=\"$(COMMITS_SINCE_RELEASE)\"
+LATEST_RELEASE = $(shell egrep "define JS_VERSION .*\"$$" src/jsutils.h | egrep -o '[0-9]v[0-9]+')
+ifdef RELEASE_LABEL
```
gfwilliams: This looks like it'll remove the commits part by default? Was there a reason for this?

It got put in because I got badgered by the community about it - it got really confusing having lots of images all of which said `1v81`.

I don't mind having the binary with the same name, but when it boots, it should really display the version and subversion...

gfwilliams:
Looks great - thanks! Please could you change the build number thing though? I'm pretty sure it'll just mess up my build arrangements if it changes back to the way it was before.

Makefile
tve: So I changed LATEST_RELEASE because it didn't work due to the fact that travis does a shallow clone and the release tags are too far back. Since what Espruino prints out at boot comes from a #define in jsutil.h I figured the Makefile might as well get it from there too.

If RELEASE_LABEL is not defined, which is the default, then the SUB_RELEASE is exactly what it was before, but it still suffers from the problem that the travis clone may not have the release tag. This is the case currently. My vote would be to switch from counting commits to putting the date in there, something like YYYYMMDD_HHMM. Alternatively we could use date and SHA, like YYYYMMDD_SSSSSS where the S's are the first 7 letters of the commit SHA.

If RELEASE_LABEL is defined, which is not default, then you get $(RELEASE_LABEL)_$(BRANCH_NAME). I set RELEASE_LABEL to "tve" in a travis env variable on my repository, this way a build of mine comes out saying something like "v1.81_tve_unify". I'll add the first few letters of the commit SHA, I think.

tve:
I reverted the release labeling. You do realize it doesn't work, right? My builds now come out with 1v81_502 even though there have been only 263 commits since release, the very last build has _496, yesterday the builds came out with 1v80 because the travis clone didn't include the RELEASE_1V81 tag. A couple of weeks ago nkolban had the same issue and people that were downloading his builds were confused because later builds had lower numbers, etc.

I added an ALT_RELEASE variable to the Makefile so I can have build numbers that are not messed-up and include 'tve', branch name, and SHA. IMHO the release labeling is a mess, it took me a while to discover that a similar algorithm lives in python as well... I'm not sure I fully like the ALT_RELEASE scheme either, but I'll leave it as-is for now until I get auto-upgrade into the esp8266 build...

gfwilliams:
Hmm, do you know why it breaks? I know it doesn't cope well with branches. I'm surprised travis didn't have all the tags though?

To be honest as long as the builds that come out on espruino.com have some kind of rising version number (regardless of what it is) it's probably more useful having it than not...

PR #721, opened 2015-11-22
enhanced navigation in reference page

muet:
- added backlinks from detail headers to (table of) content entries
- enhanced #top link to select searchbox entry and focus on it
- made navigation jumps (former anchorlinks) to postion at 20% of view port
- made completed navigation jump to blush blue on back and gold on forward

- added htmldev development mode flag to develop and test in standalone mode
  (set temp htmldev to True, run script, and open generated html in browser)

Changes to be committed:
	modified:   scripts/build_docs.py


gfwilliams:
Thanks! That looks neat, but is there any way of doing this while leaving the links to the anchors in place?

Problem is, it's handy to go to http://www.espruino.com/Reference, click on some function, and then copy the URL and paste it somewhere - either into a tutorial, forum post, etc, to go straight to the function - [like this](http://www.espruino.com/Reference#l__global_digitalRead).

At the moment you can't do that, not even by right-clicking on the link. Maybe if the links from the contents page used anchors, but the headings used your JS code to go to back to the contents page?

I'm not sure how you'd auto-highlight the heading on the actual function, but there must be a way... That one isn't that big a deal though, as it should always be at the top of the page because of the anchor.

gfwilliams:
Thanks!

PR #724, opened 2015-11-24
Move esp8266 string constants to flash

tve:
Moving string constants to flash saves ~10KB of RAM, which is very significant. The changes are a bit more intrusive than I would have liked, but it's not too bad either. I hope the ifdefs don't make things too unreadable. If you have suggestions I'm fine changing stuff around.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ after_script:
   - ls *.bin *.hex *.tgz | xargs -I {} curl -v -F "binary=@{}" "http://www.espruino.com/travis_upload.php?commit=$TRAVIS_COMMIT"
   # upload to an S3 bucket, requires S3_BUCKET, AWS_ACCESS_KEY_ID and AWS_SECRET_KEY to be set
   # in environment using travis' repository settings
-  - "if [[ -n \"$S3_BUCKET\" && -n \"$AWS_ACCESS_KEY_ID\" ]]; then 
+  - "if [[ -n \"$S3_BUCKET\" && -n \"$AWS_ACCESS_KEY_ID\" ]]; then
       echo Uploading *.tgz to $S3_BUCKET;
       curl -Ls https://github.com/rlmcpherson/s3gof3r/releases/download/v0.5.0/gof3r_0.5.0_linux_amd64.tar.gz | tar zxf - gof3r_0.5.0_linux_amd64/gof3r;
       mv gof3r*/gof3r .;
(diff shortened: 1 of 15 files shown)


gfwilliams:
Wow, thanks. That's really nicely done. About as unintrusive as I could imagine it being. The biggest changes were when using `jsiConsolePrint` to `jsiConsolePrintString` :)

There should still be some big strings in `jswrapper.c` and maybe `jslex.c` that could potentially be moved over too (although I have some plans to change jswrapper a lot soon, so it might be worth holding off on that).

tve:
The output of `DISABLE_LTO=1 make topstrings` is:
```
Top 20 from ./topstrings:
13691 ./espruino_esp8266_partial.o [total strings]
1387 ./libs/network/esp8266/jswrap_esp8266_network.o
1083 ./src/jsparse.o
885 ./src/jsinteractive.o
601 ./gen/jswrapper.o
590 ./src/jswrap_espruino.o
587 ./libs/network/socketserver.o
542 ./src/jswrap_io.o
513 ./src/jswrap_object.o
486 ./src/jsvar.o
481 ./src/jswrap_pipe.o
435 ./libs/network/esp8266/network_esp8266.o
422 ./libs/network/esp8266/ota.o
392 ./src/jswrap_waveform.o
367 ./src/jswrap_array.o
363 ./src/jslex.o
317 ./src/jswrap_interactive.o
316 ./src/jswrap_arraybuffer.o
269 ./src/jswrap_spi_i2c.o
263 ./src/jswrap_json.o
```
I'll knock down the stuff in jswrap_esp8266_network when I work on that. Otherwise I have the feeling that focusing on utility functions that chew up strings across the board is more promising than focusing on individual files. Also, right now there's a good amount of RAM available, so I feel less urge :-). We shall see how this evolves...

gfwilliams:
Ahh, someone will come along and increase the number of variables in Espruino, then all your RAM will be gone :)

I find it hard to believe there are only 600 bytes worth in jswrapper.c, have you seen it? Or is it that constant arrays don't go in RAM?

My quick hack on the Espruino Pico shows:

```
grep "[0-9]* g.*jswSymbol" espruino_1v81.307_pico_1r3.lst | awk '{print $5}' | perl -nle '$sum += hex($_) } END { print $sum'

3790 
```

So that's bytes used just for the symbol tables, and *excluding strings*. I'd imagine it's nearer 5k of read only data in there.

Issue #725, opened 2015-11-24
Make flash baud rate for ESP8266 configurable


nkolban:
The Makefile / build system for flashing the ESP8266 is currently hard-coded with a baud rate of 460800.  Unfortunately that is too high for some systems.  This request asks for this to be exposed externally as a parameter to the Makefile ... for example, a property called "BAUD" with a default of 115200.

gfwilliams:
The Makefile is pretty huge already - adding specific things for stuff like that does seem like overkill.

Why not set the default to 115200, and if someone wants more they can just modify the makefile themselves.

Honestly, searching for 115200 in the file and replacing it with 460800 is trivial compared with wading through documentation or the file trying to find out which environment variable you need to set to get the correct baud rate.

nkolban:
The default baud rate is now 115200 but can be optionally overriden with a FLASH_BAUD environment variable.

gfwilliams:
Sometimes I wonder why I bother writing anything when you post these bugs.

nkolban:
Look ... don't be vague ... say what you mean ... What have you written?  Where?  What is wrong with posting bugs? ... don't be vague ... we are all professional programmers here ... what is it you want to say?

gfwilliams:
Right after you posted the above, I posted the second comment - I thought it was pretty clear - I thought adding more options to an already complicated makefile was a bad move, and that anyone smart enough to figure out 115200 was too slow could open up the makefile, search for 115200 and replace it.

But 1 hour later, no discussion, but a commit that just did what you posted originally as if you hadn't seen my post at all. Hence my comment. It doesn't matter now, it's just a few lines - but it seems to happen quite a lot...

For instance just today you added the neopixel isArrayBuffer changes, but we spent half an hour on Gitter a few days ago talking about how get_char_array did all the checks so you didn't need any (and previous to that I also made comments to that effect on the original neopixel PR, [where I even wrote down the exact code you could use in my comment](https://github.com/espruino/Espruino/pull/692#issuecomment-155493451)).

I guess you probably have so many emails bouncing around you miss some - do you get emails when I post a reply to one of your bugs/PRs? It seems sometimes you might not. Or maybe you just scan-read and don't really take in what I'm saying - after all I know you're doing this in your free time, it's not like you have to do it, or any of the good work you've done on ESP8266 to date.

[...]

Issue #722, opened 2015-11-22
ESP8266 getIP() only returns station IP


nkolban:
The function called getIP() only returns the station IP address.  This is a problem because the ESP8266 can also be an access point.  The suggestions here is to provide an optional argument to the function which can either be "station" or "accesspoint" to determine which IP address to return.  The default if not specified should be "station".

martinayotte:
In ArduinoESP, to distinguish both IPs, the functions are named localIP() and softAPIP().

gfwilliams:
Odd - I thought there had been some discussion about this before, but I can't find it now - I think rakeshpai mentioned moving to a `networkInterfaces` object, but that does seem a bit overcomplicated.

I guess another option is to have one as the default, but to return both, so:

```
{ 
  ip : ...,
  mac : ...,
  ap : { ip : ... },
  station : { ip : ... }
}
```

But to be honest the other solutions are good as well. I guess there could be a `getAPIP` or something that only ESP8266 devices implemented.


PS. In the Wiki where we discussed getIP's implementation, it used a callback so that it could be compatible with other network interface types. For instance with AT commands (GSM, ESP8266 module) you don't want to block while waiting for serial data to be returned. Same would apply for CC3000 too, if TI's drivers weren't so bad! :)

profra:
If station and AP are working both there are 2 IPs and 2 MACs... hence it would be more clear to ask them with two different functions... see reaction on AT command AT+CIFSR... picture...
https://espruino.microco.sm/api/v1/files/b4b12a58ea4823ec6fadd16911b7121fcf4770bf.png

nkolban:
Added a mac address to the returned data from getting the IP address.  Added a second function called getAPIP() that gets the Access Point information.

nkolban:
The documentation update for the new `getAPIP()` function as well as the documentation that the mac address is now included in the results was omitted.  This needs to be done before this issue should be closed.

nkolban:
There appears to be a function called "networkPutAddressAsString" that formats a Mac address ... this looks like exactly what we need to reuse in this area ... see the following forum post for more details....

http://forum.espruino.com/conversations/278075/#comment12646517

This needs added.

nkolban:
The docs were updated to include getAPIP.  In addition, the construction of the mac address string has now been changed to use the Espruino supplied string constructor.

trusktr:
I think the Espruino WiFi docs are outdated? But getIP does give IP when not in AP mode.

[...]

PR #726, opened 2015-11-24
#722 #725

nkolban:
Implementation of issues #722 and #725


libs/network/esp8266/jswrap_esp8266_network.c
```diff
@@ -765,18 +787,31 @@ JsVar *jswrap_ESP8266_wifi_getStatus() {
...
 * * netmask - The interface netmask
 * * gw - The network gateway
+* * mac - The MAC address
 */
 JsVar *jswrap_ESP8266_wifi_getIP() {
   os_printf("> jswrap_ESP8266_wifi_getIP\n");
-  struct ip_info info;
-  wifi_get_ip_info(0, &info);
-
-  JsVar *jsIpInfo = jspNewObject(NULL, "Restart");
```
gfwilliams: I'm not sure this is doing what you want at all (or if it is, it's accidental). Just use `= jsvNewWithFlags(JSV_OBJECT)` like everything else does.

libs/network/esp8266/jswrap_esp8266_network.c
```diff
@@ -753,6 +753,28 @@ JsVar *jswrap_ESP8266_wifi_getStatus() {
...
+/**
+ * Get the ip info for the given interface.  The interfaces are:
+ * * 0 - Station
+ * * 1 - Access Point
+ */
+static JsVar *getIPInfo(int interface) {
+  struct ip_info info;
+  wifi_get_ip_info(interface, &info);
+
+  JsVar *jsIpInfo = jspNewObject(NULL, "IPInfo");
```
gfwilliams: Oops. this is the new code. I guess you haven't defined a class called `IPInfo` for this, and even if you did, it'd be wasting some RAM.

Just use `= jsvNewWithFlags(JSV_OBJECT)` like everything else does.

PR #730, opened 2015-11-25
reinstating href anchor for all reference links

muet:
Back to the roots - but keeping the enhancements: copy link location as well as browser address work.

References, such as print in console, stay just plain href anchors.

Changes to be committed:
	modified:   scripts/build_docs.py


gfwilliams:
Great, thanks!

PR #734, opened 2015-11-26
#722 Updated to used improved MAC address string construction.

nkolban:


libs/network/network.c
```diff
@@ -108,6 +119,17 @@ JsVar *networkGetAddressAsString(unsigned char *ip, int nBytes, unsigned int bas
...
+ * byte is separated by a separator character.  The numeric base of the bytes is given by the base
+ * value.
+ *
+ * For example, to create a dotted decimal string, one would use:
+ * networkPutAddressAsString(myObject,"ip", ip, 4, 10, '.')
+ *
+ * To create a Mac address, one might use:
+ * networkPutAddressAsString(myObject, "mac", mac, 6, 16, ':')
+ */
 void networkPutAddressAsString(JsVar *object, const char *name,  unsigned char *ip, int nBytes, unsigned int base, char separator) {
```
gfwilliams: You've done it again! There's documentation in the header file already!

Please can you copy your updated documentation into the header, and update it with the (small amount of) extra info that was in the header?

```
/// if nBytes<0, addresses are printed out backwards
JsVar *networkGetAddressAsString(unsigned char *ip, int nBytes, unsigned int base, char separator);

/// Given an address (pointed to by ip) put it in a string named 'name', in the given object. if nBytes<0, addresses are printed out backwards
void networkPutAddressAsString(JsVar *object, const char *name,  unsigned char *ip, int nBytes, unsigned int base, char separator);
```

gfwilliams:
Thanks!

PR #744, opened 2015-11-30
Device enum fix

lukevanhorn:


lukevanhorn:
File system / sd card access stopped working after commit a35b05c4a9f31d13aa8ee9b4bd4ba86c82760c0f.  This fixes that issue, but the device type enumeration may need a closer look.

gfwilliams:
Thanks - I'm actually really struggling with this.

SPI still appears to work fine in the old versions, and in fact when I add some debug statements to `jsspiGetSendFunction` it magically starts working - which makes me think it could be a timing issue more than anything else.

But hey, this PR does seem to fix it :)

lukevanhorn:
Definitely a symptom fix, rather than the root cause. I noticed that commenting out a few of the new serial status values also worked.  So, it seems to hit a threshold.

gfwilliams:
Thanks! Actually that's probably what it is. Espruino expects to fit IOEventFlags into a byte, and I'd read the bit about 64 and thought I had 6 bits to play with. I don't - looks like there's only 5, so ideally only 32 entries in the deviec part (and there are more).

I'll open an issue to look into it.

PR #745, opened 2015-12-01
set default values for mem & cpus

AlexanderBrevig:
Currently, following https://github.com/espruino/Espruino/blob/master/README_Building.md#building-under-windowsmacos-with-a-vm-vagrant will lead to error on windows
Adding default values for cpus and mem fixes this

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@
 *.bin
 *.pyc
 gen/*
+.vagrant/*
 boards/*.html
 boards/*.json
 zipcontents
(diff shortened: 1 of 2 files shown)


gfwilliams:
Thanks! Wow, it amazes me how they can keep breaking something that's supposed to make installs easy :)

AlexanderBrevig:
I must say though... I've been dual booting for a while now. But using a VM is sooo much better! Especially since I cannot get the st-flash to work on Linux...

PR #746, opened 2015-12-01
add script for syncing forks

AlexanderBrevig:


gfwilliams:
Was this needed for vagrant, or it's just a useful script?

I know I'm being picky, but is there a reason this isn't a bash script?

AlexanderBrevig:
it's got nothing to do with the vagrant stuff so I did separate PR
I did not get git from within the VM to work well when one of the remotes is on LAN and before I used a VM I still would do a lot of the work from the windows OS. Python made it possible for me to use same workflow on both OSes.

I understand if you want it to be .sh instead as Linux is needed to build Espruino (but not the docs or .md which is my argument for keeping it python so win folk can contribute to those even though they cant build Espruino itself)
:)

gfwilliams:
Thanks... I just assumed that anyone doing stuff on windows was using git shell or something :)

AlexanderBrevig:
to be fair, you're probably right  :)
glad to see it accepted! less copying...hehe

from mobile

PR #747, opened 2015-12-01
make vagrant up pull arm-none-eabi-gcc v4.8.4 and set it to PATH

AlexanderBrevig:
this will make it possible to do

`vagrant up && vagrant ssh && scripts/create_espruino_image_1v3.sh` and it chooses the correct compiler

### DIFF
--- a/scripts/provision.sh
+++ b/scripts/provision.sh
@@ -2,8 +2,26 @@
 #
 # This script is run by Vagrant when a new machine is provisioned.
 #
+ 
+sudo apt-get -y install git
+# add support for 64 bit (which vagrant is set to configure)
+sudo apt-get -y install lib32z1 lib32ncurses5 lib32bz2-1.0
+sudo apt-get -y install lib32ncurses5
+sudo apt-get -y install ia32-libs
+ 
+# download known working version of arm-none-eabi-gcc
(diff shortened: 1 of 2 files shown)


scripts/provision.sh
```diff
@@ -2,8 +2,26 @@
 #
 # This script is run by Vagrant when a new machine is provisioned.
 #
+ 
+sudo apt-get -y install git
+# add support for 64 bit (which vagrant is set to configure)
+sudo apt-get -y install lib32z1 lib32ncurses5 lib32bz2-1.0
```
gfwilliams: would this break on 32 bit systems? Maybe:

```
ARCH=$(uname -m)
if [ $ARCH = "x86_64" ]; then
 sudo apt-get -y install lib32z1 lib32ncurses5 lib32bz2-1.0
 sudo apt-get -y install lib32ncurses5
 sudo apt-get -y install ia32-libs
fi
```

AlexanderBrevig: don't you force 64 bit https://github.com/espruino/Espruino/blob/master/Vagrantfile#L7 ?

gfwilliams: Ahh, didn't realise

PR #749, opened 2015-12-01
add option to read tls cert files from file system

lukevanhorn:
If the option.ca / cert / key values are less than 100 characters and a file system exists, this will attempt to load from the specified path.


gfwilliams:
Looks great, thanks!

PR #755, opened 2015-12-07
Fix sockets

tve:
This pull request includes the following changes for the esp8266:
- move to Espressif SDK 1.5
- fixes in wifi library (save/restore, on-events, ...)
- async DNS lookup in socket.connect() call
- clean-up of sockets code in particular to handle errors and closing without leaks
- fix some soft reset issues
- change ICACHE_RAM_ATTR to CALLED_FROM_INTERRUPT
- improve neopixel code, move other code out of sram to make space


gfwilliams:
Great, thanks!

PR #762, opened 2015-12-11
Improve socket error events

tve:
This PR has a couple of changesets. The largest is to implement correct error handling in the esp8266 socket code. In addition:
- socketserver now keeps track of error codes and calls the error event with an error object
- the connect callback for raw client sockets is only called when a connection is actually established
- the drain callback is called when a connection is established and not only after the first write completes
- getSerial() on the esp8266 now makes more sense

I tried to ensure that the socketserver changes are compatible with other network implementations, but I don't have the HW to test. I did run the linux test suite and the http tests pass.


libs/network/socketserver.c
```diff
@@ -334,13 +338,23 @@ bool socketServerConnectionsIdle(JsNetwork *net) {
...
             closeConnectionNow = true;
+            error = sent;
+          }
         jsvObjectSetChild(socket, HTTP_NAME_SEND_DATA, sendData); // socketSendData prob updated sendData
       }
       // only close if we want to close, have no data to send, and aren't receiving data
       if (jsvGetBoolAndUnLock(jsvObjectGetChild(socket,HTTP_NAME_CLOSE,0)) && !sendData && num<=0)
         closeConnectionNow = true;
+      else if (num > 0)
+        closeConnectionNow = false; // guarantee that anything received is processed
```
gfwilliams: Isn't this already done on line 525 with:

```
if (closeConnectionNow) {
 socketClientPushReceiveData(connection, socket, &receiveData);	
 if (!receiveData) {
  // ...
```

libs/network/socketserver.c
```diff
@@ -352,9 +366,25 @@ bool socketServerConnectionsIdle(JsNetwork *net) {
         jswrap_stream_pushData(connection, receiveData, true);
       }
       jsvUnLock(receiveData);
+
+      // fire the error listeners
+      bool hadError = error < 0 && error != SOCKET_ERR_CLOSED;
+      JsVar *params[1];
+      if (hadError) {
+        params[0] = jsvNewWithFlags(JSV_OBJECT);
```
gfwilliams: Could this be a separate function? It'd reduce duplication with the other Error code below, and would make it easier to tweak later on to print an error message if there was no handler.

gfwilliams:
Looks good, but because you're treating `-1` as not an error in all cases, it looks like you've broken error handing on all the other platforms now?

```
require("http").get("http://localhost:1234", function(res) {
  res.on('data', function(data) { console.log(">" + data); });
}).on("error", function(e) {
  console.log("GOT ERROR: "+e);
});
```

So something like this should give an error:
 * if there's no server at 1234
 * If you do `nc -l 1234` to listen on 1234, but then close it after Espruino has connected but before you've replied with any headers.

It fails on Linux so should be easy enough to debug.

I know the error handling was new(ish) but I don't have time to now fix it for all the other platforms, and I'm not really sure that should be my job anyway.

Any chance you could take out the `socketserver.c` changes for now, just so we can have something put in that doesn't break what's there already?

libs/network/socketserver.c
tve: Makes sense.

libs/network/socketserver.c
tve: Mhh, sorry, I'm drawing a blank here. You added this comment to line 356-357, which is in the server socket handling and you mention line 525 which is in the client socket handling.

tve:
> Looks good, but because you're treating -1 as not an error in all cases, it looks like you've broken error handing on all the other platforms now?

I didn't think so, but I'll check. I had toyed with the idea of inserting #ifdef ESP8266 around a small number of the error handling in order not to break other targets, but I thought I could get away without.

gfwilliams:
Thanks!

IMO just changing the 'default' return value from -1 to something like ERR_IS_BUSY (-0x80000000?) on ESP8266 would fix it - so on other targets it's still treated as an error.

Sorry for making the merge conflict too. I'm happy to fix that myself when it comes to merge time.

libs/network/socketserver.c
gfwilliams: Oops. Sorry - my bad. Looks fine then. I know it's not a great solution at the moment :(

tve:
I started working on fixing the error handling but haven't gotten very far due to lack of time over the past couple of days.

tve:
I moved the error event generation into a separate function in a95ff120491fc024f34e6bfc9f4eafd8afd95f67

I had a bug in the error generation that I fixed: a1847d2fcb121300e6a5d128f5eaa71b3d94504f line 494 of socketserver.c. Basically what that does is turn any error -1, which is a normal close on esp8266 or close-or-error on other platforms into an error if we have not been able to parse response headers. That's the only way I can see to get an error on other platforms.
I now notice that maybe I should turn that into a "no response" error instead of "unknown error", although that's really saying more than the other platforms really indicate.
Stuff like connecting to a non-existent port (i.e. connection reset at connection time) turn into JsError in the linux networking code, so none of this really comes into play as far as I can see.

tve:
OK. At this point I believe I added error handling and I didn't break anything. I tested the linux version as much as I could think of and it looks OK to me, but a second set of eyes or tests would be "a good idea".
I can commit to upgrading the linux version to produce proper errors like the esp8266. I'll see whether I have energy to do the pico+esp8266, I first have to get the pico to work (windows 7 didn't automatically install the right drivers) and then solder one of my used esp-01 modules onto it. I'll leave it up to you whether you feel OK merging this in, see specific issues I should first fix, or want the PR to just hang there until I get some of the above done. Just let me know ;-)

gfwilliams:
Thanks! Looks good. Am a bit busy today, but will merge anyway. If anything looks broken I'll just file an issue for it.

PR #763, opened 2015-12-14
Refactoring conditional directives for alternative if statements.

wmarquesr:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

* [Linux kernel coding style](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892)
* [Living in the #ifdef Hell](https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/)

It might improve code understanding, maintainability and error-proneness.

### DIFF
--- a/targetlibs/nrf5x/nrf51_sdk/components/serialization/common/transport/ser_phy/ser_phy_hci.c
+++ b/targetlibs/nrf5x/nrf51_sdk/components/serialization/common/transport/ser_phy/ser_phy_hci.c
@@ -1224,16 +1224,18 @@ static void hci_tx_reschedule()
     bool     tx_exec_flag = false;
     uint32_t tx_queue_length;
     uint32_t err_code;
+    int flags_status;
 
     CRITICAL_REGION_ENTER();
     err_code = app_mailbox_get_length(tx_evt_queue_id, &tx_queue_length);
     ser_phy_hci_assert(err_code == NRF_SUCCESS);
 
 #ifndef HCI_LINK_CONTROL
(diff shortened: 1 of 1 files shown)


gfwilliams:
I'm not sure I really understand why this was changed? Why this code in particular, was it causing you problems?

Sure, I agree the new code is marginally nicer, but this is part of the Nordic SDK. Compared to the minor code change, having to maintain a modified version of an SDK that *will change* in a few months sounds like a disaster.

Maybe you should submit this (and other similar changes) to Nordic themselves, who maintain the SDK. That way you'll get your changes into  lot more projects.

wmarquesr:
The idea is always to encompass only complete/entire statements/expressions with preprocessor conditional directives. It is something that is getting popular in many different projects (started mainly on the Linux Kernel).

We agree totally that breaking statements with directives influences code quality, i.e., understanding, maintainability and error proneness. We found some evidence based on interviews with developers and an online survey regarding this negative effect on quality.

[http://www.cs.cmu.edu/~ckaestne/pdf/ecoop15.pdf](http://www.cs.cmu.edu/~ckaestne/pdf/ecoop15.pdf)

What you think about that?

gfwilliams:
I think that didn't really answer my main point about this being part of Nordic's SDK, which will end up getting updated in the next month or two anyway.

To me, the risk of maintaining a patch set over the SDK far outweighs the small improvement in readability coming from this change,

I'm going to close this, but I'd suggest that you contact Nordic themselves and suggest these code changes to them. Then your ideas will get out to a lot more people.

PR #764, opened 2015-12-14
Remove a ton of uneccesary files from nRF5x SDKs. Still more can be r…

mjdietzx:
Sorry for including these in the first place. More can be removed from components and I will look into that more.


mjdietzx:
also nRF51 and 52 sdks are supposed to be merged into one soon so I'll keep my eye on that

gfwilliams:
That's awesome - thanks! Should make life a lot nicer!

When the SDKs merge that'd be great too - does that mean that the IoT SDK will then be available for nRF52?

mjdietzx:
by the way I only tested this on nRF51 since after merging changes from nrf52 from espruino/master it wasnt compiling. i really dont think these changes will break anything in nrf52 anyways. gonna look into why its not compiling and try to clean stuff up for nrf5x devices

mjdietzx:
np! I need to get an update about when the SDKs merge it was scheduled to be within a month or so from now. and I need to double check about the IoT SDK but I did notice mbed TLS in espruino which could be very promising

gfwilliams:
Thanks - sorry about the nRF52 compile issue. Seems to be something ADC-related - the nRF52 SDK has `#ifndef NRF52` around all the ADC stuff in `nrf_adc.h` which seems a bit strange??

Worst case you could just ifndef the ADC stuff in jshardware.c.

PR #768, opened 2015-12-18
Minor changes: lwip_536 and docs stuff

tve:
- Switch lwip to a TCP MSS of 536 now that Espressif provided a build for that
- Improve http library documentation
- Remove stale esp8266 docs


gfwilliams:
Thanks! Could any of that documentation have been used?

With stuff like `60-The-Espruino-API.md` is pretty much impossible to see if any documentation has been added over what's in the normal reference. At least it hasn't been touched in the last few months which I guess is a good sign.

At least it's all still in the Git History :)

PR #784, opened 2016-01-14
Nrf5x shared sdk

mjdietzx:
-Added new Nordic SDK that is shared by both nRF51 & nRF52.
-nRF51 now uses S130 instead of S110 softdevice. Updated linkers to support this and using softdevice hex files directly from the sdk instead of a separate folder in targetlibs/nrf5x.
-Slight changes to jswarp_bluetooth.c based on more up to date SDK.
-Removed files from makefile that were being compiled that aren't needed like (fifo.c, app_uart_fifo.c, retarget.c, etc...) and also changed some defines that are needed in new sdk.

Issues: I had to stub out all ADC & TWI in jshardware.c. Also had to stub out get_battery in jswrap_bluetooth.c because it uses ADC. Had to comment out timer stuff in jshinit() because it causes UART comm with terminal to break for some reason.
-I think these are differences in SDK since they wouldnt compile for the nRF52 SDK. Should be an easy fix but havent figured it out yet. Want to make the PR now because its getting really big and so you can see whats going on.


gfwilliams:
Great! Was there some kind of merge problem though? Looks quite a few of the changes in jswrap_bluetooth are actually undoing code formatting tweaks that I'd done over the last few days...

All of Espruino is in K&R (opening bracket on the end of the line) and 2 spaces for indents so I'd been trying to go through and bring the nRF stuff in line with that.

I'll see how easy it is to fix TWI and the ADC. I literally only got them working this week, so it really sucks if they've just broken again - I'd rather not pull into the main branch until that's working.

gfwilliams:
I also seem to get:

```
IOError: [Errno 2] No such file or directory: '/home/gw/workspace/Espruino/targetlibs/nrf5x/nrf5_sdk/components/softdevice/s130/hex/s130_nrf51_2.0.0-7.alpha_softdevice.hex
```

In fact there is no `hex` dir there at all - did you forget to commit those files? I guess they might have got covered by the `.gitignore`.

Sorry about the recent conflict - I was just pushing my changes. Easy enough to fix though :)

mjdietzx:
Oh shoot I didn't know about K&R. I'll fix that. My goal is to fix all these problems (TWI, ADC, Timer) by tomorrow. Just wanted to make a PR so you know where I'm at and can decide to pull now or after I've made the fixes. I guess they got covered by the .gitignore. The softdevice hex files are there in the sdk and on my local copy

gfwilliams:
That's great! Please could you try and add in those hex files? I believe you can use `git add -f` even though they're in gitignore.

mjdietzx:
yeah i'll try that now

gfwilliams:
I wonder if changing `.gitignore` to do `./*.hex` would do it?

gfwilliams:
Just fixed ADC and pushed this - seems to work ok now :)

PR #786, opened 2016-01-15
remove stub returns for adc stuff.

mjdietzx:
forgot to remove these returns and uncomment battery in ble lib


gfwilliams:
Thanks!

PR #787, opened 2016-01-18
Merge gobs of esp8266 fixes/improvements back to mainline

tve:
This merges all the changes I've made for the esp8266 over the past month back into mainline. My last build has been very stable for me. Unless the merge introduces instability, I think this is a good v1.85 candidate from the esp8266 point of view.


libs/network/socketserver.c
```diff
@@ -487,7 +485,7 @@ bool socketClientConnectionsIdle(JsNetwork *net) {
         }
         // Now read data if possible (and we have space for it)
         if (!receiveData || !hadHeaders) {
-          int num = netRecv(net, sckt, buf, net->chunkSize);
+          int num = netRecv(net, sckt, buf, sizeof(buf));
```
gfwilliams: I think this is a merge error? It should definitely be chunkSize now, as `buf` is a `char*`

src/jswrap_flash.c
```diff
@@ -114,8 +116,10 @@ doesn't return all `0xFF`) you'll need to call `erasePage` to clear the
 entire page.
  */
 void jswrap_flash_write(JsVar *data, int addr) {
-  if (jsvIsUndefined(data))
-    return jsExceptionHere(JSET_ERROR, "Data is not defined");
+  if (jsvIsUndefined(data)) {
+    jsExceptionHere(JSET_ERROR, "Data is not defined");
```
gfwilliams: Does this really matter, or is it a personal code style thing? I've been doing it a lot as it seems a bit more concise.

src/jswrap_interactive.c
```diff
@@ -507,7 +508,10 @@ void _jswrap_interface_clearTimeoutOrInterval(JsVar *idVar, bool isTimeout) {
       jsvRemoveChild(timerArrayPtr, child);
       jsvUnLock2(child, timerArrayPtr);
     } else {
-      jsExceptionHere(JSET_ERROR, isTimeout ? "Unknown Timeout" : "Unknown Interval");
+      if (isTimeout)
+        jsExceptionHere(JSET_ERROR, "Unknown Timeout");
```
gfwilliams: Again, just personal code style? I assume the original option probably uses marginally less flash memory

src/jswrap_onewire.c
```diff
@@ -132,7 +132,7 @@ bool jswrap_onewire_reset(JsVar *parent) {
     ["rom","JsVar","The device to select (get this using `OneWire.search()`)"]
   ]
 }
-Select a ROM - reset needs to be done first
+Select a ROM - always performs a reset first
```
gfwilliams: Does this mean that modules that use OneWire (DS18B20) should be modified?

targets/esp8266/jshardware.c
```diff
@@ -692,8 +666,19 @@ bool jshIsUSBSERIALConnected() {
...
 void jshUSARTKick(
     IOEventFlags device //!< The device to be kicked.
-  ) {
-  esp8266_uartTransmitAll(device);
+) {
+  // transmit anything that is sitting in the uart output buffers
+  if (device == EV_SERIAL1 || device == EV_SERIAL2) {
+    // Get the next character to transmit.  We will have reached the end when
+    // the value of the character to transmit is -1.
+    int c = jshGetCharToTransmit(device);
```
gfwilliams: There does actually appear to be some UART TX empty interrupt [like this](https://github.com/espressif/ESP8266_RTOS_SDK/blob/master/examples/driver_lib/driver/uart.c#L390) but I don't know if it's exposed at all? If so, that might really help here?

gfwilliams:
Actually I'll merge this anyway and will fix the socketserver.c issue

libs/network/socketserver.c
tve: gnash, my bad :-(

src/jswrap_flash.c
tve: I found it confusing 'cause jsExceptionHere is not returning anything and I was changing jsExceptionHere. But you're right, I shouldn't have gratuitously modified it.

src/jswrap_interactive.c
tve: I modified jsExcpetionHere to a macro that places the format string into flash, which requires the second argument to be a string constant. I looked into putting together a second macro that allows the format string to be any expression (like here), but then noticed that there are only 3-4 places in the codebase where it's not a constant string, so I decided it was cleaner overall to just modify these places rather than add the confusion of a second macro. I tried to keep the non-esp8266 code as straight-forward as possible while adding strings-in-flash support.

src/jswrap_onewire.c
tve: Should: yes, have to: no.

targets/esp8266/jshardware.c
tve: The TX is not interrupt driven at all. It's possible to make it interrupt driven.

tve:
Thanks for merging. Looks like I need to run it through another series of tests.

src/jswrap_interactive.c
gfwilliams: That's great - thanks! I'll try and remember to do it this way in the future :)

PR #789, opened 2016-01-18
Efm32integration

LaplaceG:
The JS interpreter is now running on the EFM32 and working with the Espruino web IDE.
Use PD0/PD1 for TX/RX (Pin 4/6 on the expansion header on the EFM32GG starter kit).

No advanced hardware functions implemented yet.


gfwilliams:
This is great - thanks for making this such a clean pull request!

f4tt5ylp.wp4.txt
```diff
@@ -0,0 +1 @@
+Merge of espruino/master into efm32integration
```
tve: what's the purpose of this file?

gfwilliams: Good spot - looks like it's a temporary file created by Git (I think so you can edit the commit message). Not sure how it happened - maybe if you do a `git commit` in one window and edit the message, then do it again in a second window.

LaplaceG: Yeah, this is something git created. Should be removed.

gfwilliams: Just done.

PR #790, opened 2016-01-18
Update Makefile

MaBecker:
make "stat" work with Linux and Mac OS X


gfwilliams:
Thanks for doing the PR! I would have got around to the other bug eventually - I was just a bit too busy :)

PR #791, opened 2016-01-18
fix socket unsent data error; update esp8266 socket tests

tve:
The diff of the tests is not good 'cause of cr-lf vs lf line ending changes :-(.


gfwilliams:
Thanks!

PR #793, opened 2016-01-21
Efm32integration

LaplaceG:
Flash and time now working on EFM32

### DIFF
--- a/scripts/build_platform_config.py
+++ b/scripts/build_platform_config.py
@@ -93,8 +93,10 @@ if not LINUX:
     flash_page_size = 4*1024
   if board.chip["family"]=="EFM32GG":
     flash_page_size = 4*1024
+    flash_saved_code_pages = round((flash_needed+flash_page_size-1)/flash_page_size + 0.5) #Needs to be word-aligned and we have plenty of space, so we're rounding up
+  else:
+    flash_saved_code_pages = (flash_needed+flash_page_size-1)/flash_page_size
   # F4 has different page sizes in different places
-  flash_saved_code_pages = (flash_needed+flash_page_size-1)/flash_page_size
   total_flash = board.chip["flash"]*1024
 
   if "saved_code" in board.chip:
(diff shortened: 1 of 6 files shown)


scripts/build_platform_config.py
```diff
@@ -266,7 +268,7 @@ def codeOutDevicePin(device, pin, definition_name):
   #codeOut("#define JSVAR_CACHE_SIZE                "+str(200)+" // Number of JavaScript variables in RAM")
 else:
   codeOut("#define JSVAR_CACHE_SIZE                "+str(variables)+" // Number of JavaScript variables in RAM")
-  codeOut("#define FLASH_AVAILABLE_FOR_CODE        "+str(flash_available_for_code))
+  codeOut("#define FLASH_AVAILABLE_FOR_CODE        "+str(int(flash_available_for_code)))
```
gfwilliams: Is there really a need for this? Surely these values are integers anyway?

LaplaceG: My python produced outputs like 1024.0 for these defines without explicitly casting them to ints  so I guess there's a different default for this.
On a side note, is flash_saved_code_pages in the same file supposed to be an int or a float? Leaving it as a float has the side effect of the MAGIC words not being word-aligned..

gfwilliams: Oh great. God knows why everyone likes Python so much :(

I just tested and on Python 2, `2/1 == 1`, but on Python 2 it's `1.0`.  All those things should be integers - so I guess they should all be cast to ints.

What system are you running the build on such that the default is Python 3? I'm using Ubuntu here and while I have 3 installed, the default is still 2.

LaplaceG: Yes, Python is nice because you don't have to declare variable types. But then again Python is hit because you don't have to declare variable types. Basically, I've installed everything manually, so that's why I have Python 3.

I have a patch for the other variables as well for EFM32, but now I can extend it to the other board as well. 

Should I start pushing to master, or continue to push to efm32integration and issue pull requests?

gfwilliams: That'd be great - doing what you're doing works well really.

PR #798, opened 2016-01-26
Reduce Advertising Interval to 750ms

beaufortfrancois:
As seen in https://code.google.com/p/chromium/issues/detail?id=580548#c4, reducing the Advertising Interval makes it easier to connect to Chrome OS devices.

When Chrome OS 50 reaches stable channel, it might be good to revisit this decision and see if we can reset the Advertising Interval to 1s in order to keep battery's life longer.

R=@gfwilliams


gfwilliams:
Awesome - thanks! The difference in battery between 750ms and 1000ms won't be that bad at all (not like 250ms!)

PR #799, opened 2016-01-26
Efm32integration

LaplaceG:


gfwilliams:
Great, thanks! Are the project files really needed? At least from my experience they tended not to be very portable between different computers/OSes/Eclipse versions.

LaplaceG:
The project files are not needed. They are just there to make it easier for anyone that wants to start developing for the efm32.
The project files are made for Simplicity Studio, which is a Silicon Labs version of eclipse.
I doubt that they will be compatible with standard eclipse, but at least it's portable between all versions of Simplicity Studio.
In the end it's your choice whether you think it's a good idea to have them there.

gfwilliams:
Thanks - I guess it's worth leaving them in then... If they keep changing, please can you try not to keep committing the changes though, unless it's absolutely required? 

If you have building instructions it might be an idea to put them in https://github.com/espruino/Espruino/blob/master/README_Building.md ?

PR #802, opened 2016-01-29
Switched to console over USB (debug) and started implementing setWatch

LaplaceG:


gfwilliams:
Wow, you're in a roll this week :)

PR #803, opened 2016-02-01
Efm32integration - Now with pin wakeup

LaplaceG:


gfwilliams:
Looks great - thanks!

PR #804, opened 2016-02-03
Update bitmap_font_3x5.c

MaBecker:
fixed some syntax errors in definition of const unsigned short LCD_FONT_3X5


gfwilliams:
Thanks! That's strange... I'm sure it had worked at some point, but I can't see how!

PR #806, opened 2016-02-11
Dev dfu ota

mjdietzx:
DFU OTA now supported on both nRF51 & nRF52 devices.
Bonding in BLE connections now supported.

Either do a button less DFU  by writing to the DFU service characteristic in Espruino's BLE profile or hold down button 4 when starting up. Boot loader is added to project as a pre compiled hex file (unmodified from the Nordic SDK except changed to be single bank). Also a bootloader_settings.hex file is provided in the project to write a flag so we can flash the softdevice+bootloader+application and the application will properly run initially (instead of having to program the application by DFU). All these files are merged together for espruino...hex (optionally can build without merged softdevice or bootloader by commenting out line in makefile).

Saving to flash functionality removed for now. Also DUSE_DEBUGGER, DUSETAB_COMPLETE, USE_GRAPHICS commented out for now because flash is very tight. We need to use less flash on nRF51 (currently Espruino is about 1kb too big to be updated over the air - 0x20200 image size limit on nRF51 for single bank).

Also linker files were fixed for the new SDK, softdevices (they were slightly wrong). And ADC stuff will not be compiled for nRF52 since it isn't compiling (#ifndef NRF52).


mjdietzx:
best way for DFU OTA now is either Nordic Master Control Panel app on pc or android. Working on making this available in web bluetooth API and may be able to implement in Espruino online IDE.

gfwilliams:
Thanks! Did the ADC not work for nRF52 then? I thought we'd sorted that.

Looking at this, it's going to break anything using nRF51 for the moment, so I think I'll have to stick it on a branch for now (as I'm hoping people are starting to use the micro:bit port now).

When you say the linker was wrong... was it? With central link=0 and periph=1, the definitions referenced here show what the correct start address should be: https://github.com/espruino/Espruino/pull/806/files#diff-5add3d291e0d616dac204cab0b3a10f6L51

In fact on initialisation it may now warn you that the addresses are wrong?

Also, is using the external low speed oscillator required? Using it will stop it working on pretty much anything that's not the devkit, as it seems most people leave the LSE off.

One other thing - what's with the reduced advertising interval? Surely advertising 20x more often would kill anything battery powered?

mjdietzx:
Yeah nRF52 hasn't been compiling with ADC stuff - I think its ADC peripheral is slightly different than nRF52 but i haven't really looked at it.

Yeah I think it is a good idea to keep this as a branch for now.

Sorry I think you're right about the linker. Ill revert that.

I just set the external clock and advertising interval for when I was testing. Your right about that so I'll revert those as well. I didn't realize people were developing with micro:bit already? is there an official release for any nordic devices? ill have to be more careful when working on this now

gfwilliams:
Well, there's http://www.espruino.com/MicroBit and there are [travis builds](http://www.espruino.com/binaries/travis/e42ca309cdd9e55cea592bf28ff08cfbfa5c95fb/) available. There are definitely 2 people I know of using it, probably a few more now as I've pointed a few people with micro:bits at it.

I've also been toying with using it on a cheap fitness watch - although for that, DFU is pretty much required :) http://forum.espruino.com/conversations/280747/

mjdietzx:
Hey, are you sure the ideal ram settings are correct? When I change the linkers and try to test Espruino doesn't run (no welcome message or BLE advertising). once i change it back to RAM start 0x20001F00 everything works fine.

mjdietzx:
btw, what do you think the easiest ways to save flash are? if we enable DEFINES+=-DUSE_DEBUGGER -DUSE_TAB_COMPLETE and USE_GRAPHICS in microbit we overflow flash by ~10kb. I think we need to shave off 12kb for DFU OTA to work properly. And if we want to save flash we need a bit more..

I know theres alot of stuff in Nordic SDK taking up flash like app_timer.

gfwilliams:
>  When I change the linkers and try to test Espruino doesn't run

Strange... Did anything else change? The existing build works fine at those addresses - maybe it's something to do with how the DFU stuff sets things up?

> what do you think the easiest ways to save flash are? 

Have you tried `find_big_rom.sh`? 10k is quite a lot to try and skim off.

Potentially we could fiddle with what's inlined and what isn't... @tve had some luck with ESP8266 by removing the mentions of `inline` in `jsutils.h`...

Also, at the moment Espruino uses its own maths libs, and I wonder whether we could move over to using GCC's? Originally I'd had problems with the GCC ones pulling in malloc and a bunch of other stuff, but I think I was just using the wrong libraries.

> app_timer

Is app_timer taking up flash, or just RAM? I guess potentially we could hack up some stubs that used Espruino's `setTimeout` code for app_timer?

mjdietzx:
If I remember correctly all that changed was:
service changed indications now present
added the dfu service
added bonding capabilites

so maybe this accounts for it? I have a bit of time to look into it today.

As for app_timer it would save us 1.8k of flash. so still a aways to go.. I think bonding takes up alot of flash.. I don't think bonding is actually a requirement for DFU but maybe I'm wrong. I can look into it but if we can take it out that will save atleast another 5k of flash and shouldn't be difficult at all.

I'm hoping just by eliminating Nordic specific stuff we can save around 10k but we'll see..

gfwilliams:
Hi - did you ever look at this a bit more?

I fiddled around with some stuff and I've managed to drop around 11kB off the micro:bit image. I've also got the ADC working (I think!) on nRF52.

I just tried to merge in your changes and it looks like we're now only 1.5k short when building with everything enabled, so even getting a relatively small amount off the Nordic libs would do it.

I added a branch for it with the merge done - it's here: https://github.com/espruino/Espruino/tree/nordic_dfu_ota

Problem is, I can't seem to get it working on the micro:bit at all - with or without DFU :(

mjdietzx:
Hey,

Nice! We can definitely trim off 1.5kb. We still have the problem with (TIMER1?) stopping after a while? I can't remember but that was because app_timer in the ble right? I think removing app_timer.c will save a bit over 2.1kb of flash and .2kb of RAM and we need to do this anyways.

Also bonding probably costs us around 7kb of flash and around .6kb of RAM. It says it is required in Nordic's documentation for DFU but I don't think it is. I was thinking about looking into removing bonding capabilites unless you think this is a feature we should keep in Espruino.

Anyways sorry I've left this branch stale for a while. As for mbed not working thats strange. I'm going to try to take a look today but not sure if I'll get to it. And will have some time early next week to get bck to Espruino.

mjdietzx:
starting to look at the branch now. I'm wondering if mbed is starting in bootloader mode. I provided a precompiled hex file of the bootloader straight from the SDK. 

there are two ways of getting into bootloader mode: buttonless DFU, holding down a button while booting.

So for nRF51 when gpio 16 is low during startup we will startup in bootloader mode instead of application mode (buttons on our DK are active low). So is it possible mbed is in bootloader mode and thats why it looks like its not working at all? Maybe when you run espruino you can detect it advertising as DFU_Targ? Otherwise I need to continue looking but just off the top of myhead - not sure what mbed pcb does

gfwilliams:
Thanks! With bonding, I guess it'd be useful in general - to be able to limit the people that can connect and reprogram the device? 7k seems pretty hefty though. If that were something that could be left out on nRF51, I can't imagine it'd be missed much.

Yes, Timer1 stops after a while (easiest way to check is to `show(0xFFFFFFF)` on the micro:bit) - I think app_timer actually uses RTC1? If it uses Timer1 as well then that could be the issue.

But app_timer is a nightmare - APP_TIMER_BUF is actually 450 bytes of RAM! If that could be improved/replaced it'd be way better. I think it is used by the BLE advertising stuff though? Not sure exactly.

I'll have a look and see if it has entered DFU. That certainly sounds like it could be the issue.

gfwilliams:
Just tried - I don't seem to be able to see it I'm afraid.... It should show as a device in the master control panel?

The nRF51 devkit build does work though. Am I right in thinking that `DFU_UPDATE_BUILD=1` actually just builds a hex file for the updater app, so it's not something you can upload using the mbed bootloader?

mjdietzx:
Yeah I didn't put too much thought into DFU_UPDATE_BUILD. But the way it is now:

If DFU_UPDATE_BUILD is defined we want to build Espruino to be updated via DFU. This means we know that there is already the SoftDevice and bootloader present on the board. When we do the DFU OTA we only want to send the image of Espruino (not merged with the SoftDevice or bootloader). So we do not merge and send only the espruino firmware build for the device.

mjdietzx:
So I don't think this can be uploaded via the mbed bootloader because you would need espruino merged with the softdevice for this.

mjdietzx:
But the bootloader does not need to be present for Espruino and the softdevice to run properly. Everything would work the same but if somebody tried to put the device in bootloader mode it would reset, see no bootloader is present and start back up in application mode.

mjdietzx:
Yeah app_timer only uses RTC1. Also it is using so much ram because of line 73 in jswrap_bluetooth.c. I think APP_TIMER_OP_QUEUE_SIZE should be set to 1 (maybe even 0?) https://devzone.nordicsemi.com/question/723/how-big-should-app_timer_op_queue_size-be/

It is only used in ble_conn_params.c. I should probably look into it a bit more and find out why it is so useful before messing with it..

anyways won't get anything useful done today. gonna be a battle with flash - I'm thinking removing bonding capabilites for nRF51 might give us the biggest bang for our buck for now.

not sure if you want to keep this PR open, but I'll work out of nordic_dev_ota branch you merged this into.

gfwilliams:
Hmm, yeah - I was unsure about app_timer - I'd copied that define from the original BLE examples. I just tested and with 0 it works(ish) but resets when you disconnect from BLE. With 1 it's stable so I'll use that :)

Yeah, I'll close the PR - feel free to do open a new one later on though

PR #810, opened 2016-02-23
Make crypto compile for ESP8266

urish:
Before this fix, compiler complained about `invalid initializer`.


gfwilliams:
Great - thanks!

PR #811, opened 2016-02-26
fix #809 esp8266 socket close

### LINKED ISSUE #809: http server runs out of sockets
Reported by Wilberforce:

When testing and debugging, after a while to http server code fails to start with an out of sockets error.
This happens after a number of uploads from the web ide.
I believe it is due to the old http server sessions not shutting down properly.
I have my board set up so it auto joins my local wifi.
At the moment I just hit the reset button on the board, and re upload, however it would be handy to shutdown the connections cleanly. Can this be down with an E.event on after a .reset?
Wifi.disconnect and wifi.connect() does do the job, however it would be great if an event could trigger.

tve:
I have not done extensive testing, but it does fix the issue as far as I can tell.

### DIFF
--- a/libs/network/esp8266/network_esp8266.c
+++ b/libs/network/esp8266/network_esp8266.c
@@ -475,9 +475,9 @@ static void doClose(
     if (rc != 0) {
       setSocketInError(pSocketData, rc);
     }
-    // FIXME: do we get a disconnected callback or is this it? If we don't get a callback we can
-    // go straight to SOCKET_STATE_CLOSED
-    pSocketData->state = SOCKET_STATE_DISCONNECTING;
+    // we do not get a disconnected callback so we go straight to SOCKET_STATE_UNUSED
+    pSocketData->state = SOCKET_STATE_UNUSED;
+    pSocketData->creationType = SOCKET_CREATED_NONE;
 
   } else {
     int rc = espconn_disconnect(pSocketData->pEspconn);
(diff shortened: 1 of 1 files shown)


gfwilliams:
Cool - thanks!

wilberforce:
I'm not sure of this is related, I'm getting this:
```
 1v85 Copyright 2016 G.Williams
Espruino is Open Source. Our work is supported
only by sales of official boards and donations:
http://espruino.com/Donate
WARNING: the esp8266 port is in beta!
Flash map 4MB:512/512, manuf 0xe0 chip 0x4016
>echo(0);
=undefined
> server.listen(80);
ERROR: Unable to create socket
=undefined
>require('ESP8266').printLog()
ted socket 10 inbound to port 80 from 192.168.15.41:55217
202082> net_esp8266: accepted socket 11 inbound to port 80 from 192.168.15.41:55220
223065> net_esp8266: error -15 creating listening socket 12: ESPCONN_ISCONN
240000> Wed Mar 09 00:52:37 2016, heap: 1872
=undefined
```

I think I have the patch above.:
```
>process.env
={
  "VERSION": "1v85",
  "BUILD_DATE": "Feb 26 2016",
  "BUILD_TIME": "08:21:56",
  "GIT_COMMIT": "a9f4524670426b5856a6aaf14426fec2d325a486",
  "BOARD": "ESP8266_BOARD",
  "CHIP": "ESP8266",
```

PR #818, opened 2016-03-13
Create a Dockerfile for the ESP8266 build

urish:
This docker file contains both the Espressif SDK and the xtensa tools, and makes it super-easy to create builds for ESP8266

### DIFF
--- a/Dockerfile-esp8266
+++ b/Dockerfile-esp8266
@@ -1,0 +1,23 @@
+FROM ubuntu:14.04
+
+RUN apt-get update && apt-get install -y software-properties-common && apt-get update \
+ && add-apt-repository ppa:terry.guo/gcc-arm-embedded \
+ && apt-get update \
+ && apt-get install -y git gcc-arm-none-eabi build-essential python curl
+
+RUN git clone https://github.com/espruino/Espruino espruino
+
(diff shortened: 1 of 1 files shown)


gfwilliams:
Cool - thanks! Just wondering, do you think this is worth merging into the main Docker file? It wouldn't take that much longer to complete, and then there would be one system that would be capable of compiling every Espruino firmware?

urish:
That makes sense, though, it might break it some point if the s3.voneicken.com links become invalid.

Your call :)

gfwilliams:
Well, I'll just pull this in for now. Maybe someone will care about it in the future. I should probably just make a GitHub repo with those files in...

MaBecker:
# esp_iot_sdk_v1.5.0
curl -Ls http://bbs.espressif.com/download/file.php?id=989 | tar -xvf -

## liblwip_536_for_ESP8266_NONOS_SDK_V1.5.0.zip
wget --content-disposition "http://bbs.espressif.com/download/file.php?id=1021"

gfwilliams:
What's that for? I'm more confident in @tve's server than I am in bbs.espressif.com :)

MaBecker:
That is the source where tve has pulled his one from

tve:
s3.voneicken.com is a CNAME for a bucket in AWS S3, so you will have better luck than downloading from china...

PR #820, opened 2016-03-20
add IN_FLASH_MEMORY  for esp

MaBecker:


src/jsutils.h
```diff
@@ -77,6 +77,14 @@ char *flash_strncpy(char *dest, const char *source, size_t cap);
 #define CALLED_FROM_INTERRUPT
 #endif
 
+#if defined(ESP8266)
+// For the esp8266 we need the posibility to store arrays in flash, because mem is so small
+#define IN_FLASH_MEMORY   __attribute__ ((section(".irom.literal"))) //__attribute__((aligned(4)))
```
tve: why is the aligned(4) commented out?

MaBecker:
cant remember, is it needed ?

wilberforce:
@MaBecker  - yes the `attribute((aligned(4)))` is required.

It makes sure that the strings start on a 4 byte boundary

wilberforce:
This is a little cleaner:

Line 49:
````#define FLASH_STR(name, x) static char name[] __attribute__((section(".irom.literal"))) 
__attribute__((aligned(4))) = x
````

````
#define IN_FLASH_MEMORY   __attribute__ ((section(".irom.literal"))) __attribute__((aligned(4)))
#define FLASH_STR(name, x) static char name[] IN_FLASH_MEMORY = x
````

wilberforce:
I've noticed the travis builds are failing on the other commits. The non-ESP8266 branch of the code needs to have the empty #define made:

````
#if defined(ESP8266)

/** Place constant strings into flash when we can in order to save RAM space. Strings in flash
    must be accessed with word reads on aligned boundaries, so we'll have to copy them before
    regular use. */
#define IN_FLASH_MEMORY   __attribute__ ((section(".irom.literal"))) __attribute__((aligned(4)))
//#define FLASH_STR(name, x) static char name[] __attribute__((section(".irom.literal"))) __attribute__((aligned(4))) = x
#define FLASH_STR(name, x) static char name[] IN_FLASH_MEMORY = x

/// Get the length of a string in flash
size_t flash_strlen(const char *str);

/// Copy a string from flash to RAM
char *flash_strncpy(char *dest, const char *source, size_t cap);

/** Read a uint8_t from this pointer, which could be in RAM or Flash.
    On ESP8266 you have to read flash in 32 bit chunks, so force a 32 bit read
    and extract just the 8 bits we want */
#define READ_FLASH_UINT8(ptr) ({ uint32_t __p = (uint32_t)(char*)(ptr); volatile uint32_t __d = *(uint32_t*)(__p & (uint32_t)~3); $

#else

#define IN_FLASH_MEMORY
/** Read a uint8_t from this pointer, which could be in RAM or Flash.
    On ARM this is just a standard read, it's different on ESP8266 */
#define READ_FLASH_UINT8(ptr) (*(uint8_t*)(ptr))
````

An the last macro defined for the non-esp8266 case, means that the code can be written once, with the block:


So instead of:
````
+#if defined(ESP8266)
 +    verts[idx+0] = (short)(x1+(((READ_FLASH_UINT8(&vectorFontPolys[vertOffset+i+0])&0x7F)*size+(VECTOR_FONT_POLY_SIZE/2))/VECTOR_FONT_POLY_SIZE));
 +    verts[idx+1] = (short)(y1+(((READ_FLASH_UINT8(&vectorFontPolys[vertOffset+i+1])&0x7F)*size+(VECTOR_FONT_POLY_SIZE/2))/VECTOR_FONT_POLY_SIZE));
 +#else
      verts[idx+0] = (short)(x1+(((vectorFontPolys[vertOffset+i+0]&0x7F)*size+(VECTOR_FONT_POLY_SIZE/2))/VECTOR_FONT_POLY_SIZE));
      verts[idx+1] = (short)(y1+(((vectorFontPolys[vertOffset+i+1]&0x7F)*size+(VECTOR_FONT_POLY_SIZE/2))/VECTOR_FONT_POLY_SIZE));
 +#endif    
````

This can be written:

````
 verts[idx+0] = (short)(x1+(((READ_FLASH_UINT8(&vectorFontPolys[vertOffset+i+0])&0x7F)*size+(VECTOR_FONT_POLY_SIZE/2))/VECTOR_FONT_POLY_SIZE));
 +    verts[idx+1] = (short)(y1+(((READ_FLASH_UINT8(&vectorFontPolys[vertOffset+i+1])&0x7F)*size+(VECTOR_FONT_POLY_SIZE/2))/VECTOR_FONT_POLY_SIZE));
````

MaBecker:
__attribute__((aligned(4))) is not needed because this has to be 2 byte, check details in issue 770

all my tests were based on 
````
#if defined(ESP8266)
#define IN_FLASH_MEMORY __attribute__ ((section(".irom.literal"))) //__attribute__((aligned(4))) 
#else
#define IN_FLASH_MEMORY 
#endif
#define READ_FLASH_UINT16(ptr) (READ_FLASH_UINT8(ptr) | (READ_FLASH_UINT8(((char*)ptr)+1)<<8) )  
````


sorry for confusion

thanks for the code improvement for graphics.c

@gfwilliams please decide how to implement

wilberforce:
Ah ok.

Then perhaps we need to have macros for flash, and another for alignment?

How about (To keep short):
#define IROM   .....

#define ALIGN4

And then these could get stacked?

However I'm not sure that you can drop the alignment of 4, as I think it needs to start on a 4 byte boundary..

gfwilliams:
Sorry for the delay responding to this - I'm actually on holiday this week, but I'll merge all this in when I get back on Tuesday next week.

gfwilliams:
Merged in... I kept the align (while not needed here, it would be needed if you were to use `IN_FLASH_MEMORY` for something that was 32 bits)

PR #824, opened 2016-03-29
#define USE_AES to make SHAx functions available on ESP8266

wilberforce:
I have tested with the Linux build and the Esp8266 build.

The linux builds as before.

For the esp   ./build-crypto.sh:

````
#! /bin/bash
export ESP8266_BOARD=1
export FLASH_4MB=1
export USE_CRYPTO=1
export ESP8266_SDK_ROOT=/home/esp8266/Espruino/esp_iot_sdk_v1.5.0
export PATH=~/esp-open-sdk/xtensa-lx106-elf/bin:$PATH
export COMPORT=/dev/ttyUSB0
export release=1
make $*
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1001,37 +1001,32 @@ ifeq ($(BOARD),MICROBIT)
 endif
 
 ifdef USE_CRYPTO
+  DEFINES += -DUSE_CRYPTO
   INCLUDE += -I$(ROOT)/libs/crypto
   INCLUDE += -I$(ROOT)/libs/crypto/mbedtls
   INCLUDE += -I$(ROOT)/libs/crypto/mbedtls/include
   WRAPPERSOURCES += libs/crypto/jswrap_crypto.c
+  SOURCES += \
+libs/crypto/mbedtls/library/sha1.c \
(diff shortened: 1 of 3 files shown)


MaBecker:
@wilberforce any changes to ESP8266_BOARD.py ?

gfwilliams:
Thanks! Just merged. I took out the `ifndef ESP8266` bit as I don't think it's needed (nothing seemed to use Crypto but not TLS).

wilberforce:
@gfwilliams thanks!
@MaBecker - I guess the ESP8266_BOARD.py should have an update so the web ide knows that the Sha crypto functions are available. The docs should be updated too... Perhaps something mentioning that esp8266 can now do web sockets ;-)

wilberforce:
@MaBecker I tested a build with crypto and graphics - it still fits, although I did not try on a board.

I pushed the boat out and tried crypto, graphics and the file system lib, however this didn't fit ( pretty close mind)

wilberforce:
Board.py already updated!

https://github.com/espruino/Espruino/blob/master/boards/ESP8266_BOARD.py

PR #825, opened 2016-03-29
Added deepSleep and modemSleep

urbiworx:
As discussed here: http://forum.espruino.com/conversations/281163/

### DIFF
--- a/targets/esp8266/jswrap_esp8266.c
+++ b/targets/esp8266/jswrap_esp8266.c
@@ -428,5 +428,35 @@ void jswrap_ESP8266_neopixelWrite(Pin pin, JsVar *jsArrayOfData) {
 
 #endif
 }
+//===== ESP8266.deepSleep
+/*JSON{
+  "type"     : "staticmethod",
+  "class"    : "ESP8266",
+  "name"     : "deepSleep",
+  "generate" : "jswrap_ESP8266_deepSleep",
+  "params"   : [
+    ["micros", "JsVar", "Number of microseconds to sleep."]
+  ]
 
+}*/
+void   jswrap_ESP8266_deepSleep(JsVar *jsMicros) {
+    if (!jsvIsInt(jsMicros)) {
+    jsExceptionHere(JSET_ERROR, "Invalid microseconds.");
+         return;
+    }
+    int sleepTime = jsvGetInteger(jsMicros);
+    system_deep_sleep(sleepTime);
 
+}
+
+//===== ESP8266.modemSleep
+/*JSON{
+  "type"     : "staticmethod",
+  "class"    : "ESP8266",
+  "name"     : "modemSleep",
+  "generate" : "jswrap_ESP8266_modemSleep"
+}*/
+
+void   jswrap_ESP8266_modemSleep() {
+    wifi_set_sleep_type(MODEM_SLEEP_T);
+}
--- a/targets/esp8266/jswrap_esp8266.h
+++ b/targets/esp8266/jswrap_esp8266.h
@@ -41,4 +41,7 @@ void   jswrap_ESP8266_neopixelWrite(Pin pin, JsVar *jsArrayOfData);
 
 uint32_t crc32(uint8_t *buf, uint32_t len);
 
+void   jswrap_ESP8266_deepSleep(JsVar *jsMicros);
+void   jswrap_ESP8266_modemSleep();
+
 #endif /* TARGETS_ESP8266_JSWRAP_ESP8266_H_ */


gfwilliams:
Thanks... In future though, could you add some documentation after the JSON to explain what the functions do - I'll add it now.

targets/esp8266/jswrap_esp8266.c
```diff
@@ -428,5 +428,35 @@ void jswrap_ESP8266_neopixelWrite(Pin pin, JsVar *jsArrayOfData) {
...
+}*/
+void   jswrap_ESP8266_deepSleep(JsVar *jsMicros) {
+    if (!jsvIsInt(jsMicros)) {
+    jsExceptionHere(JSET_ERROR, "Invalid microseconds.");
+         return;
+    }
+    int sleepTime = jsvGetInteger(jsMicros);
+    system_deep_sleep(sleepTime);
+
+}
```
tve: Mhh, this looks identical to https://github.com/tve/Espruino/pull/3/files, would have been nice to mention that it's a copy&paste of someone else's work?

targets/esp8266/jswrap_esp8266.c
```diff
@@ -428,5 +428,35 @@ void jswrap_ESP8266_neopixelWrite(Pin pin, JsVar *jsArrayOfData) {
...
+/*JSON{
+  "type"     : "staticmethod",
+  "class"    : "ESP8266",
+  "name"     : "modemSleep",
+  "generate" : "jswrap_ESP8266_modemSleep"
+}*/
 
+void   jswrap_ESP8266_modemSleep() {
+    wifi_set_sleep_type(MODEM_SLEEP_T);
+}
```
tve: What's the point of this function? The esp8266 runs in modem sleep mode by default, so this has zero effect...

urbiworx:
Actually i did not want to create the impression that this is my work and I did state this in http://forum.espruino.com/conversations/281163/ , however you might be right I could have added an additional comment to the source to make it clear, sorry for that. 
I did some further reading in the meantime and my understanding is now that I would have to trigger the force sleep method to add some use to the modem sleep method. I will try this out as soon as I have my usb meter or do you have a differing opinion on that?

tve:
modem_sleep is the default mode at boot time and it's not changed anywhere, so there's no point in switching to modem_sleep explicitly.

urbiworx:
My idea would be to use wifi_fpm_open(); to force immediate modem sleep (after set modem sleep). So calling ESP8266.modemSleep() would then immediately start the modem sleep instead of occasionally starting it.

tve:
I'm not sure you understand modem sleep. In modem sleep mode, the esp8266 automatically turns off the receiver in STA mode between AP beacons according to the AP's DTIM setting. So with DTIM=3 you get 300ms periods of low power (~16mA) followed by short bursts of high power (70-200mA) as the esp8266 receives the beacon and checks whether it has anything to be received. All this happens automatically while in modem sleep mode.

urbiworx:
Ok, you are right, I did some further reading and agree that I had some misconceptions regarding the modem sleep. Thanks for taking the time and explaining it. 

So the only further thing that could be implemented in the ESP8266 class would be lightSleep with external wake-up (as using a configurable wake up GPIO differs from the jshSleep concept)? 
The jshSleep in jshardware.c itself (without external wake-up) could only be implemented when Espruino has an updated SDK containing the wifi_fpm_set_wakeup_cb method as far as I know.

Is my understanding correct so far?

gfwilliams:
So what's the upshot of this? Pull out the 'light sleep' code, but leave deep sleep in there?

tve:
Yup, pull out the modem sleep code...

@urbiworx the current SDK version used by espruino does support forced light sleep.

urbiworx:
I have to agree with tve about the modemSleep - the only way that this would make sense would be to implement also the method noSleep (or comparable) which would turn of sleep completely at the ESP so you could turn it on again with modemSleep. However I could not find a usecase for turning off the sleep functionality, so that would also make no sense. Sorry for the confusion

@tve: Would an external lightSleep method like this following make any sense?
`/*JSON{
 "type"     : "staticmethod",
 "class"    : "ESP8266",
 "name"     : "lightSleep",
 "generate" : "jswrap_ESP8266_lightSleep",
 "params"   : [
   ["pin", "pin", "Pin for wake up"],
   ["jsMode", "JsVar", "wake up mode"]
 ]
}*/
void   jswrap_ESP8266_lightSleep(Pin pin, JsVar *jsMode) {
	if (!jshIsPinValid(pin)) {
		jsExceptionHere(JSET_ERROR, "Pin is not valid.");
		return;
	}
	if (!jsvIsInt(jsMode)) {
		jsExceptionHere(JSET_ERROR, "Invalid mode.");
         return;
    }
	int mode = jsvGetInteger(jsMode);

	wifi_station_disconnect();
    wifi_set_opmode(NULL_MODE);    
	gpio_pin_wakeup_enable(GPIO_ID_PIN(pin),(GPIO_INT_TYPE) mode);
	wifi_set_sleep_type(LIGHT_SLEEP_T);
	wifi_fpm_open();
}
`
Please note that I did not try it out yet. Just wanted to make sure something like that would be usable before I invest too much effort.

gfwilliams:
Ideally this should be part of `jshSleep` - the idea of Espruino is to be able to sleep automatically where possible.

If there's no network connection, no timer, and one watch then it could just do this?

Although I wonder whether it doesn't do something like this already? The RTOS should be aware of when it is busy and when it isn't...

At the end of the day, space in flash is getting tight, and *there's no point just adding functionality because you can*. It's got to actually be stuff that will get used.

tve:
LIght sleep indeed needs to be integrated with jshSleep and the idle loop 'cause it takes effect when the idle loop returns to the RTOS dispatch and not immediately when wifi_fpm_open is called. It's definitely useful but needs some serious work & testing, which is why I haven't implemented it...

PR #827, opened 2016-03-31
Allow esp8266 > 1mb flash access via flash module

wilberforce:


gfwilliams:
Thanks! Do you think it's worth removing `os_printf` from flash read? The flash write is expected to be slow, but having a slower read could be annoying.

wilberforce:
Will do - Not sure how that got through! I'll do a couple of other tidy ups too

PR #828, opened 2016-04-01
add `tone()` and `noTone()` for ESP8266

urish:
Hi, I have implemented `tone()` and `noTone()` for the ESP8266 port. 

I tested them on a WeMos D1 Mini device with a speaker connected, and they seem to produce the correct tones (I tested several frequency). The implementation uses the internal ESP8266 hardware timer (timer 1), and was inspired by [ESP8266/Arduino](http://github.com/ESP8266/Arduino).

I copied the Arduino code constants for setting up the timer interrupts as I haven't found any example of using the ESP8266 hardware timers inside Espruino, but I am pretty sure I missed something. Otherwise, I think that we can adopt [esp8266_peri.h](https://github.com/esp8266/Arduino/blob/master/cores/esp8266/esp8266_peri.h) instead of defining all the constants inside the C file.

I would really love to see this merged, as I want to use Espruino in a demo I am giving in a talk next month in London.

Thoughts?

### DIFF
--- a/targets/esp8266/jswrap_esp8266.c
+++ b/targets/esp8266/jswrap_esp8266.c
@@ -470,3 +470,99 @@ void   jswrap_ESP8266_modemSleep() {
   wifi_set_sleep_type(MODEM_SLEEP_T);
 }
 
+// Helper functions and definitions for for tone()
+LOCAL uint32_t tonePinBitmask = 0;
+LOCAL uint8_t toneState = 0;
+
+LOCAL void CALLED_FROM_INTERRUPT toneIsr() {
+  if (toneState) {
+    GPIO_REG_WRITE(GPIO_OUT_W1TS_ADDRESS, tonePinBitmask);
+  } else {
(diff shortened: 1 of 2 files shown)


gfwilliams:
Hi - I'm afraid I don't want to take this in as-is - every version of Espruino does PWM output (including ESP8266) using `analogWrite` - adding a new function called `tone` that does the same thing in a slightly different way would just be confusing.

Could you have a go at using your code to drive the Espruino utility timer? https://github.com/espruino/Espruino/blob/master/targets/esp8266/jshardware.c#L1092

So call `jstUtilTimerInterruptHandler` from your interrupt routine. If you do that, you'll get proper PWM on *any* pin (even at once), using `analogWrite`. It'll also do Waveform input and output - so would be a huge improvement all around (the existing util timer code isn't preemptive, so is useless for doing PWM at higher speeds)

urish:
Hi @gfwilliams, thanks for the feedback!

 I will have a go at using my code to drive the Espruino utility timer and send another PR.

gfwilliams:
Closing this as inactive now.

PR #835, opened 2016-04-05
Update jswrap_esp8266.c

MaBecker:
make sure neopixelWrite() is working even if pinMode(pin,"output") is missing in coding.

### DIFF
--- a/targets/esp8266/jswrap_esp8266.c
+++ b/targets/esp8266/jswrap_esp8266.c
@@ -346,7 +346,7 @@ void jswrap_ESP8266_neopixelWrite(Pin pin, JsVar *jsArrayOfData) {
     return;
   }
 
-  if (!jshGetPinStateIsManual(pin))
+  if (jshGetPinStateIsManual(pin) != JSHPINSTATE_GPIO_OUT)
     jshPinSetState(pin, JSHPINSTATE_GPIO_OUT);
 
   // values for 160Mhz clock


gfwilliams:
This looks very wrong to me... `jshGetPinStateIsManual` returns a boolean, so should not be compared with JSHPINSTATE_GPIO_OUT.

This should have produced a warning when compiling really.

MaBecker:
So I misunderstood this doc:

https://github.com/espruino/Espruino/wiki/Architecture-of-GPIO

What I want to do is to set the given pin to output if not set to output by code 

[make_output.txt](https://github.com/espruino/Espruino/files/204267/make_output.txt)

gfwilliams:
That is exactly what:

```
 if (!jshGetPinStateIsManual(pin))
      jshPinSetState(pin, JSHPINSTATE_GPIO_OUT);
```

(The existing code) does. Just like would happen for `digitalWrite`.

So this should work fine.

You don't want it to set the output mode to OUT always, because there are very good reasons for wanting to use OUT_OPENDRAIN sometimes (especially with WS2811);

MaBecker:
code was working fine with IDE, when using save() power off/on, the code was only working after adding e.g. pinMode(D4,"output"), this is the reason why I started changing this. 

I guess the difference is with the pinMode settings to "input" after power up

so it should be  

````
if ( jshPinGetState(pin) != JSHPINSTATE_GPIO_OUT)
     jshPinSetState(pin, JSHPINSTATE_GPIO_OUT);
`````

gfwilliams:
Do you have the same issue with digitalWrite? (you should do)

It could be a problem with `jshGetPinStateIsManual` not getting reset when the device restarts?

As I said above, setting to GPIO_OUT regardless is not a good solution. If you run WS2812 off of full 5v then you can't use 3.3v logic levels to drive them reliably, and have to use OUT_OPENDRAIN. Your changes would make it impossible to do that.

Issue #697, opened 2015-11-08
Constant strings (and other data) moved to flash memory


nkolban:
With the port of Espruino to the ESP8266, an interesting characteristic has been uncovered.   It appears that on some architectures, constant strings (which are strings that are defined as constants ... usually between double quotes) are placed in precious data RAM storage as opposed to a potential alternative which is that they be held in flash memory.   There have been a number of discussions in this topic area in general.  This issue is being raised to examine what opportunities might exist within the Espruino code base to take advantage of this.

Here are some links to potentially related materials:
- https://github.com/esp8266/Arduino/blob/master/doc/reference.md#progmem
- http://www.esp8266.com/viewtopic.php?f=9&t=6376&p=33326#p33041
- https://github.com/SuperHouse/esp-open-rtos/issues/11

From a notional perspective, it would be ideal if the Espruino code could be compiled or linked in some fashion that would "just" place the constant data in flash.  However, that may not be possible.  If it weren't possible, the next question would be "If a preparation could be performed on String constants in the code base, would this issue warrant the rework of the code base to achieve that?".

For example, if instead of coding:

```
if (strcmp("Hello", myData) == 0)
```

we might have to code:

```
if (strcmp(F("Hello"), myData) == 0)
```

would such an undertaking to work through as much of the source code as we could be worth it?  Obviously a default macro would result in zero change to the data ... so the cost would be in the dimensions of "time" and "benefit".   It is suspected that one will trade app execution speed for reduction in RAM usage.

gfwilliams:
There were some good ideas here: https://github.com/SuperHouse/esp-open-rtos/issues/11

I can across a semi-similar thing when looking at porting to Arduino, and in the end I gave up. IMO it's just too much effort to wrap every function call this way (not to mention it's a massive source of potential bugs). It's kind of frustrating the toolchain can't deal with it in a better way.

Since Espruino has its own implementation of the string functions (on properly embedded systems quite a few of the built-in functions pull in stuff that really bloats the binary) perhaps you could just reimplement those functions so that they do aligned 32 bit reads, then all strings could go into Flash and you'd be fine?

gfwilliams:
... just to add that as far as I can gather, since the ESP8266 isn't harvard architecture you don't have to worry about code meant for strings being used for something else because it will 'just work'. It actually makes things a lot easier.

As that post you found says, you *could* just make every read an aligned 32 bit read and everything would be fine, if slow and a bit bloated.

[...]

PR #841, opened 2016-04-13
jsvObjectSetChildAndUnLock string constants to Flash for ESP8266

wilberforce:
https://github.com/espruino/Espruino/issues/697#issuecomment-208298721

The  `do { ... } while (0)` trick, reduced the changes quite considerably.

This yields about 600 bytes or so of heap - which goes a long way to reducing memory issues when you try to use web sockets on this board.

### DIFF
--- a/src/jsutils.h
+++ b/src/jsutils.h
@@ -74,6 +74,14 @@ int flash_strcmp(const char *mem, const char *flash);
 /** Read a uint16_t from this pointer, which could be in RAM or Flash. */
 #define READ_FLASH_UINT16(ptr) (READ_FLASH_UINT8(ptr) | (READ_FLASH_UINT8(((char*)ptr)+1)<<8) )
 
+#define __concat(o,l) __to_rom##l
+#define jsvObjectSetChildAndUnLock(obj,name,child) jsvObjectSetChildAndUnLock_FLASH_CTR(obj,__COUNTER__,name,child)
+#define jsvObjectSetChildAndUnLock_FLASH_CTR(obj,lbl,name,child) do { \
+    FLASH_STR(__concat(obj,lbl),name);\
+    jsvObjectSetChildAndUnLock_flash(obj,__concat(obj,lbl), child ); \
+    } while(0)
+
+#define jsvObjectSetChildAndUnLockVar(obj,name,child) jsvObjectSetChildAndUnLock_flash(obj,name,child)
 #else
 
 #undef USE_FLASH_MEMORY
--- a/src/jsutils.h
+++ b/src/jsutils.h
@@ -68,11 +68,18 @@ char *flash_strncpy(char *dest, const char *source, size_t cap);
    
 #define jsvObjectSetChildAndUnLockVar(obj,name,child) jsvObjectSetChildAndUnLock_flash(obj,name,child)
 
+/** 
+    ESP8266 store constants in flash */
+#define IROM_CONST __attribute__((section(".irom.const"))) __attribute__((aligned(4)))
 #else
 
 /** Read a uint8_t from this pointer, which could be in RAM or Flash.
     On ARM this is just a standard read, it's different on ESP8266 */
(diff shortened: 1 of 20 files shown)


src/jsutils.h
```diff
@@ -59,12 +59,28 @@ char *flash_strncpy(char *dest, const char *source, size_t cap);
     and extract just the 8 bits we want */
 #define READ_FLASH_UINT8(ptr) ({ uint32_t __p = (uint32_t)(char*)(ptr); volatile uint32_t __d = *(uint32_t*)(__p & (uint32_t)~3); ((uint8_t*)&__d)[__p & 3]; })
 
+#define __concat(o,l) __to_rom##l
+#define jsvObjectSetChildAndUnLock(obj,name,child) jsvObjectSetChildAndUnLock_FLASH_CTR(obj,__COUNTER__,name,child)
```
tve: I don't understand why this is needed. What is the path from the strings in flash to here? Why are the changes to the JswSymPtrs not all contained in the generated jswrapper.c file?

wilberforce:
@tve 
> Is there a reason you created another section as opposed to reusing the same as the FLASH_STR macro?

I was trying to minimise the changes to the generator code and had a conflict when the in the linker when the const keyword was used.

Using this:
`#define IROM_CONST __attribute__((section(".irom.literal"))) __attribute__((aligned(4)))`

````
LD espruino_esp8266_partial.o
targets/esp8266/esp8266_board_utils.c:44:1: error: __ESPCONN_MEM causes a section type conflict with jswSymbols_process
 FLASH_STR(__ESPCONN_MEM, "ESPCONN_MEM");
 ^
gen/jswrapper.c:1087:24: note: 'jswSymbols_process' was declared here
 static const JswSymPtr jswSymbols_process[] IROM_CONST = {
                        ^
lto-wrapper: xtensa-lx106-elf-gcc returned 1 exit status
````

This can be avoided by dropping the const keyword here:

`static const JswSymPtr jswSymbols_E[] IROM_CONST = {`

`static JswSymPtr jswSymbols_E[] IROM_CONST = {`

However this would either need a macro in `scripts/build_jswrapper.py` and I could not work out how to make the generator conditional on the board, and I thought @gfwilliams would prefer not to have conditional code for one board.

Do you know another way of avoiding `rror: __ESPCONN_MEM causes a section type conflict with jswSymbols_process` without creating a new section?

src/jsutils.h
wilberforce: This is a different change - separate from JswSymPtrs stuff.

It moves all of the string literals used in `jsvObjectSetChildAndUnLock` to flash. It saves about 600 bytes of heap.

tve:
You do need the two section as I've discovered myself. Sorry for the question.

gfwilliams:
@tve @wilberforce  I had a build error on the ESP8266 branch with the `const` keyword last week I think - the issue wasn't that stuff had the `const` keyword, but that *something else was being put into read-only memory that **didn't** have the const keyword*.

If you find what variable it is and whack  `const` on it, the issue should go away...

wilberforce:
@gfwilliams Help! - I'm struggling here.

I want to revert the changes made at 95f5d97 so I can pull @tve changes made back in the master.

At this stage I'm ready to dump everything in this pull request and re-create it.

Is there a way to unpull and revert my fork before these changes?

gfwilliams:
Well, you can definitely do `git cherry-pick` on a new branch in order to pull in your changes that *weren't* the one you wanted - there'll be a bunch of info online.

But - how much does memory this really save now? Surely now @tve's changes are in, the need to grab a few extra bytes isn't as high - I'm not a huge fan of what you've done with `ota.c` - it just seems like it's really obfuscated what's there for the sake of 100 bytes or so.

Next PR will be to change GetChildAndUnLock, then jsvNewFromString, then strcpy, and slowly the whole of Espruino's source is getting these random `#define ESP8266` all over it.

PR #843, opened 2016-04-17
move jswSymbol tables to flash on esp8266; also create Telnet serial device

tve:
**This PR is not ready to merge**

I believe I need to clean-up the following:
- add changelog entries
- keep JswSymPtr and JswSymList packed on non-esp8266
- clean-up src/jswrap_object.c so non-esp8266 platforms don't do the extra string copy

Anything else that jumps out?


src/jswrapper.h
```diff
@@ -58,19 +58,27 @@ typedef enum {
...
+#define PACKED_JSW_SYM
+#endif
+
 /// Structure for each symbol in the list of built-in symbols
 typedef struct {
   unsigned short strOffset;
-  void (*functionPtr)(void);
   unsigned short functionSpec; // JsnArgumentType
-} PACKED_FLAGS JswSymPtr;
+  void (*functionPtr)(void);
```
tve: The reordering here is to pack data better without the PACKED_FLAGS and to allow normal word access to functionPtr

tve:
LGTM

scripts/build_jswrapper.py
```diff
@@ -254,14 +255,49 @@ def codeOutBuiltins(indent, builtin):
...
+    int idx = (searchMin+searchMax) >> 1;
+    const JswSymPtr *sym = &symbolsPtr->symbols[idx];
+    unsigned short strOffset = READ_FLASH_UINT16(&sym->strOffset);
+    //os_printf_plus("min=%d max=%d idx=%d sym=%p off=%d=%x\\n", searchMin, searchMax, idx, sym, strOffset, strOffset);
+    int cmp = flash_strcmp(name, &symbolsPtr->symbolChars[strOffset]);
+    if (cmp==0) {
+      unsigned short functionSpec = READ_FLASH_UINT16(&sym->functionSpec);
+      //os_printf_plus("%s found %p: %p %x\\n", name, sym, sym->functionPtr, functionSpec);
+      if ((functionSpec & JSWAT_EXECUTE_IMMEDIATELY_MASK) == JSWAT_EXECUTE_IMMEDIATELY)
+        return jsnCallFunction(sym->functionPtr, functionSpec, parent, 0, 0);
```
gfwilliams: Would it be possible to re-use some of the code (like CallFunction/etc)? I just see this being a bit of a nightmare to maintain otherwise.

src/jsutils.h
```diff
@@ -57,21 +60,30 @@ size_t flash_strlen(const char *str);
 /// Copy a string from flash to RAM
 char *flash_strncpy(char *dest, const char *source, size_t cap);
 
+/// Compare a string in memory with a string in flash
+int flash_strcmp(const char *mem, const char *flash);
```
gfwilliams: I guess we could `#define flash_strcmp strcmp` when not on ESP8266 and it might help us remove a bunch of ifdefs from the code itself?

src/jswrap_object.c
```diff
@@ -222,8 +222,20 @@ void jswrap_object_keys_or_property_names_cb(
 
     while (symbols) {
       unsigned int i;
-      for (i=0;i<symbols->symbolCount;i++) {
-        JsVar *name = jsvNewFromString(&symbols->symbolChars[symbols->symbols[i].strOffset]);
+      unsigned char symbolCount = READ_FLASH_UINT8(&symbols->symbolCount);
+      unsigned short strOffset = READ_FLASH_UINT16(&symbols->symbols[i].strOffset);
+      for (i=0;i<symbolCount;i++) {
+#ifndef USE_FLASH_MEMORY
+        JsVar *name = jsvNewFromString(&symbols->symbolChars[strOffset]);
```
gfwilliams: Maybe we should just make jsvNewFromString work directly from flash with READ_FLASH_UINT8? It's used in a few places with constant strings, so would let us move quite a lot of extra stuff into flash?

src/jswrapper.h
gfwilliams: Yes, that's a great idea. It was dumb of me not to do it in the first place :)

src/jsutils.h
wilberforce: agreed - however this gets confusing reading the code for a board which it does not apply - it *looks* like it's using `flash_strncmp` - but isn't!

would something like `safe_strcmp` be more neutral?

gfwilliams: Yes, that sounds like a great idea! potentially the READ_UINT8 stuff could be renamed too.

src/jswrap_object.c
wilberforce: Good idea. It would also handle the `jsvObjectSetChildAndUnLock`  case as this function calls `jsvNewFromString` (Searches first I think)

scripts/build_jswrapper.py
```diff
@@ -313,10 +353,16 @@ def codeOutBuiltins(indent, builtin):
 codeOut('');
 codeOut('');
 
-codeOut('const JswSymList jswSymbolTables[] = {');
+# output the strings, possibly with __attribute__ to put them into flash
```
wilberforce: You can make this conditional on the -B{board-type} command arg:

````
if "-BESP8266_BOARD" in sys.argv:
    codeOut("static const JswSymPtr jswSymbols_"+codeName+"[] FLASH_SECT= {\n  "+",\n  ".join(listSymbols)+"\n};");
  else:
    codeOut("static const JswSymPtr jswSymbols_"+codeName+"[] = {\n  "+",\n  ".join(listSymbols)+"\n};");
````

This way the generated code does not have to have as many @ifdef conditionals

File: .travis.yml [hunk 6]
```diff
@@ -16,7 +16,7 @@ before_install:
  - sudo add-apt-repository -y ppa:terry.guo/gcc-arm-embedded
  - sudo apt-get update
  - sudo apt-get install libsdl1.2-dev gcc-arm-none-eabi
- - curl -Ls http://s3.voneicken.com/xtensa-lx106-elf.tgx | tar Jxf -
+ - curl -Ls http://s3.voneicken.com/xtensa-lx106-elf-20160330.tgx | tar Jxf -
```
wilberforce: What is the elf object with the 20160330 date?

tve: it's the compiler from a newer version of esp-open-sdk, corresponding to the 2016-03-30 commit of open-sdk.

scripts/build_jswrapper.py
tve: Sure. I thought that having two versions makes it cleaner 'cause you can actually read them, but I can try to mix and see what happens...

scripts/build_jswrapper.py
tve: Isn't what I have cleaner? Fewer lines of code and no conditional in the part that humans get to read...

src/jsutils.h
tve: The only place flash_strcmp is called is in jswrapper.c. In the current way that's written there is no extra ifdef because of it. But if I reorg the #define above may be useful.

What do you think of when you read "safe_strcmp"? Safe from what? Interrupts? Buffer overruns?

src/jswrap_object.c
tve: Isn't jsvNewFromString used a lot, including in all network buffer code? What the performance impact of this for the purpose of eliminating one #ifdef?

scripts/build_jswrapper.py
gfwilliams: Yeah, I know... I can just see me (or someone else :) changing just one of these in the future and totally screwing everything up

src/jsutils.h
gfwilliams: Well, safe from having bad things happen :) I know what you mean.

Yeah, I guess I was thinking it'd simplify the binary search code (and also I bet there are a whole bunch of `strcmps` against constant strings that could easily be switched over for ESP8266).

src/jswrap_object.c
gfwilliams: Yeah, it is used quite a bit (although quite a few of those will be with constant strings), but honestly I think in the scheme of things it's not actually going to make any noticeable difference to performance (like <1%).

I guess the other option is to have a 'flash-only' version of `jsvNewFromString` - I can just see this pattern of copy string to memory getting copied all over code as people try and get rid of strings from RAM, and the 64 byte buffer is going to cause trouble sooner or later when it's just copy+pasted.

src/jsutils.h
wilberforce: It would be useful to have the flash_strcmp, flash_strcpy and flash_strlen functions available globally, so that other optimisations can be made.

Would using `#define strcmp(x,y)   flash_strcmp(x,y)` make the need for `#ifdef` conditionals reduce, so that all strcmp are using this code?

Then there would be no need for special versions of `jsvNewFromString` as they would just *work*

src/jswrap_object.c
```diff
@@ -222,8 +222,20 @@ void jswrap_object_keys_or_property_names_cb(
...
+#else
+        // On the esp8266 the string is in flash, so we have to copy it to RAM first
+        // We can't use flash_strncpy here because it assumes that strings start on a word
```
wilberforce: One option here could be to use the python wrapper to pad the strings so that they do start on a word boundary

tve:
I'll revisit the PR according to comments in the coming days, couldn't do it tonight...

src/jswrap_object.c
gfwilliams: Hmm - that's a good idea - it might make things a lot easier.

I always forget, but I wonder if the endianness is such that instead of strcmp, a simple uint32 compare could be used to compare 4 characters at a time - that'd work well with ESP8266 and would be super fast too.

tve: You need big-endian for that...

gfwilliams: Shame... just looked and there is a 'rev' instruction in thumb code for swapping endianness, so it would still be pretty quick to do. Not sure if it's worth it though :)

Makefile
```diff
@@ -1797,16 +1799,30 @@ $(ESP_COMBINED512): $(USER1_BIN) $(USER2_BIN)
...
 	  [ "$$str" ] && echo "$$str $$f"; \
 	done | \
 	sort -rn >topstrings
 	$(Q)echo "Top 20 from ./topstrings:"
 	$(Q)head -20 topstrings
-	$(Q)echo "To get details: $(OBJDUMP) -j .rodata.str1.4 -s src/FILENAME.o"
+	$(Q)echo "To get details: $(OBJDUMP) -j .rodata.str1.1 -j .rodata.str1.4 -s src/FILENAME.o"
+
+# Same as topstrings but consider all read-only data
+topreadonly: $(PARTIAL)
```
wilberforce: Thanks for the topreadonly script!

running topreadonly - crypto is using quite a bit of rodata!

A simple change to the static data liberates more heap!

libs\crypto\mbedtls\library\sha256.c:134
static const uint32_t K[] __IN_FLASH_MEMORY__ =

and
libs\crypto\mbedtls\library\sha512.c:95
static const uint64_t K[80] __IN_FLASH_MEMORY__ =

Before:
 "freeHeap": 9248

after:
"freeHeap": 11248

I can do a PR once these changes are applied if you like

gfwilliams: Might be an idea - it's a real shame though, because this is a third part library that so far I've managed to use as-is. If I need to update it at a later date I'll need to re-apply these hacks (and let's face it, these hacks are going to end up *everywhere* over time).

I'm not sure if it's possible, but ideally we'd just have a list of these symbols somewhere, and would automatically force them into flash with a linker script or some other pass.

wilberforce: Perhaps we could attempt and upstream patch?

There are references to inrom type #defines, but nothing I can see that places the lookup tables in rom

gfwilliams: > Perhaps we could attempt an upstream patch?

I'm pretty sure they'd say 'No.'. Especially as it's ARM's code and you're suggesting adding random hacks to it in order to support another manufacturer's processors :)

> Is the config.h "out of the box"

No, that's been modified to include only what's needed. Feel free to play around and see, but I doubt you're going to get around the fact that the compiler needs telling to 

> Can this be done on a single declaration 

It might be tricker if it's static, as it will effectively be unnamed then :( There should still be some way though.

I believe you can use objcopy or something to put *everything from an entire object file into flash* - I'm sure I saw some ESP8266 hacks that did that. If the only stuff in those files is read-only then that could be an option.

tve: I'm a bit lost, what is the goal here?

My goal was to get a bit beyond the 1400 jsvars and more heap so I could load programs that I currently can't run reliably and so that the thing wouldn't run out of heap as soon as an extra buffer is needed due to a network hiccup.With the jswsym change that has been fully accomplished.

gfwilliams: Yeah, sorry - this should probably be discussed on a separate issue.

scripts/build_platform_config.py
```diff
@@ -286,7 +286,7 @@ def codeOutDevicePin(device, pin, definition_name):
   codeOut("#define FLASH_SAVED_CODE_START            "+str(flash_saved_code_start))
   codeOut("#define FLASH_SAVED_CODE_LENGTH           "+str(int(flash_page_size*flash_saved_code_pages)))
   codeOut("#define FLASH_MAGIC_LOCATION              (FLASH_SAVED_CODE_START + FLASH_SAVED_CODE_LENGTH - 4)")
-  codeOut("#define FLASH_MAGIC 0xDEADBEEF")
+  codeOut("#define FLASH_MAGIC 0xBADBEEF8")
```
tve: I changed the flash magic 'cause I've seen problem when upgrading older boards, if this is a problem, I can revert it.

src/jsinteractive.c
```diff
@@ -185,7 +185,7 @@ NO_INLINE void jsiConsolePrintString(const char *str) {
   }
 }
 
-#ifdef FLASH_STR
+#ifdef USE_FLASH_MEMORY
```
tve: As described in jsutils.h I had to switch from using defined(FLASH_STR) as an implicit indication that we're doing flashy stuff with constant strings to using an explicit #define just for that.

tve:
It turned out that the ESP version of the binary search could be used straight for other platforms with just a tad extra ifdefing. So now there's only one version :-). I briefly tested the linux version but it would be good for someone to test the ARM version to make sure I didn't break anything obvious.

scripts/build_platform_config.py
gfwilliams: As mentioned in the forum, you'll end up with unreliable code when upgrading from *any build* and using previously saved code, as the code can contain pointers to functions in flash memory - so I'd rather keep this as it was before as it doesn't fix the problem at all.

For example `a=print;save()` then `a("Hello")` will most likely break *each time the firmware is updated*.

The best solution would be to change the instructions so that the saved code (or at least the magic number) was overwittten each time the firmware was updated

wilberforce:
The Travis checks have passed, is it just a case of uploading the Travis build to a real espurino board?

How about posting to a forum and asking someone to check?

tve:
LGTM

MaBecker:
@wilberforce @tve I can run tests, please send a link to download that build.

wilberforce:
@MaBecker 
lastest travis builds:
http://www.espruino.com/binaries/travis/master/

MaBecker:
@wilberforce  got it, start testing now

gfwilliams:
That's great - thanks!

It's probably fine to merge now?

IMO it'd still be nice to have a `jsvNewFromFlashString` function though - it's a bit horrible in jswrap_object, and is a pattern that I expect will get repeated in the future :)

MaBecker:
@gfwilliams, this build works without  issues for me 

tested:  wifi,  save code,E.on(), neopixel with graphics, serial  and tcp WEB IDE

@tve & @wilberforce  great work - thanks

tve:
@MaBecker which platform did you test on?

@gfwilliams I'm done with changes. Someone else will have to tackle jsvNewFromFlashString, unfortunately, I need to get my Espruino based irrigation controller to work asap...

MaBecker:
@tve  on ESP8266 12E

MaBecker:
extended testing for this build and figured out that my i2c oled app is not working correct  any more after power off/on. 

I guess this is a side effect of a different pull request as part in this travis build I am testing, that saves gpios states, check the console logs below.

This side effect is not part of espruino_1v85.tve_master_cee7141_esp8266.

Soure code  and some console output:

`````
/* esp8266 12E 4MB

travis build 

espruino_esp8266_user1.bin	23-Apr-2016 04:05	448K

_____                 _
|   __|___ ___ ___ _ _|_|___ ___
|   __|_ -| . |  _| | | |   | . |
|_____|___|  _|_| |___|_|_|_|___|
          |_| http://espruino.com
 1v85.97 Copyright 2016 G.Williams
Espruino is Open Source. Our work is supported
only by sales of official boards and donations:
http://espruino.com/Donate
Flash map 4MB:512/512, manuf 0xe0 chip 0x4016
>dump();
pinMode(D0, "input_pullup");
pinMode(D4, "input_pullup");
pinMode(D5, "input_pullup");
pinMode(D12, "input_pullup");
pinMode(D13, "input_pullup");
pinMode(D14, "input_pullup");
pinMode(D15, "input_pullup");
=undefined

*/

// 0.91 OLED from adafruit,5pin GND,VCC,RST,SCL,SDA 

I2C1.setup({sda:D4,scl:D5});

function start(){
 g.drawString("Hello OLED !-!",2,2);
 g.flip(); 
}

var g = require("SSD1306").connect(I2C1,start,{height:32,rst:D12});

save();

E.on("init",start);

/*
*** after power off and on ***
Erasing Flash.....
Writing........
Compressed 25600 bytes to 5155
Checking...
DERROR: No ACK
ERROR: No ACK
ERROR: No ACK
ERROR: No ACK
ERROR: No ACK
ERROR: No ACK
.....
*/

/*
>dump();
function start() {g.drawString('Hello OLED !-!',2,2),g.flip();}
var g = Object.create(g.__proto__.constructor["prototype"]);
g.buffer = new ArrayBuffer([.....]);
g.flip = function () {h.forEach(function(a){b.writeTo(d,[0,a])});var a=new Uint8Array(129);a[0]=64;for(var c=0;c<this.buffer.length;c+=128)a.set(new Uint8Array(this.buffer,c,128),1),b.writeTo(d,a)};
g.setContrast = function (a) {b.writeTo(d,0,129,a)};
E.on("init", start);
I2C1.setup({"sda":D4,"scl":D5});
pinMode(D0, "input_pullup");

---> digitalWrite(D12, 0);   <----  why is that and why 0 and not 1 ?

pinMode(D13, "input_pullup");
pinMode(D14, "input_pullup");
pinMode(D15, "input_pullup");
=undefined


`````

gfwilliams:
Your code should never have worked properly - you need to call `var g = require("SSD1306").connect...` in `onInit`, because it sends I2C commands to initialise the display.

D12 is probably set to 0 because the code was halfway through resetting the OLED when you called `save`

gfwilliams:
Merging this in now,,,

PR #850, opened 2016-04-29
move shaX.c constants to flash, 1072 bytes additional heap

wilberforce:


wilberforce:
The websocket errors were out of memory:

````
{ "setpoint": 24 }
ERROR: Out of Memory!
Execution Interrupted during event processing.
at line 1 col 18
````

Increasing vars to 1700 fixes this.

File: libs/crypto/mbedtls/config.h [hunk 1]
```diff
@@ -118,6 +118,11 @@
 #define MBEDTLS_PLATFORM_CALLOC_MACRO(X,Y) jsvMalloc((X)*(Y))
 #define MBEDTLS_PLATFORM_FREE_MACRO(X) jsvFree(X)
 
+/* use flash safe version of memcpy so that rodata can be moved to irom */
+#ifdef ESP8266
+#define memcpy(d,s,n) flash_memcpy(d,s,n)
```
tve: So you're imposing this penalty for the sake of moving ~1KB of vars to flash?

wilberforce: This is not memcpy globally, it only applies to the crypto shaX.c functions, so the impact is minimal.

tve: I see, I didn't catch that.

tve:
I don't really understand all this. The problem you were having is running out of jsvars. You could have increased jsvars to 1700 and even 1800 without moving 1KB of additional strings to flash. 100 jsvars takes 1600 bytes of space...

gfwilliams:
As @tve says, `ERROR: Out of Memory!` is from running out of variable storage - not heap.

You could probably get around it just by removing some comments from your code or minifying it.

Makefile
```diff
@@ -505,7 +505,7 @@ BOARD=ESP8266_BOARD
 # Enable link-time optimisations (inlining across files), use -Os 'cause else we end up with
 # too large a firmware (-Os is -O2 without optimizations that increase code size)
 ifndef DISABLE_LTO
-OPTIMIZEFLAGS+=-Os -std=gnu11 -fgnu89-inline -flto -fno-fat-lto-objects -Wl,--allow-multiple-definition
+OPTIMIZEFLAGS+=-Os -std=gnu11 -fgnu89-inline -fno-fat-lto-objects -Wl,--allow-multiple-definition
```
gfwilliams: Is it a good idea to remove lto? It has some huge benefits for code size and performance,

Makefile
```diff
@@ -1737,6 +1737,11 @@ combined: $(ESP_COMBINED512)
 # generate partially linked .o with all Esprunio source files linked
 $(PARTIAL): $(OBJS) $(LINKER_FILE)
 	@echo LD $@
+#ifdef USE_CRYPTO
+	$(Q)$(OBJCOPY) --rename-section .rodata=.irom0.text libs/crypto/mbedtls/library/sha1.o
+	$(Q)$(OBJCOPY) --rename-section .rodata=.irom0.text libs/crypto/mbedtls/library/sha256.o
+	$(Q)$(OBJCOPY) --rename-section .rodata=.irom0.text libs/crypto/mbedtls/library/sha512.o
+#endif
```
gfwilliams: # in a makefile is a comment, so the ifdef will have no effect - is that intentional?

libs/crypto/mbedtls/config.h
gfwilliams: Shame it's needed really, but it's good that you've managed to avoid modifying mbedTLS.

Makefile
wilberforce: Thanks - updated

Makefile
wilberforce: It didn't have much impact on size, and when it's in place, you can't target the `rodata` section. It only add 100 bytes or so to each section

gfwilliams:
Thanks - and sorry, I just saw the other post about lto :)

gfwilliams:
Just merging - but I think there might be some confusion about what the `Out of Memory` error means... There are probably easier ways around it :)

wilberforce:
> You could have increased jsvars to 1700 and even 1800 without moving 1KB of additional strings to flash. 100 jsvars takes 1600 bytes of space...

If you increase js vars, doesn't this steal from RAM? So by adding strings to flash, that makes up for the increase in jsvars?

> Also, moving js code to flash could have freed up jsvars for the websocket buffers without even increasing beyond 1600.

I don't think that it was websocket buffers that were the issue. What was happening was the interpreter was running out of memory, and then things started to fail. The first thing that happens is timeouts fail, and things start falling apart!

wilberforce:
> Just merging - but I think there might be some confusion about what the Out of Memory error means... There are probably easier ways around it :)

Probably - I thinking I'm pushing the envelope with what can be done on an ESP8266 running Espruino - I have a pid module to clean up and submit next. I figure that on a board that has such little RAM, the more the interpreter has the better ;-)

gfwilliams:
Yes, absolutely - however in Espruino the interpreter doesn't use the heap *at all*, and only uses the stack and the memory that's in `JsVars`. The heap is basically just used by the ESP8266 side of things - so your problems stem from the number of variables available.

There are almost certainly some things you can do with your JS code that would let you cram it in to a much smaller space :)

tve:
Ughh, it's probably OK to disable LTO, but then it would have been good to revert all the makefile pre-linking that went with it, the special disable_lto #define, and to update the Makefile with some comments about the why.
When I added LTO to the linking the idea was to use -O2 or -O3 where it makes a significant difference and enables the inlining. No inlining of functions in jsutils.c without... But then we discovered that -O3 results in way too much code explosion and -O2 isn't really faster than -Os but results in more code. So I ended up with -Os where LTO doesn't provide much benefit (some sharing of string constants, probably).

wilberforce:
The testing I did removing the lto flags did very little to the files size, only adding a couple of 100 bytes per section. There is still the lto fat switch in place for the regular build, however removing this flag makes no change to the final binary sizes, so I guess it would be safe to remove. I can remove the lto flags and #define completely, if you think that is the best way to go.

wilberforce:
Increasing for 1700 to 1900 vars seems to be ok with this build:


```
process.memory();
={ "free": 601, "usage": 1299, "total": 1900, "history": 4 }
require("ESP8266").getState();
={
"sdkVersion": "1.5.0",
"cpuFrequency": 160, "freeHeap": 5704, "maxCon": 10,
"flashMap": "4MB:512/512",
"flashKB": 4096,
"flashChip": "0xe0 0x4016"
}
```

It seems stable enough to to support http fetches, and works as a web socket server, with 2 clients, so the network traffic memory usage does not appear to be hindered.

MaBecker:
"freeHeap": 5704 is too small

tve:
In order to do LTO I had to do the whole prelinking step. If you're removing LTO you should simplify the linking again: https://github.com/espruino/Espruino/blob/master/Makefile#L1705-L1712

wilberforce:
> "freeHeap": 5704 is too small
@MaBecker 

To small for what? This was the size of heap pre @tve 's updates when the vars were 1400. As has been pointed out by Gordon, there appears to be confusion over "OUT OF MEMORY" errors. These are running out of JSvars as the interpreter itself does not use Heap - only the Wifi and network side use heap.

In my testing, this amount of heap seemed sufficient - your use case will be different - Have you tried a build with 1900 vars?

tve:
Too small for reliable use. Several others and myself experienced intermittent crashes due to running out of heap. If you turn on AP and STA at the same time it will crash, etc. Part of why I spent the effort to get the jsw symbols into flash is to have enough heap available so espruino is reliable.

wilberforce:
@tve - Thanks. What do you think the minimum workable heap size is then?

I'm still getting OUT OF MEMORY with 1700 vars (even after minifying source)- trying to find the sweet spot here!

PR #867, opened 2016-06-17
Add options for make process

jumjum123:
Adds customizing options, see description in makefile.
Idea is to support userdefined make of firmware
- GENDIR
- SETDEFINES
- UNSUPPORTEDMAKE
- PROJECTNAME
- BLACKLIST
- VARIABLES


gfwilliams:
Great - thanks!

gfwilliams:
Turns out, after all that just putting a comment in rather than a print breaks the build - at least for me.

Did this work for you? I've just committed a fix so it builds on mine.

jumjum123:
Hmm, there must have been a wrong sequence in my testing.
Sorry for that.
BTW, just checked your fix, sometimes its crazy how things can be fixed ;-)

gfwilliams:
I know, it's a pretty awful fix... There didn't seem to be a nice way of doing it short of adding a no-op function (maybe the Print would have been better ;)

PR #868, opened 2016-06-23
replace jsError() with jsExceptionHere()

MaBecker:
enable try { }  and catch(err) { } for read and write to I2C devices


gfwilliams:
Great - thanks!

PR #873, opened 2016-07-09
nRF5 SDK 11.0.0 and production SoftDevices Update

mjdietzx:
Updated Nordic's SDK/SoftDevices. Note that right now I have only verified it is compiling for nRF51/nRF52. I need to test more and will let you know when I think PR is ready to merge (a few hours or tomorrow). Just want you to see work I've done incase you are doing similar stuff with the Bootloader. Also note that you should use the bootloader from the newest nRF5 sdk: https://developer.nordicsemi.com/nRF5_SDK/nRF5_SDK_v11.x.x/


mjdietzx:
Didn't see this branch: https://github.com/espruino/Espruino/commits/nrf_bootloader. Anyways we pretty much did the same thing. Only differences were I updated linkers since there were some changes and included the new fstorage section. Also had __START definition fix and removed heap from the startup files. Also I didn't include some sub directories of SDK components/ that we don't use (I think just ANT and proprietary RF).

mjdietzx:
closed because application is hard faulting and haven't found where problem is yet.

gfwilliams:
Thanks for this - Pain about the hardfault though. Is it able to start Espruino at all, or it fails even before that?

mjdietzx:
Yeah I think it's at least getting to main. I haven't found where it's hard faulting yet though

gfwilliams:
Hmm - if more RAM is used now, it could be it's lack of stack that's doing it?

mjdietzx:
The stack size: https://github.com/mjdietzx/Espruino/blob/cfc4c7788088f6199e015e2d06c81cf3ce815e5f/targetlibs/nrf5x/nrf5_sdk/components/toolchain/gcc/gcc_startup_nrf52.s#L36 is more than enough (TODO: we should calculate how much stack we actually need, and can define __STACK_SIZE in ASM defines as well has heap so we don't have to modify startup files every-time. Is this already done at all, or should I make an issue?

gfwilliams:
Well, the stack should grow down from the end of memory, and as there's no malloc in Espruino the memory area doesn't have to be protected... I think all that's needed is a safety-net in the Linker (like is there now) so that it'll fail if there isn't enough RAM? As you say, that number could be quite a bit lower though!

mjdietzx:
ah ok, yeah that sounds like something we should have

mjdietzx:
I'll try this and let you know if still hard faulting

mjdietzx:
Found bug. Cleaning up and will re-open this PR once finished/tested.

gfwilliams:
Awesome - thanks!

mjdietzx:
OK, this is good to go on my end. Tested and working nRF51 and nRF52 DKs.

gfwilliams:
Great - thanks! Just pulling it in.

PR #880, opened 2016-07-20
added STM32F411 support

notthetup:
Added Makefile, board file and image for STM32F411-DISCOVERY which is another dev board very similar to the STM32F4DISCOVERY except it uses the STM32F411 instead of the STM32F407.


gfwilliams:
Thanks! Wow - I didn't even know those existed :)

notthetup:
Yea. They're on the 2nd page of the list. I have a few lying around at work. Managed to get Espruino running on it. Thought I'd PR.

PR #881, opened 2016-07-20
cfmakeraw should run earlier in setup

ganehag:
From: http://linux.die.net/man/3/cfmakeraw

**cfmakeraw()** sets the terminal to something like the "raw" mode of the old Version 7 terminal driver: input is available character by character, echoing is disabled, and all special processing of terminal input and output characters is disabled. The terminal attributes are set as follows:
```
termios_p->c_iflag &= ~(IGNBRK | BRKINT | PARMRK | ISTRIP
                | INLCR | IGNCR | ICRNL | IXON);
termios_p->c_oflag &= ~OPOST;
termios_p->c_lflag &= ~(ECHO | ECHONL | ICANON | ISIG | IEXTEN);
termios_p->c_cflag &= ~(CSIZE | PARENB);
termios_p->c_cflag |= CS8;
```

Thus cfmakeraw needs to run before any other settings are changed.

I also added *path* to `jsvConfigObject configs[]` to allow setting the *path*.


gfwilliams:
Great - thanks!

PR #882, opened 2016-07-22
Blacklist extended for Javascript commands

jumjum123:
Javascrip commands have no class information in JSON description. This
was not handled in old version.
In this version an artificial class "--" is used to support blacklisting
them.
If you don't want to have console.log or eval, for whatever reason, this
can be blacklist now.


jumjum123:
Add option to get all variables, used during make, without compiling and linking.
Gives a lot of information to better understand make process, to find oddities, compare boards, etc.
Will be used in web-service for creating your own firmware.

gfwilliams:
I could be wrong here, but if you say "__.length" in the blacklist, does that not just blacklist every instance of `length` everywhere?

I'm not quite sure how the blacklist works, but it feels like just being able to have an empty classname might be more simple?

jumjum123:
In my Firmware creator is a tab for blacklist handling.
To get some overview a 2nd collection of tabs for all classes is used.
Since general commands don't belong to a class, I had to find a way to add
them.
Therefore I created a kind of "artificial class".
To make the story short, its grown historical.

I'm thinking about a better way. Do you think, you can add this version for
now, and change it once we have a better solution ?
Before I forget, at least in my test, length is not removed in all classes.

gfwilliams:
Yes, not problem - thanks! :)

PR #899, opened 2016-08-16
fix: http.get leak via "method" field duplicate

nailxx:
`http.get` cause a memory leak in the following scenario:

```javascript
var opts = {
  host: 'lala.com',
  patch: '/'
};

setInterval(function() {
  http.get(opts, function(res) { ... })
}, 1000);
```

Every tick `opts` would get an additional copy of `opts.method = "GET"` field.


gfwilliams:
Nice - thanks!

PR #903, opened 2016-08-23
Remove incorrect line number in documentation

yerpj:
I just removed the static reference to line 606 of Makefile, which has been modified since. Thus, the line number is not relevant anymore. I tried to find a good location for the addition of two lines, which is just before the "ifdef ESPRUINO_1V3" section


gfwilliams:
Thanks! Yeah, a static line number reference wasn't super helpful :)

PR #904, opened 2016-08-23
Serial6 is now exposed as an object on NUCLEOF401RE & NUCLEOF411RE

yerpj:
The STM32F4x1RE microcontrollers embeds 3 USART (USART1,USART2,USART6), but the current board descriptions for NUCLEO401RE and NUCLEO411RE are only exposing 3 Serial objects , respectively Serial1, Serial2, Serial3. This PR allows using Serial1, Serial2 and Serial6 as usual. Even if Serial[3..5] are potentially existing objects, they can not be used as they are not physically existing.

### DIFF
--- a/boards/NUCLEOF401RE.py
+++ b/boards/NUCLEOF401RE.py
@@ -38,7 +38,7 @@ chip = {
   'ram' : 96, # 0x0001 8000 long, from 0x2000 0000 to 0x2001 7FFF
   'flash' : 512, # 0x0008 0000 long, from 0x0800 0000 to 0x0807 FFFF
   'speed' : 84,
-  'usart' : 3,
+  'usart' : 6,
   'spi' : 4,
   'i2c' : 3,
   'adc' : 1,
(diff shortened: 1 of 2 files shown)


gfwilliams:
Ok, thanks - however it will produce some unhelpful error when using Serial3-5? I guess it's better than not having Serial6.

yerpj:
From my understanding, the build process creates as many Serials as there is usarts, as stated in the board definition file ``` 'usart' : x, ``` . If x is 1, there will be only Serial1. If x is 5, there will be Serial1, Serial2, Serial3, Serial4, and Serial5. If this is true, then we cannot really expose Serial6 without exposing dummy Serials[3..5].

gfwilliams:
Well, it should be possible to change the definition at https://github.com/espruino/Espruino/blob/master/src/jswrap_serial.c#L109 from ` "#if" : "USART_COUNT>=3"` to ` "#if" : "USART_COUNT>=3 && !NO_USART5"` and then change the Makefile for the Nucleos - but changing the usart count is the quickest way to get Serial6, it just means people will get some odd errors :)

PR #908, opened 2016-08-29
flash_read_doc_fix

muet:
addr parameter description of Flash.read() fixed to read ...'reading' from (vs. ...'writing' from).


gfwilliams:
Great, thanks!

PR #911, opened 2016-09-02
added checksum for "GO" command in Espruino bootloader

yerpj:
As for command "WRITE", the command "GO" now computes the checksum before jumping to the given address.

### DIFF
--- a/targets/stm32_boot/main.c
+++ b/targets/stm32_boot/main.c
@@ -156,8 +156,12 @@ int main(void) {
               addr |= _getc_blocking()  << 16;
               addr |= _getc_blocking()  << 8;
               addr |= _getc_blocking();
+              chksumc = ((addr)&0xFF)^((addr>>8)&0xFF)^((addr>>16)&0xFF)^((addr>>24)&0xFF);
               chksum = _getc_blocking();
-              // TODO: check checksum
+              if (chksumc != chksum) {
+                _putc(NACK);
+                break;
+              }
               _putc(ACK);
               setLEDs(7); // jumping...
               unsigned int *ResetHandler = (unsigned int *)(addr + 4);


gfwilliams:
Great, thanks! Just checking - has this been tested?

yerpj:
Not yet, unfortunately. 
I don't really know how to flash the bootloader into Espruino... I just compiled it with 
```PICO_1V3=1 BOOTLOADER=1 RELEASE=1 make``` and I suspect I have to flash it with ST-LINK at address 0x08'00'00'00, right?

**Comment by gfwilliams** [comment]
See http://www.espruino.com/Pico#advanced-reflashing

But this is why it never got added in the first place - I needed to be able to test it with an ST tool. Given USB has error correction in it anyway, the chances of the CRC ever being wrong are minimal.

It's absolutely **vital** changes to the bootloader are tested and don't break stuff. If I get 1000 boards made that can't have the firmware updated and I have to take them all back, reflash them, and ship them out again then that's basically a whole year's earnings gone for me.

**Comment by yerpj** [comment]
dfu-util log:
```
dfu-util --alt 0 -D bootloader.dfu 
dfu-util 0.5

(C) 2005-2008 by Weston Schmidt, Harald Welte and OpenMoko Inc.
(C) 2010-2011 Tormod Volden (DfuSe support)
This program is Free Software and has ABSOLUTELY NO WARRANTY

dfu-util does currently only support DFU version 1.0

Opening DFU USB device... ID 0483:df11
Run-time device DFU version 011a
Found DFU: [0483:df11] devnum=0, cfg=1, intf=0, alt=0, name="@Internal Flash  /0x08000000/04*016Kg,01*064Kg,03*128Kg"
Claiming USB DFU Interface...
Setting Alternate Setting #0 ...
Determining device status: state = dfuERROR, status = 10
dfuERROR, clearing status
Determining device status: state = dfuIDLE, status = 0
dfuIDLE, continuing
DFU mode device DFU version 011a
Device returned transfer size 2048
Dfu suffix version 11a
DfuSe interface name: "Internal Flash  "
file contains 1 DFU images
parsing DFU image 1
image for alternate setting 0, (1 elements, total size = 10944)
parsing element 1, address = 0x08000000, size = 10936
done parsing DfuSe file
```

Espruino IDE console log (only last lines), consisting of the last few lines of the flashing Espruino flashing process, followed by the connection to it and a rapid test 
">1+1
=2
> "
```
Sent
Writing 256 bytes at 0x805bf00...
Sending block "1Î", wait 50ms
Sent
Sent
Sending block "\b\u0005¿\u0000²", wait 50ms
Sending block "ÿ\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000ÿ", wait 50ms
Sent
Sent
Writing 32 bytes at 0x805c000...
Sending block "1Î", wait 50ms
Sent
Sending block "\b\u0005À\u0000Í", wait 50ms
Sending block "\u001f\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000\u0000@\u0001\u0000 \u0001\u0000\u0000\u0000", wait 50ms
>>> Write complete!
Set Slow Write = false
>>> Flashing Complete
Sent
Disconnect callback...
>>> Disconnected
Sent
Sent
Sent
Sent
Sent
Sent
Sent
Sent
Sent
Sent
Sent
Sent
>>> Connecting...
Set Slow Write = true
[object Object]
Connected [object Object]
Found a prompt... great!
>>> Sending...
---> "echo(0);\nconsole.log(\"<\",\"<<\",JSON.stringify(process.env),\">>\",\">\");echo(1);\n"
Sending block "echo(0);\nconsol", wait 50ms
Sent
Sending block "e.log(\"<\",\"<<\",", wait 50ms
Sent
Sending block "JSON.stringify(", wait 50ms
Sent
Sending block "process.env),\">", wait 50ms
Sent
Sending block ">\",\">\");echo(1)", wait 50ms
Sent
Sending block ";\n", wait 50ms
Got "echo(0);\r\n< << {\"VERSION\":\"1v86\",\"BUILD_DATE\":\"Jul  6 2016\",\"BUILD_TIME\":\"15:44:29\",\"GIT_COMMIT\":\"7450cab7e6b18274deb47cb8940c936f1a9b44e0\",\"BOARD\":\"PICO_R1_3\",\"CHIP\":\"STM32F401CDU6\",\"CHIP_FAMILY\":\"STM32F4\",\"FLASH\":393216,\"RAM\":98304,\"SERIAL\":\"27005600-13513430-36363435\",\"CONSOLE\":\"USB\",\"EXPORTS\":{\"jsvLock\":230441,\"jsvLockAgainSafe\":230427,\"jsvUnLock\":230401,\"jsvSkipName\":86049,\"jsvMathsOp\":93433,\"jsvMathsOpSkipNames\":93495,\"jsvNewFromFloat\":230837,\"jsvNewFromInteger\":230873,\"jsvNewFromString\":233745,\"jsvNewFromBool\":230857,\"jsvGetFloat\":86641,\"jsvGetInteger\":84249,\"jsvGetBool\":92025,\"jspeiFindInScopes\":105561,\"jspReplaceWith\":111961,\"jspeFunctionCall\":99169,\"jspGetNamedVariable\":95029,\"jspGetNamedField\":97117,\"jspGetVarNamedField\":96709,\"jsvNewWithFlags\":230677}} >> >\r\n=undefined\r\n>"
Loading http://www.espruino.com/json/PICO_R1_3.json
Device found (connectionId=10)
>>> Connected to port COM43
Sent
>>> Sent
Board JSON loaded
Firmware >1.43 supports faster writes over USB
Set Slow Write = false
FIRMWARE: Current 1v86, Available 1v86
Sending block "1", wait 50ms
Sent
Sending block "+", wait 50ms
Sent
Sending block "1", wait 50ms
Sent
Sending block "\r", wait 50ms
Sent
```
I never used dfu-util but it seems that flashing the new bootloader succeeded. After that, I still could upgrade the Espruino firmware as well as interract trough the console. 

@gfwilliams what kind of tests do you usually require in order to validate a piece of code?

**Comment by gfwilliams** [comment]
Cool, well, it's working then.

> what kind of tests do you usually require in order to validate a piece of code?

Well, something that actually executes that code would be good. The normal bootloader doesn't use CMD_GO at all.

You'd need to come up with some test case where you used (ideally ST's bootloader application) to send a CMD_GO, so you could be sure it had sent the correct CRC.

**Comment by yerpj** [comment]
I just proceeded to some basic tests:
1) powering up a Pico while BTN is pressed
2) entering bootloader by sending 0x7F
3) requesting some infos with CMD_GET
4) requesting ID with CMD_GET_ID
5) trying to exit bootloader and jumping to address 0x08010000 which seem to be the start address of Espruino, right?

capture d ecran 2016-09-12 17 09 36

After that sequence, the Pico LEDs are both ON ( as asked on line 166 ```setLEDs(7); // jumping...``` ) but nothing is happening. It seems that the Espruino base address is wrong OR some extra config has to be done before jumping to user code... Any idea?

**Comment by gfwilliams** [comment]
The binary starts at 0x08010000, but that's not where the entrypoint is. The *address of that* is stored in location 0x08010004

Even if you do that it may fail, since USB is now already initialised. Your best bet it to write some code into RAM and then jump to that (but you'll have to add 1 to the address since it's Thumb).

But I can't help any more. The whole reason I didn't add those few lines at the beginning was it was something I didn't want to have to test out, because it's a total pain. The last thing I want to do is to spend more time than I'd have spent myself, showing you how to test it out. *Especially* when it's a part of the bootloader that doesn't get used by anything at the moment.

**Comment by yerpj** [comment]
Your point of view makes sense. It won't help much to add this checksum.

**Comment by gfwilliams** [comment]
Yeah, sorry - it's just there are cases where it could be used (for instance if I ever have to replace the bootloader from the bootloader itself), and it'd be a disaster if it turned out to be broken. By contrast the consequences of it not checking the CRC are actually pretty small.

Actually looking at the code and your test, despite it crashing, you've basically proved it's passed the CRC and works... Sorry about that - I'll pull it in anyway.

**Comment by yerpj** [comment]
Hello again,

FYI, I managed to get the CMD_GO work like a charm. I made modifications on bootloader code main.c and utils.c/utils.h 

main.c: (updated CMD_GO code)
```c 
case CMD_GO: // read memory
              _putc(ACK);
              addr = _getc_blocking() << 24;
              addr |= _getc_blocking()  << 16;
              addr |= _getc_blocking()  << 8;
              addr |= _getc_blocking();
              chksumc = ((addr)&0xFF)^((addr>>8)&0xFF)^((addr>>16)&0xFF)^((addr>>24)&0xFF);
              chksum = _getc_blocking();
              if (chksumc != chksum) {
                _putc(NACK);
                break;
              }
              _putc(ACK);
              setLEDs(7); // jumping...
              deInitHardware();
              jumpToApplication(addr);
              break;
```

utils.c : (added 2 functions)

```c
typedef void (*pFunction)(void);

bool jumpToApplication(unsigned int addr) {
  uint32_t JumpAddress = *(__IO uint32_t*)(addr + 4);
  pFunction Application;
  if (JumpAddress==0 || JumpAddress==0xFFFFFFFF)
    return false;
  Application=(pFunction) JumpAddress;
  __set_MSP(*(__IO uint32_t*) addr);
  Application();
  return true; // should never get here
}

void deInitHardware() {
#if defined(STM32F3)
 RCC_APB1PeriphClockCmd(RCC_APB1Periph_PWR, DISABLE);
 RCC_AHBPeriphClockCmd( RCC_AHBPeriph_GPIOA |
                        RCC_AHBPeriph_GPIOB |
                        RCC_AHBPeriph_GPIOC |
                        RCC_AHBPeriph_GPIOD |
                        RCC_AHBPeriph_GPIOE |
                        RCC_AHBPeriph_GPIOF, DISABLE);
#elif defined(STM32F2) || defined(STM32F4)
 RCC_APB2PeriphClockCmd(RCC_APB2Periph_SYSCFG, DISABLE);
 RCC_AHB1PeriphClockCmd(RCC_AHB1Periph_GPIOA |
                        RCC_AHB1Periph_GPIOB |
                        RCC_AHB1Periph_GPIOC |
                        RCC_AHB1Periph_GPIOD |
                        RCC_AHB1Periph_GPIOE |
                        RCC_AHB1Periph_GPIOF |
                        RCC_AHB1Periph_GPIOG |
                        RCC_AHB1Periph_GPIOH, DISABLE);
#else
 RCC_APB1PeriphClockCmd(RCC_APB1Periph_PWR, DISABLE);
 RCC_APB2PeriphClockCmd(
       RCC_APB2Periph_ADC1 |
       RCC_APB2Periph_GPIOA |
       RCC_APB2Periph_GPIOB |
       RCC_APB2Periph_GPIOC |
       RCC_APB2Periph_GPIOD |
       RCC_APB2Periph_GPIOE |
       RCC_APB2Periph_GPIOF |
       RCC_APB2Periph_GPIOG |
       RCC_APB2Periph_AFIO, DISABLE);
#endif
  
NVIC_DisableIRQ(OTG_FS_IRQn);
}
```

I compiled for the Pico 1v3.
Then, the test protocol is as following:
1) make sure to store in flash a visual application (in my case, a function that turns ON and OFF both LEDs each 2 seconds)
2) power ON the Pico while pressing the BTN
3) enter bootloader (means sending 0x7F)
4) request to jump to user code (means sending 0x21, 0xDE)
5) wait for ACK
6) send the user addr (means sending 4 bytes followed by the 1-byte checksum)

I tested with wrong checksums, then the bootloader sends a NAK and the command is aborted. Then I tested with a correct checksum (0x09, for the address 0x08010000) and the user app is correctly invoked. 

I do not particularly want you to merge anything about this update, as I understand that it is currently not very helpful for the Espruino project. I wrote this comment mainly to centralize the knowledge.

Anyway, if you care about this new piece of code, I still can make a new PR.

gfwilliams:
Thanks! Wow, yeah, I just looked up GO and the bootloader's implementation was completely wrong :)

When you've done the command and jumped to Espruino, does USB come back up ok? For the Pico I'd added the ability to jump back out of the bootloader if the button was held down for a few seconds, but I never got it to work reliably because of the issue of resetting USB.

But yes, It'd be great to pull that code back in.

One small gotcha - I think `pFunction Application` should really be static. There's a post here: http://stackoverflow.com/questions/16490315/stm32f4-memory-jumping

Where they suggest that because `__set_MSP` resets the stack, `Application` is stored on the stack and so won't be there any more. In reality it's not an issue because it'll be in a register at that point (if optimisations are enabled) but if the compiler changes (or you compile in debug mode?) it might break. 

But `__set_MSP` is really cool though - I hadn't seen that before. It'd be nice to use `jumoToApplication` in place of `jumpToEspruinoBinary` for the main bootloader. I actually hit that problem yesterday (I'd flashed the WiFI boards with a 96kB RAM bootloader and spent ages trying to debug why I couldn't get a 128kB RAM Espruino binary to work :)

gfwilliams:
Just looked at the original - at least it wasn't as wrong as I'd imagined :) Any idea why it didn't work before? It kind of feels like it should have, even if the stack would have been in slightly the wrong location.

yerpj:
> When you've done the command and jumped to Espruino, does USB come back up ok?

I am still working on that, I wonder if this is a device problem or a host problem. I will try to add a delay between the USB_DeInit process and the jumpToApplication process. Or maybe the USB DeInit has to be done in a better way, don't know yet.

> One small gotcha - I think pFunction Application should really be static. There's a post here: 
> http://stackoverflow.com/questions/16490315/stm32f4-memory-jumping

I totally agree. In my case it is a global variable, but protecting it with a static constraint is a good idea.

gfwilliams:
> I wonder if this is a device problem or a host problem

It is probably a host problem - they don't like it when USB disconnects and then reconnects quickly. A small delay after the deinit might fix it...

> In my case it is a global variable

Ahh, perhaps you changed that after pasting the code in?

yerpj:
>It is probably a host problem - they don't like it when USB disconnects and then reconnects quickly. A small delay after the deinit might fix it...

i will try that soon.

>Ahh, perhaps you changed that after pasting the code in?

I must be a bit tired, the declaration was totally local... I will change it into a static var.

yerpj:
>It'd be nice to use jumoToApplication in place of jumpToEspruinoBinary for the main bootloader.

Does it mean you'd prefer to call jumpToApplication(0x08010000) instead of jumpToEspruinoBinary() by replacing each call?

gfwilliams:
I think so, yes. Perhaps there could be another boolean argument (deInitUSB) that was true *apart* from when the bootloader called jumpToApplication right at the start if the button wasn't pressed (so there was no delay for the normal bootloader).

yerpj:
I'll first validate the deInit/unmount process in order to allow a proper re-enumeration of the virtual COM port, and then I will re-implement the jumpToApplication(<ADDR>,<deInitUSB>) function.

gfwilliams:
Great - thanks for putting all the work into this! (By the way, I tweaked the bootloader this morning to move the button held test to before USB is initialised - it might be an idea to git pull)

yerpj:
> (By the way, I tweaked the bootloader this morning to move the button held test to before USB is initialised - it might be an idea to git pull)

```c

// if button is not set, jump to the address of the binary
  if (!isButtonPressed()) {
    jumpToEspruinoBinary();
    // we could return here - binary might be very obviously corrupted
  }

```
This piece of code is executed before any USB initialization (inside ```initHardware()```), I don't really understand what you modified this morning?

gfwilliams:
It's at the bottom of this change: https://github.com/espruino/Espruino/commit/abe11d37d8dadd12d931a7c1da4a870a31d91588

There was other code in main.c that tried to detect when the button had been held down for a certain amount of time - but by the time that has done its thing, USB was initialised. Instead, it now just blocks with the green LED on while the button is pressed (or if >3 secs it drops through to Espruino)

yerpj:
OK I'll give up trying to re-enumerate the USB after a jump to application. I mostly tried to make a soft-disconnect :
```c 
// set bit SDIS in OTG_FS_DCTL reg
    USB_OTG_dev.regs.DREGS->DCTL |= 0x02;
```
Even if I came accross some forum posts speaking about successful re-enumeration through this soft-disconnect, no way of replicating it on the Pico :-/

I am sorry about the issue of this PR, I will certainly have another try in a few days because I don't like to give up , but in a near future I would prefer to have a try on another target such as a STM32F4DISCOVERY, as it will be much easier to debug through JTAG.

gfwilliams:
No problem - thanks for all your help!

With the disconnect - I guess it's also important it works on the STM32F1 as well... :s But yes, it should be easy enough to compile the bootloader for the discovery boards, and it would make debug a lot easier!

Issue #920, opened 2016-09-12
fillCircle() Filled circles in Graphics library 


owenmcateer:
A nice to have addition to the graphics library would be a filled circle _Graphics.fillCircle(x, y, rad)_
Here is my proposed code.

gfwilliams:
Nothing seems to be attached?

Ideally whatever solution there was would use the fillRect call for each scanline (rather than drawPixel)

owenmcateer:
Here is my idea. https://github.com/espruino/Espruino/pull/921
This is un-tested and using _graphicsFillRectDevice()_ may not be the fasted way of doing so.

gfwilliams:
not sure if it's fastest, but it's not bad and a lot quicker than doing it in JS - thanks!

PR #921, opened 2016-09-12
920/fill circle

owenmcateer:
https://github.com/espruino/Espruino/issues/920


gfwilliams:
Ok, merged. Needed a few tweaks first to get it to compile, but once compiling it works

PR #924, opened 2016-09-16
Boot v1.5 and SDK 1.5.4

wilberforce:
This PR updates the ESP8266 bootloader to 1.5, and the SDK to 1.5.4


nkolban:
I think you might want to check out #925 as well as this seems to upgrade to SDK 2.0.0

PR #933, opened 2016-10-03
Issue 915

ddm:
Implemented `NRF.updateServices` to allow BLE notifications (indication not available yet).

Initial tests indicate this works but feedback is more than welcome.

### DIFF
--- a/boards/NRF52832DK.py
+++ b/boards/NRF52832DK.py
@@ -22,7 +22,7 @@ info = {
  'default_console' : "EV_SERIAL1",
  'default_console_tx' : "D6",
  'default_console_rx' : "D8",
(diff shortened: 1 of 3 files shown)


gfwilliams:
Great, thanks! Please can you just remove the change to the default baud rate on NRF52832DK? Espruino tools generally default to 9600, so when boards don't use it it's a massive support headache as everyone starts complaining that nothing is working.

ddm:
Sure thing sorry about that. I use 115200 in my test setup because it plays well with the buspirate but I forgot to revert it when I pushed.
The other changes are mainly to suppress warnings in the file I was working in or comments which were addressed by the PR.

gfwilliams:
Great - thanks!

Issue #915, opened 2016-09-08
Updating BLE data for GATT Characteristics


gfwilliams:
See here:

http://forum.espruino.com/conversations/292530/#comment13202382

Looks like `NRF.updateServices({ ... })` is the easiest/least confusing for users.

ddm:
Just to make sure I understand things correctly: `updateServices` should take an object with a structure similar to the `setServices` but with the following caveat: only `value` (and maybe `maxLen`?) should be taken into account (I don't think reconfiguring services should be covered by it since calling `setServices` again serves that purpose already)

When compared to the simple `write` example you gave on the [forum](http://forum.espruino.com/comments/13202256/), `updateServices` has the advantage that several values can be updated "atomically". This could be convenient if these values are somehow correlated.
The downside is that the argument has to be correctly documented in order not to confuse users who are familiar with `setServices` (it will be a similar looking structure but with a different purpose).

ddm:
A corollary is that maybe `updateServices` should only allow existing services and characteristics to be updated. If so should we drop the unwanted entries of throw an error?
The simple `write`  avoids all these issues since only existing values can be accessed but at the cost of the atomic updates of multiple values.

gfwilliams:
Yes, that all sounds good.

I'd iterate through the structure in a similar way to setServices, but then if you can't find the relevant UUID just throw an error.

For now it would be easier to just ignore any non-'value' entries, and it has the benefit that a user could keep the structure they had used previously, update the value element, and re-use it.

ddm:
Good idea.
I guess I have everything I need to give it a try :)

gfwilliams:
I just pushed some changes that fix issues related to setServices. It's work doing a pull because it might save you some problems when debugging :)

ddm:
I just pulled the changes thanks!

gfwilliams:
Did you ever have any luck with this, or should I push on with it?

ddm:
Sorry, I should have touched base before I went on holiday. I am currently away until the end of the month and I don't have my dev boards with me. I still had a few issues with the code I had written before I left but, for the most part, the call to `sd_ble_gatts_hvx` seemed to work. I'll be back at the office  in October.

ddm:
I think I am going to need a little help here... I pushed some working code on my repo: https://github.com/espruino/Espruino/compare/master...ddm:issue-915

[...]

PR #934, opened 2016-10-04
Enable simple READs using NRF.updateServices

ddm:
This improves READ supports related to https://github.com/espruino/Espruino/issues/915.

It is now possible to update a characteristic even when no connection is active. Notifications must be explicitly requested by using `notify: true` when calling `NRF.updateServices`.

Documentation has been updated to reflect the changes.


gfwilliams:
Nice - thanks!

PR #942, opened 2016-10-11
add a function to lock console

lancernet:
E.lock() .it will be lock console if you set password


gfwilliams:
Hi,

Thanks - but it seems there's something wrong with the pull request. Looks like you just added `jswrap_espruino.c` as an extra file in the main directory, when it should be in `src/jswrap_espruino.c`. There is also no change to the `.h` file and it seems like you may have deleted the repository as well so I can't merge manually.

I'm copy/pasting the `E.lock` function now - but am renaming it to `E.lockConsole` and adding some documentation.

PR #944, opened 2016-10-12
fix bug with jswarp_pin_writeAtTime

lancernet:


gfwilliams:
That's the intended behaviour. It's so you can schedule things at the same time (since the JS takes a while to execute).

If you want to set the pin state after a delay, you cn just use digitalPulse

lancernet:
thank you

gfwilliams:
No worries - thanks for contributing, but sorry I couldn't accept it this time

PR #950, opened 2016-10-16
add new pin mode:"auto"

lancernet:
auto set pin mode at next using

### DIFF
--- a/src/jswrap_io.c
+++ b/src/jswrap_io.c
@@ -377,7 +377,7 @@ void jswrap_io_pinMode(
     jshPinSetState(pin, m);
   } else {
     jshSetPinStateIsManual(pin, false);
-    if (!jsvIsUndefined(mode)) {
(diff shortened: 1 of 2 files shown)


gfwilliams:
Hi - what was the reason behind this? Do you think that the existing method of using 'undefined' like it says in the docs is confusing?

lancernet:
I am in the implementation of the project for some reason recently, the programmer previously left the job, new employees do not understand the hardware, when using pinMode, often do not understand the use of undefined parameter, ports can be configured to what mode. So I added a new mode auto, my English is not very good, so use the translation software. All of the above content comes from the translation software

gfwilliams:
Ok, thanks for the explanation - it's a good translation!

I know it is difficult given the language, but in future it would really help if you could try and document the new arguments in the `JSON` area above. I'll do it this time though.

PR #967, opened 2016-11-07
Esp32 RTOS functions wrapper and a task for console read

jumjum123:
- added wrapper to use functions around tasks and queues. rtosutils is
c-level, jswrap_rtos wraps utils to Espruino
- added option for rtos (RTOS=1) to makefile
- adding a task for reading uart and writing to a queue
- use RTOS-option to switch between old polling and new queue based
input
- docu for the changes


jumjum123:
With new version of ESP-IDF we run into problems with Js command save(). This need to be fixed first. Could be a problem with interrupts (?)

gfwilliams:
Could be watchdog timer if `save()` takes too long? Or it could just be writing to the wrong addresses and wiping itself out.

PR #975, opened 2016-11-13
Update vagrant and makefile

clowrey:


clowrey:
I am still learning how GIT / GitHub works so I may have not done this entirely correct...

gfwilliams:
Looks good - thanks! In reality I actually use a newer compiler than the one you get from the ppa (same link as was in there before, just a newer tar file) but the one that's there should work, and hopefully won't need updating!

Just before I merge - did you really need `python27` installed? Or was that while you were trying to sort out why hexmerge wouldn't run?

clowrey:
You're right you can remove the python27 addition, I later found out it was already installed in the default vagrant setup.

PR #990, opened 2016-12-06
Corrects flashing memory address

IrregularShed:
The instructions at http://www.espruino.com/ESP8266_Flashing are correct - this file just needs a tweak.

### DIFF
--- a/targets/esp8266/README_flash.txt
+++ b/targets/esp8266/README_flash.txt
@@ -11,7 +11,7 @@ esptool.py --port [/dev/ttyUSB0|COM1] --baud 115200 write_flash \
 esptool.py --port [/dev/ttyUSB0|COM1] --baud 115200 write_flash \
   --flash_freq 80m --flash_mode qio --flash_size 32m \
   0x0000 "boot_v1.6.bin" 0x1000 espruino_esp8266_user1.bin \
-  0x3FC000 esp_init_data_default.bin 0x37E000 blank.bin
+  0x3FC000 esp_init_data_default.bin 0x3FE000 blank.bin
 
 For 1MB flash use 0xFC000/0xFE000 and for 2MB flash use 0x1FC000/0x1FE000 on the last line.
 


gfwilliams:
I think if you look at the line below the one you changed, it tells you what you have to change it to for different devices.

The code that's there is for the 512kB ones. Perhaps change the line you said (as yours is most common), but also change the docs below to say about the 512k ones too.

IrregularShed:
It does... but that's not what the problem is :) on a 4MB one with the existing addresses you would end up overwriting a chunk of Espruino code and data with `blank.bin` (I know because I just did it while trying out the NodeMCU flasher)!

gfwilliams:
You mean saved JS code, or the Espruino interpreter?

512k and 1M parts use the same underlying binary, so I don't see how it could be overwriting the interpreter.

The problem is that when you update the firmware you're pretty much guaranteed that the saved code will no longer work, so you really do *want* to overwrite it. If you reference *any* built-in function or object in your code, it'll store its address... But then on a new firmware the addresses will be different.

opichals:
By using the change in this PR it appears it fixed my problems described here https://github.com/espruino/Espruino/issues/642#issuecomment-274273961

gfwilliams:
But does this work on the 512kB parts?

I know there is a comment immediately after the change which talks about 1M parts. Maybe we need to explicitly copy the flashing command and say: This one is for 512k, this one is for 1M.

opichals:
Not sure what you mean.

The file states above that the particular change concernes the 4Mb ESPs and it does note the correct addresses for other variants.

gfwilliams:
So you're saying that with the changes suggested, the command *will still work for flashing the 512k ESP8266?*

opichals:
I think it will, as the command for 512k ESPs is at line 4 and is not touched in the PR (https://github.com/IrregularShed/Espruino/blob/c8f3a67e6012cef9bdfd566dd2c5fc566f6eb121/targets/esp8266/README_flash.txt#L4).

The 4M command is at line 11 https://github.com/IrregularShed/Espruino/blob/c8f3a67e6012cef9bdfd566dd2c5fc566f6eb121/targets/esp8266/README_flash.txt#L11

gfwilliams:
Cool - sorry, that was dumb - I missed that :(

PR #994, opened 2016-12-13
Nucleo l476 rg

VVESTM:
Hello,

Here is a first pull request to support Espruino on ST NucleoL476RG board with Low Layers API.
I have created a new stm32_ll directory to deal with low layer interface as it is slightly different from previous SPL.

To compile it, just use : 
    NUCLEOL476RG=1 RELEASE=1 make
As this board embeds a ST-Link, you can also use DEBUG=1 option.

The port is not fully functional yet but I prefer to share it now. It is mainly missing I2C, SPI and JshAnalog functions.
Anyway, it is possible to use USART, Flash, digital input, output, timers, IRQ, ...

Gordon, I would appreciate to have your feedback on this work when you get time. 
For information, I am off during Christmas period. 

Regards
Vincent


gfwilliams:
Hi,

This looks good - really clean considering you've had to add a whole new set of libs.

Some things I noticed:

* The Adafruit LCD stuff seems a bit odd to add here - and I'm not sure it's referenced anywhere? If it was used it should probably go into `libs/graphics` anyway
* I'd also try and avoid committing the DSP lib examples and CMSIS documentation - although I realise I may have left some in for other platforms :)
* There's already some code to force 32 bit writes to flash here: https://github.com/espruino/Espruino/blob/d228d53cd26f609b0a321c2603df58b6731368de/src/jswrap_flash.c#L409  - if you just made that 7 calls instead of 3 you could avoid having to hack around with the heatshrink compressor I think?
* It would be kindof nice to ifdef it so other platforms still used 4 byte alignment - I'm not sure how much difference it makes but it definitely makes regressions less likely :)
* Pretty much all the modules that use the `flash` library will assume a 32 bit write, so won't work... I wonder whether it's worth trying to emulate 32 bit writes in jshardware.c so that 32 bit writes just pad the value with 0xFFFFFFFF (which shouldn't change the flash state for the other word).

But yeah, looks great! Hope you and Fred have a good Christmas!

VVESTM:
Hi,
Thanks for your feedback Gordon.

Initially, I add quite a lot of things in targetlibs/stm32l4 in order to ease the merge with new FW version. But you're right, it is too much. I will clean this directory...

For the flash memory alignment, I will effectively add a switch to not impact other platforms.
The patch you propose is not sufficient. With L4, each write needs to be 64 bits, we cannot write a 32 bits to the flash.
The patch you propose will align the last write on a 64 bits but it needs to be done at each write. Perhaps something can be done in heatshrink_encode ?

If I use padding, it means I will loose half of the memory... :(

I will provide you a new patch set once I integrate your remarks.
By the way, if I remember well, you have some samples of this NucleoL476RG board ? If it is not the case, tell me...

Regards
Vincent

gfwilliams:
Thanks - yes, I seem to have a 476 Nucleo here.

> The patch you propose will align the last write on a 64 bits but it needs to be done at each write.

I still don't think you need to modify heatshrink unless I am missing something. All of the flash writing is done by `jsfSaveToFlash_writecb` (and there is already 32 bit alignment code) - so all you need to do is tweak it to align to 64 bits instead (which you've done).

With your changes it will already align writes to 64 bits, and by putting 7x jsfSaveToFlash_writecb at the end you then ensure that the final 64 bits is always written.

I mean, you've added this to heatshrink:

```
       while(polled%8 != 0){
        /*  in this case, fill the memory with 0x0
            (data will be written in flash only on 64 bits multiples) */
        callback(0x0, cbdata);
        polled++;
      }
```

But that is no different from calling `jsfSaveToFlash_writecb(0,cbData);` 7 times - *it's exactly the same callback*, and I am already doing it to align to 32 bits - without having to hack something into a 3rd party library.

VVESTM:
OK,

I misunderstood your first post. I thought you want to remove all my patch related to flash.
In fact, only the patch done in heatshrink needs to be replaced by your proposal.
I agree in this case.

Thanks for feedback. I will push a new set of patch.
Vincent

gfwilliams:
Great - yes, I agree there isn't really a way around it.

VVESTM:
Hello Gordon,

First of all, I wish you an happy new year. I hope this year will be full of success for you and Espruino !!!

I have updated my branch with your remarks. Basically, I have : 
    removed unused files from targetlibs/stm32l4/lib/
    added a switch to manage 64 bits alignment for flash

Regards
Vincent

gfwilliams:
Great - thanks! Merging :)

VVESTM:
Thanks Gordon.
Will push later for the missing components (SPI, I2C and analog)

PR #998, opened 2016-12-15
fix jshGetPinString pin undefined bug

lancernet:
thanks


gfwilliams:
Great - thanks! I'm leaving this open for now because I need to update the comments for jshGetPinString and also check that everywhere it's called the buffer is the correct size (I guess it should be 10, because it needs the trailing 0 as well)

PR #1003, opened 2016-12-19
fixed JshPinState #defines have changed bug

lancernet:
i tested it,


MaBecker:
Hi Gordon,

as far as I remember,  this is already fixed with PR #992

gfwilliams:
Yes, I think so.

It's an annoying one as (I think) pinStateToString must only be for debugging purposes anyway :)

PR #1017, opened 2017-01-09
update FatFs (C)ChaN from R0.10a to R0.10c

hmut:
I've created the diffs for the filesystem version upgrade as suggested by Wilberforce in the forum. The changes probably does't solve any reported issues on ESP8266, but five bugs in the lib and adds a feature. From changelog:

Fixed a hard error in the disk I/O layer can collapse the directory entry.
Fixed LFN entry is not deleted on delete/rename an object with lossy converted SFN.
Added a configuration option for the platforms without RTC. (_FS_NORTC)
Fixed volume label created by Mac OS X cannot be retrieved with f_getlabel().
Fixed a potential problem of FAT access that can appear on disk error.
Fixed null pointer dereference on attempting to delete the root direcotry.

### DIFF
--- a/libs/filesystem/fat_sd/ff.c
+++ b/libs/filesystem/fat_sd/ff.c
@@ -1,5 +1,5 @@
 /*----------------------------------------------------------------------------/
-/  FatFs - FAT file system module  R0.10a                (C)ChaN, 2014
+/  FatFs - FAT file system module  R0.10c                (C)ChaN, 2014
 /-----------------------------------------------------------------------------/
 / FatFs module is a generic FAT file system module for small embedded systems.
 / This is a free software that opened for education, research and commercial
@@ -111,6 +111,12 @@
 /                   Fixed f_close() invalidates the file object without volume lock.
 /                   Fixed f_closedir() returns but the volume lock is left acquired.
 /                   Fixed creation of an entry with LFN fails on too many SFN collisions.
+/ May 19,'14 R0.10b Fixed a hard error in the disk I/O layer can collapse the directory entry.
+/                   Fixed LFN entry is not deleted on delete/rename an object with lossy converted SFN.
+/ Nov 09,'14 R0.10c Added a configuration option for the platforms without RTC. (_FS_NORTC)
+/                   Fixed volume label created by Mac OS X cannot be retrieved with f_getlabel().
+/                   Fixed a potential problem of FAT access that can appear on disk error.
+/                   Fixed null pointer dereference on attempting to delete the root direcotry.
 /---------------------------------------------------------------------------*/
 
 #include "ff.h"			/* Declarations of FatFs API */
@@ -125,7 +131,7 @@
 
 ---------------------------------------------------------------------------*/
 
-#if _FATFS != 29000	/* Revision ID */
+#if _FATFS != 80376	/* Revision ID */
 #error Wrong include file (ff.h).
 #endif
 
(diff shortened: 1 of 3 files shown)


gfwilliams:
Great, thanks! Just checking - has this been tested at all?

hmut:
I've tested superficially on esp8266. The build is too large on other platforms?

gfwilliams:
By superficially, do you mean you've tried it with an SD card, or just got it compiling?

On other platforms, you need to do a compile with RELEASE=1

I'm a bit busy at the moment but I'll give this a go on the Original and Pico boards when I get a chance - thanks!

hmut:
I tried with an SD card (only fs.readdir("/")) successfully with #define _WORD_ACCESS 0 in file libs/filesystem/fat_sd/ffconf.h on the esp8266.

I tried the espruino_1v90.38 build of ESPRUINO_1V3 as one of the failed travis builds and get a similar error with the original fat_sd-code:

`/usr/bin/../lib/gcc/arm-none-eabi/6.2.1/../../../../arm-none-eabi/bin/ld: region `FLASH' overflowed by 7248 bytes`

Please let me know how I can help!

wilberforce:
Where are you setting _WORD_ACCESS == 0 ?
	/* Enable word access to the FAT structure */

This could get done in a conditional test for the esp_8266 build

hmut:
You find the option _WORD_ACCESS == 0 in file libs/filesystem/fat_sd/ffconf.h, near the bottom of the file. It would be great to have a conditional test for the esp8266 build, thanks for considering!

gfwilliams:
> region 'FLASH' overflowed by 7248 bytes

I think you're probably using an old compiler, or not using RELEASE=1 - the newer compilers do a much better job size-wise.

Just tested and it works, so merging!

Is _WORD_ACCESS==0 really required for ESP8266? I'd have thought it'd be ok as it's just reading via SPI (it has to be byte by byte anyway). If it is needed then an ifdef ESP8266 would be an idea though

hmut:
I've used the exaxt same compiler version as your travis test system and applied the same exported vars, which does not include RELEASE=1. Did you try another travis build to check the size?

I don't know if the #define _WORD_ACCESS 0 is really needed, but with the #define _WORD_ACCESS 1 my esp8266 crashes immediatly on fs.readdir(). At least setting the option to 0 was useful for me.

The option _WORD_ACCESS is dropped in version 13 of the library. The default is #define _WORD_ACCESS 0 up to version 12.

gfwilliams:
Yes, the compiler used in Travis is the old one - it needs changing really. IMO I should just package everything up into one Travis build at the same time - it'd be a lot more friendly on their systems.

So in v13 there is no _WORD_ACCESS at all? Because it is 0 all the time, or they thought it wasn't needed?

gfwilliams:
Actually I just disabled it completely anyway. The comment below _WORD_ACCESS seems to imply that I shouldn't have used it even on ARM

hmut:
The investigation why the option is dropped in 13 may end in a very long night...
Isn't the standard endianess on arm little endian? The option promises "improved performance and reduced code size" and it was no problem on any official boards so far?

Thanks a lot for merging and fixing the esp8266 build!

gfwilliams:
ARM can be any endianness as far as I know - it's implementation dependent :s

No, it wasn't a problem on existing boards, but if they suggest that I shouldn't use _WORD_ACCESS then I should probably do what they say :)

PR #1024, opened 2017-01-13
adding spi and i2c to NucleoL476RG board

VVESTM:
Hello Gordon, 

Here are two commits to add SPI and I2C support to Nucleo L476 RG board.

Vincent


gfwilliams:
Great - thanks!

PR #1026, opened 2017-01-15
Puck docs: Added detail and minor fixes to magnetometer functions

rsmeral:
As discussed in http://forum.espruino.com/conversations/298349/#comment13406335


gfwilliams:
Looks great - thanks!

PR #1027, opened 2017-01-21
Make sntpSync use setTime() not to break setInterval()

opichals:
The setTime() properly sets the jsiLastIdleTime which is used
in the _jswrap_interface_setTimeoutOrInterval to trigger timers.


gfwilliams:
Cool - thanks!

PR #1032, opened 2017-01-23
set WIFI_DBG 0  if def RELEASE

MaBecker:
do not include Wifi  debug statements when using RELEASE=1


MaBecker:
Hi @gfwilliams,

this is the second part of RELEASE in ESP source covering the WIFI part, please commit - thanks

PR #1033, opened 2017-01-23
set NET_DBG 0 if def RELEASE

MaBecker:
do not include NET debug statements when using RELEASE=1


MaBecker:
Hi @gfwilliams,

Reason behind this: remove DBG statements to shrink binary file size, as there is a limit.

Not sure any more if this is the correct way connecting it to RELEASE. 

Should this be better controlled by a separate setting like RELEASE_ESP8266_NET ?   

Same for #1032,  use RELEASE_ESP8266_WIFI

What do you suggest ?

gfwilliams:
That looks fine to me really... I think with RELEASE you should expect that most debugging info is removes

PR #1039, opened 2017-01-28
ld errors with mispelt BOARD definition

OwenBrotherwood:
It is very easy to make a mistake writing in the board name when doing BOARD=1 RELEASE=1 make
Make a note of it.

---

Background:
I was getting ld errors and was busy wasting time with google: it was just a missing digit in my BOARD
Possible need for a flag, BOARDDEFINED, in connection with Makefile: in the meantime, a heads up

```
LD espruino
/usr/bin/ld: cannot find -lstdc++
collect2: error: ld returned 1 exit status
make: *** [espruino] Error 1

```

### DIFF
--- a/README_Building.md
+++ b/README_Building.md
@@ -1,8 +1,10 @@
 Building
 ========
 
-**Note:** If you're swapping between compiling for different targets, **you
+**Note:** 
+- If you're swapping between compiling for different targets, **you
 need to call `make clean`** before you compile for the new target.
+- If you have a ld error, double check the board name in the <BOARD>=1 make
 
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks - yes, it's not ideal. Not sure it's worth fixing though - but the docs are a good idea.

Worth noting that if you had all the needed libs installed I think it would actually build fine, but just for linux

OwenBrotherwood:
I work thru the docs idea with a "low noise" input :)
I have found the forum and gitter groups so I am now comfortable.

PR #1043, opened 2017-01-31
Update Dockerfile to use latest arm build tools

urish:
This fixes the build for the Micro:Bit inside Docker


gfwilliams:
Great - thanks!

PR #1045, opened 2017-01-31
Nucleo l476 rg analog

VVESTM:
Hello Gordon,

Here are some updates for Nucleo L476 RG board.
It adds adc and dac functionnalities.

Regards
Vincent

### DIFF
--- a/boards/NUCLEOL476RG.py
+++ b/boards/NUCLEOL476RG.py
@@ -42,7 +42,7 @@ chip = {
   'spi' : 4,
   'i2c' : 3,
   'adc' : 1,
-  'dac' : 0,
+  'dac' : 2,
   'saved_code' : {
     # code size 300000 = 0x493E0 starts at 0x0800 0000 ends at 0x0804 93E0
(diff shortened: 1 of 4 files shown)


gfwilliams:
Cool - thanks!

VVESTM:
Thanks Gordon
With this pull request, the SW concerning stm32 low layer api on NucleoL476RG is pretty complete.

Vincent

gfwilliams:
Wow, thanks! How portable do you think this is to different parts?

There's a post on the forum at the moment where someone is trying to port to the Nucleo F413:

http://forum.espruino.com/conversations/299344/#comment13447980

VVESTM:
Low Layer API are done to ease the portage on differents chips. It will be portable if there is no big difference on the IP used.
Unfortunately, this is not yet available on all the platforms. The first family to benefit of this API was L4. It will come for F4 during first or second quarter. So need to wait a little bit more for F413...
Once the LL is available on F4, it should be straight forward to run my software on this board.

PR #1049, opened 2017-02-04
Add NRF.setScanResponse() method

urish:
Lets the user specify/override the scan response data

### DIFF
--- a/libs/bluetooth/jswrap_bluetooth.c
+++ b/libs/bluetooth/jswrap_bluetooth.c
@@ -538,6 +538,45 @@ void jswrap_nrf_bluetooth_setAdvertising(JsVar *data, JsVar *options) {
 /*JSON{
     "type" : "staticmethod",
     "class" : "NRF",
+    "name" : "setScanResponse",
+    "generate" : "jswrap_nrf_bluetooth_setScanResponse",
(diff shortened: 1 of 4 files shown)


libs/bluetooth/jswrap_bluetooth.c
```diff
@@ -538,6 +538,45 @@ void jswrap_nrf_bluetooth_setAdvertising(JsVar *data, JsVar *options) {
...
+```
+NRF.setScanResponse([0x07,  // Length of Data
+  0x09,  // Param: Complete Local Name
+  'S', 'a', 'm', 'p', 'l', 'e']);
+```
+*/
+void jswrap_nrf_bluetooth_setScanResponse(JsVar *data) {
+  uint32_t err_code;
+  ble_advdata_t advdata;
+  jsble_setup_advdata(&advdata);
```
gfwilliams: Just wondering - do you need this? Otherwise it looks perfect - thanks!

urish: Good catch, thanks!

gfwilliams:
Thanks - looks fantastic. You don't fancy implementing anything else do you? :)

PR #1053, opened 2017-02-06
Fix HID on Windows

nailxx:
This is a fix of an old standing problem (http://forum.espruino.com/conversations/275155/) when Windows ignores Espruino’s HID role in most cases.

The reason is Espruino in such cases serves both a CDC device and a HID to OS simultaneously. However, both devices share VID/PID. And iff a CDC-driver from STM32 is installed in the system Windows refuses to consider the device as a HID. In other words, Espruino HID works on Windows until the driver will be installed.

The changes resolve the conflict. Tested on Win10, Mac, Linux

### DIFF
--- a/targetlibs/stm32usb/usbd_cdc_hid.c
+++ b/targetlibs/stm32usb/usbd_cdc_hid.c
@@ -16,7 +16,7 @@
 #define CDC_CMD_EP                    0x82  /* EP2 for CDC commands */
 
 #define HID_IN_EP                     0x81
-#define HID_INTERFACE_NUMBER          0
+#define HID_INTERFACE_NUMBER          2
 #endif
 
 extern USBD_HandleTypeDef hUsbDeviceFS;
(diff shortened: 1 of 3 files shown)


gfwilliams:
This is fantastic - thanks! I've been avoiding this because I don't understand or like messing with USB descriptors :)

I see you changed bDeviceClass - so does this change enumeration even when USB HID isn't enabled?

I just want to make sure it won't cause problems without since this code is also used for the bootloader - and if I ship out new boards with a dodgy bootloader it'll be really bad news :)

nailxx:
You’re welcome.

We’ll test the no-HID scenario tomorrow to be sure. I’ll put a note on results.

gfwilliams:
Great, thanks!

nailxx:
We’ve done a couple of test and all they have passed.

Take a fresh Espruino (Iskra JS) with no HID enabled, Windows 10 with a STM32 driver and try:

- Power up the board via external power supply, enter bootloader, plug via USB, try to connect and communicate … OK
- Plug via USB, enter bootloader, try to connect and communicate … OK

If you’d take a look at `~/.arduino15/packages/arduino/hardware/avr/1.6.17/cores/arduino/USBCore.cpp` available after installing Arduino IDE, you’ll see that ATmega32Ux Arduinos use the same class/subclass/protocol combo. Although it is used only when the Arduino is not in bootloader mode, their bootloader “Caterina” presents itself as a CDC only. Perhaps, it is to keep bootloader as small as it can be.

To me it looks like the merge is safe. However, I insist upon you’ll check the behavior by yourself because it is severe to your business. I don’t want a catastrophe.

gfwilliams:
Great - thanks for checking! I'll give it a try and make sure.

This'll be a huge help - it's always been annoying that HID didn't work on Windows.

gfwilliams:
Just tested - the bootloader seems to work great on all platforms. Merging - thanks! :)

PR #1058, opened 2017-02-09
add characteristic properties to object

Whizzard:
see [Nordic Documentation](http://infocenter.nordicsemi.com/topic/com.nordic.infocenter.s132.api.v3.0.0/structble__gatt__char__props__t.html)

Example output:
```
BluetoothRemoteGATTCharacteristic {
  "uuid": "0x4242",
  "props": { "read": true, "write": true, "write_wo_resp": true, "notify": true, "indicate": false, "broadcast": false, "auth_signed_wr": false },
  "handle_value": 19, "handle_decl": 18 }
```

### DIFF
--- a/targets/nrf5x/bluetooth.c
+++ b/targets/nrf5x/bluetooth.c
@@ -527,6 +527,17 @@ static void on_ble_evt(ble_evt_t * p_ble_evt)
             JsVar *o = jspNewObject(0, "BluetoothRemoteGATTCharacteristic");
             if (o) {
               jsvObjectSetChildAndUnLock(o,"uuid", bleUUIDToStr(p_chr->uuid));
+              JsVar *p = jspNewObject(0,0);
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks - I'm afraid I did this at about the same time though :(

https://github.com/espruino/Espruino/commit/766a577e3f5da127ce4815544a31cb7562dfd96c

gfwilliams:
Sorry to waste your work - I can't believe we did basically the same thing within half an hour of each other

Whizzard:
No problem. Sounds funny that you did it just the same time.
I was digging into the notification things and found this by accident and it did not look too complicated.

Anyway, I found my solution a bit nicer - with the flags being a nested object. ;-)

Whizzard:
Ahem. Yes. :-)
Not my day.

PR #1066, opened 2017-02-13
add erase_flash

MaBecker:


gfwilliams:
Thanks - good idea :) Of course if there was an all-in-one 1M image file, you'd have a much simpler flash command, *and* saved code would get cleared out?

PR #1067, opened 2017-02-13
add convert library

lancernet:
add functions:
floatToBytes
doubleToBytes
bytesToFloat
bytesToDouble


gfwilliams:
There's no way I can merge this onto master because it's based off the ESP8266 fork.

I also don't understand why we need this, when Typed arrays do it all already? For instance:

```
new Float32Array([1234]).buffer
```

If you wanted something that seemed a bit easier, maybe you could just make a JS module that wrapped around Typed Arrays?

PR #1074, opened 2017-02-16
add ESP32 image and pins

MaBecker:
use scripts/build_board_docs.py ESP32
to generate boards/ESP32.html


MaBecker:
please wait - names are in wrong order

MaBecker:
now way - have to try again

PR #1095, opened 2017-03-03
Add provision-framework.sh - Update travis

wilberforce:


gfwilliams:
Hi - please can you avoid merging stuff like this into the master branch without asking me to review? That's why we had the ESP32 branch.

I mean, Travis is showing everything was failing to build - and yet you merge it in anyway???

wilberforce:
Sorry Gordon, yes the last commit is correct:
https://github.com/espruino/Espruino/pull/1100

I had closed this #1095, it was an update in the ESP32 branch. I did all of these testing in the ESP32 branch.

I wanted to ensure that it was ok in the master branch, as the build system depends on it and I would have reversed the change if the build had not worked. The only way to test the Travis system is do a commit into it.

I'll keep you in the loop in the future. Sorry.

The good side is now you can clone the Espruino and go:

`source ./Espruino/scripts/provision-framework.sh PICO_1V3`

or

`source ./Espruino/scripts/provision-framework.sh ESP32`

or 

`source ./Espruino/scripts/provision-framework.sh ESP8266`

or 


If the required build tools are not in place, it will install them and the libraries, simplifying the build process.

It would be great if we could ask others to test this - would a forum post be the best bet?

gfwilliams:
Thanks! Yes, I think a forum post would be good.

@OwenBrotherwood recently wrote some docs on building, so if he's got time it'd be great if he could take a look. This should massively simplify the build!

OwenBrotherwood:
I try the work ("It would be great if we could ask others to test this") and get back on this.

Ping if there is some specific areas to be addressed, otherwise I take a general "Look from Outside" (dummies look) at the general area of simplifying the build.

wilberforce:
@OwenBrotherwood Thanks for looking at this.  A couple of notes.

a. The apt dependencies still need to be done. The script could be extended to do this - but it would need to be an option
b. At  this stage I was not expecting back-to-back installs. The last board env var would need to be cleared as what is occuring above it ESP32=1 is still set AND PUCKJS is set
Again the script could be modified.
For now it might be easier to open a new session - then the ENV would be all cleared down for each board.

I appreciate you trialing this!

wilberforce:
@OwenBrotherwood 

I've updated, so you can now switch boards.

https://github.com/espruino/Espruino/pull/1104

`source ./Espruino/scripts/provision-framework.sh PUCKJS`
This is still an issue as the prerequisites need to be installed - I'm not familiar with what is required.

wilberforce:
https://github.com/espruino/Espruino/blob/master/README_Building.md

```
sudo apt-get update
sudo apt-get install -y \
  lib32z1 lib32ncurses5
wget https://launchpad.net/gcc-arm-embedded/5.0/5-2016-q3-update/+download/gcc-arm-none-eabi-5_4-2016q3-20160926-linux.tar.bz2
tar xjf gcc-arm-none-eabi-5_4-2016q3-20160926-linux.tar.bz2
sudo mv gcc-arm-none-eabi-5_4-2016q3-20160926 /usr/local
export PATH=/usr/local/gcc-arm-none-eabi-5_4-2016q3/bin:$PATH
cd ~/source/repos/github/espruino/Espruino
make clean && DFU_UPDATE_BUILD=1 PUCKJS=1 RELEASE=1 make
ls -l *puckjs*
```

So is a different compiler version needed?

OwenBrotherwood:
( off topic: no reply needed: emotions ok: travis builds have been in my thoughts from the first time I read somewhere about the used compiler creating bigger builds: however, I have kept myself on doc and learning curve)

gfwilliams:
Puck.js shouldn't need a different compiler - but especially when @OwenBrotherwood was trying to get things working, using the newest available compiler could have caused some issues (major GCC version changes have tended to break the build in the past).

PUCKJS should be in the Makefile though? It is on my version of the repo :)

There's no build in travis because it also needed `nrfutils` installing to build the zip file for the firmware updates. That's definitely something that could be added pretty easily with the current arrangement though.

OwenBrotherwood:
( FYI No issues with newest compiler: both versions provide "good" zip for puck.js: however I am using the version in README for conformity: I think we are talking about two different things with the comment "So is a different compiler version needed?" as a possible question from concerning esp contra something else This could create too much noise if I have to keep commenting to ensure that the work I did is understood: l stop commenting here as it may defeat the PR purpose)

wilberforce:
The purpose is two fold.

A. simply the Travis builds by only pulling in the framework and compiler required.
B. Make it easier for people to build the firmware from a clean install

So 
`source ./Espruino/scripts/provision-framework.sh ESP32`

The idea is that the board name gets passed to the script.

With PUCKJS I think you picked the most complicated exception!

Anyway, the multiple boards setting have been fixed in the esp32 branch.

I still can't get the PUCKJS case to work, it getting compile errors not related to libraries.

All the other boards that are in the Travis config work.

wilberforce:
> There's no build in travis because it also needed nrfutils installing to build the zip file for the firmware updates. That's definitely something that could be added pretty easily with the current arrangement though.

Yes a case could be added that does the pip install if it's not installed for nrfutils. Is there a family of boards this applies too?

PR #1097, opened 2017-03-03
add clock stretch

MaBecker:
many thanks to @frida for sharing this code

### DIFF
--- a/targets/esp8266/i2c_master.c
+++ b/targets/esp8266/i2c_master.c
@@ -8,6 +8,7 @@
  * Modification history:
  *     2014/3/12, v1.0 create this file.
  *     2015/10/11, modified to work with any pin for Espruino
+ *     2016/05/23, modified to work with clock streching and bitrate
  *
  * ESPRSSIF MIT License
  *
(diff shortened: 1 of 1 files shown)


targets/esp8266/i2c_master.c
```diff
@@ -45,17 +46,19 @@
 } while(0)
 
 // For ~100khz at 80Mhz or 160Mhz:
-//#define i2c_master_wait    os_delay_us
+#define i2c_master_wait(x)    os_delay_us(x*y)
```
gfwilliams: merging anyway because it's ESP8266 - but this is pretty nasty :)  (using a local variable y that's not passed in as an argument)

Also you might as well just delete the commented out code - it's Git so they're still in history if anyone cares

MaBecker:
>but this is pretty nasty :) (using a local variable y that's not passed in as an argument)

have to check ....

>Also you might as well just delete the commented out code - it's Git so they're still in history if anyone cares

ouch - thanks for pointing this out!  added  #1098

PR #1104, opened 2017-03-07
check for existing tool folders, store last $BUILD

wilberforce:


gfwilliams:
Neat - thanks! Do we actually need to set a global `BOARD` variable in the first place? It seems like that may be what was tripping up the Makefile?

gfwilliams:
Also, do you have any objections to me moving http://s3.voneicken.com/xtensa-lx106-elf-20160330.tgx to our EspruinoBuildTools repo now? I might also move the GCC arm compiler, since it'll be way faster to decompress a tar.gz than to install from the package manager

wilberforce:
> Do we actually need to set a global BOARD

Yes this is how Travis sets which framework to install. I have set it up so that only the exceptions - Linux, esp32 and esp8266 are handled, al else gets the arm framework. Otherwise every combination of board would need to be added to `provison-framework.sh` and it would be a maintenance nightmare.
The issue with the make file was due to both ESP32=1 and PUCKJS=1 both being set. The patch above fixes that.

wilberforce:
> Also, do you have any objections to me moving http://s3.voneicken.com/xtensa-lx106-elf-20160330.tgx to our EspruinoBuildTools repo now? I might also move the GCC arm compiler, 

No objections. We should probably tell tve although he hasn't been around these parts lately.

Ideally the framework should also be there, so it can be built from scripts.. e.g. The lwip falcon stuff and the sdk 2.0 so that the repositiory also builds the framework libs.

wilberforce:
This was in the esp32 branch - so if you are going to move esp32 and arm you might want to push to master first

gfwilliams:
Does travis not set `BOARD` up itself now, so there's no need for the provision script to do it? https://github.com/espruino/Espruino/blob/master/.travis.yml#L40

But yeah, I'll merge this in to master.

@tve are you ok with us moving xtensa-lx106-elf-20160330.tgx to GitHub? It might save some bandwidth on your server :)

wilberforce:
> Does travis not set BOARD up itself now, so there's no need for the provision script to do it?

It is set here for when an end user calls the provision script - in this case the board is not set.

It is retained in the env of the calling shell, so if the user then builds another board, it unsets the last set so we don't get the Makefile issues @OwenBrotherwood discovered.

gfwilliams:
>  it unsets the last set so we don't get the Makefile issues @OwenBrotherwood discovered.

But if it never set `BOARD` in the first place presumably you wouldn't have the issues either?

Sorry - It seems I'm not really understanding what's going on here.

Is your intention that instead of this:

```
source ./Espruino/scripts/provision-framework.sh PUCKJS
make clean && DFU_UPDATE_BUILD=1 PUCKJS=1 RELEASE=1 make
```

you do this:

```
source ./Espruino/scripts/provision-framework.sh PUCKJS
make clean && DFU_UPDATE_BUILD=1 RELEASE=1 make
```

eg, without specifying the board on the make command-line?

I'm not sure why else you'd need to specify a BOARD environment variable...

wilberforce:
If you did this

`source ./Espruino/scripts/provision-framework.sh PUCKJS`

And then 

`source ./Espruino/scripts/provision-framework.sh ESP32`

Both PUCKJS=1 and ESP32=1 are set. This breaks the Makefile.

The script would need to know all of the possible boardnames, and if one was set, all of the others would need to be unset. This means maintaining a list within the script, which could get out of step with the Makefile if people didn't know it needed to be updated.

gfwilliams:
I still don't understand.

If `./Espruino/scripts/provision-framework.sh` does not set `BOARD=` or `PUCKJS=1/etc` then you would have no need to unset it, and everything would surely 'just work'?

None of the build steps set `BOARD=` anything before, so I'm not sure why you have to do it now?

The usual way of building is to write `BOARDNAME=1 make` - obviously if you export BOARDNAME and then change it then you have a problem - but I don't see why you'd do that unless you're absolutely sure you're only working with one board.

wilberforce:
Ok. I understand what you mean.

I was following this pattern:

https://github.com/espruino/Espruino/blob/master/README_Building.md

In the part for the esp8266 where it suggests setting up a batch file that sets your env vars.
I was trying to maintain that.
The idea was that the script set up all of the vars you needed to build - the $PATH for gcc, the ESP_{framework} and also set the board.

Then you can just type:
`make clean && make`

If the script does not set it, the. Will be required as part of the `make`

gfwilliams:
Ahh, thanks - that explains it :) That was something only the ESP8266 docs suggest.

Hopefully if I do https://github.com/espruino/Espruino/issues/1105 then the problem will go away, as there will only be one variable needed, `BOARD` - so as you say you'll never be able to have both set

tve:
> Also, do you have any objections to me moving http://s3.voneicken.com/xtensa-lx106-elf-20160330.tgx to our EspruinoBuildTools repo now?

I don't mind either way. "My server" is AWS S3, so it's fast, cheap, and reliable. If you're proposing to move it into a git repo, then IMHO that's a bad idea, but it's your choice.

Issue #1047, opened 2017-02-03
nRF52: Get RSSI of connection when in Central mode


gfwilliams:
http://forum.espruino.com/conversations/299679/#comment13453621

billsalt:
Certainly a feature request, but it really is needed information to help authenticate that you're dealing with the correct device and/or let the user know when it is wandering out of range.  

Having the RSSI only while scanning makes it difficult, because there can me many, many devices (at least in my office there are >20).  Hoping this can sneak into the next release (or a beta?). 

Happy to test!

Whizzard:
Have just added this in my repo.
First I wanted to introduce the API to the GATT instance, but it also worked with the existing static NRF object, so I left it for now.
The changes are minimal. 
What do you think?

billsalt:
Great!  I'm a bit of a NOOB; can you give me a quick example of usage?  And how I can test it out?

Whizzard:
The branch is not reviewed, merged or released yet, so you have to build it first and flash it to your device.

Then when you are connected to another peripheral, just do:
```
NRF.setRSSIHandler(function(rssi) { console.log(rssi); });
```
like when you are the peripheral yourself.

Documentation: http://www.espruino.com/Reference#l_NRF_setRSSIHandler

billsalt:
That’s just the way I thought it should work, so great!  I’ll try to carve out some time and see if I can work my way through a build in the next few days.

Thank you for doing this!!

Bill

--
Bill Saltzstein
Code Blue Communications, Inc.

billsalt@consultcodeblue.com
www.consultcodeblue.com
http://www.linkedin.com/in/billsaltzstein (LinkedIn)

+1 425-442-5854
Skype: billsaltz

Whizzard:
If it's for the puck.js I just compiled it (do not have the devices at hand, so you may test on your own risk):
[espruino_1v91.2685_puckjs.zip](https://github.com/espruino/Espruino/files/828509/espruino_1v91.2685_puckjs.zip)

How to flash the ZIP to your puck is documented here:
http://www.espruino.com/Puck.js -> Firmware Updates

billsalt:
Thanks!  I certainly can load a puck, but I also use the Nordic nRF52832 demo board.  I’ve got both, but you just made it very easy to use the puck.  Advantage of the kit is I can use the serial console log without having to wire anything up.  Really helpful since what I’m trying to do is simultaneous master/slave work.

But I can certainly to the first pass testing with the puck.

[...]

PR #1107, opened 2017-03-08
[WIP] NRF52: track RSSI changes when being central and connected to peripheral

Whizzard:
This is a quite simple solution that does not affect the API.
Scanning for peripheral + central is handled via the same callback, just the internal handle is different.

Solves #1047


Whizzard:
OK - not perfect from API perspective:
- you can not distinguish peripheral and central side if you have both
- you can not distinguish multiple connections as central (if possible)

So we'd better have a separate callback for the central side...

gfwilliams:
Fixed with https://github.com/espruino/Espruino/pull/1148/files

Thanks for your help with the PR, but as discussed I reckon this works better (having the potential for multiple callbacks)

PR #1111, opened 2017-03-11
change toggle function

lancernet:
change toggle function to return after toggle pin state

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ gcc-arm-none-eabi*
 /functions.html
 targetlibs/nrf5x/examples
 /.vscode
+/libs/lanruino/*
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 *.srec
 *.bin
 *.pyc
+*.make
 /*.zip
 gen/*
 .vagrant/*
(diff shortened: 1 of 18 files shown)


gfwilliams:
I'm afraid I can't merge this... This change should require changing maybe 3 lines, but this PR changes 18 files!

Also your solution doesn't work I'm afraid - since it reads the input rather than the state of the output (which was stored in the 'on' variable anyway).

I've just committed something that will add this functionality thugh

lancernet:
Sorry, i just learned to use github. I am sorry to bring you inconvenience

gfwilliams:
No problem :) It is a good idea to return the value.

When you add your code to GitHub you can often click to  'compare' it and see what has changed - you can then tell if you've changed what you expected to change as it is easy to commit things by accident :)

lancernet:
thanks you gfwilliams

wilberforce:
@gfwilliams 

https://github.com/espruino/Espruino/blob/267fc7bc61df9db1956dc361bd03da8dce408c28/src/jswrap_pin.c#L185

If the pin in invalid nothing is returned and the compiler complains. Can we return a javascript undefined with a return type of bool?

gfwilliams:
Just fixed. No, we can't - I've switched it to return false.

It's super unlikely anyway, since you'd have to do something like `Pin.toggle.call(undefined)` - in which case you get what you deserve really :)

PR #1112, opened 2017-03-11
remove ESP8266: now build in graphics lib

MaBecker:


gfwilliams:
Thanks! missed that :)

PR #1120, opened 2017-03-25
nRF5x: not all services returned in `NRF.findDevices`

urish:
If some device advertised multiple 16-bit service UUIDs in a single advertisement section, only the first one was returned.


gfwilliams:
Awesome, thanks!

PR #1121, opened 2017-03-25
nRF5x: support integer UUIDs in requestDevice

urish:
This aligns the behavior with [Web Bluetooth](https://webbluetoothcg.github.io/web-bluetooth/#standardized-uuids), where you can specify service UUIDs either as strings or integers. 

In addition, it also allows you to use the full 128-bit UUID to match a 16-bit UUID (e.g. `"0000FFE5-0000-1000-8000-00805F9B34FB"` will match service `FFE5`).


gfwilliams:
Very nice - thanks again!

PR #1123, opened 2017-03-29
Fix typo in rotating setAdvertising example

joebowbeer:
Remove extraneous byte (`0x03`) from rotating `setAdvertising` documentation.


gfwilliams:
Are you sure about this? Does it not work the other way? Because that's exactly what the eddystone module does, which works fine:

https://github.com/espruino/EspruinoDocs/blob/master/modules/ble_eddystone.js#L12

joebowbeer:
Aha. My bad. The 0x03 is the length field... I missed it in the preceding example. Closing pull request.

BTW, Why is this array example missing from the online docs?

https://www.espruino.com/Reference#l_NRF_setAdvertising

PR #1128, opened 2017-04-03
nRF5x: add "connectable" option to setAdvertising()

urish:
This allows the device to be non connectable, useful for beacons


gfwilliams:
Thanks! Any chance of adding bleConnectable as a flag inside bleStatus though? It's what I've tended to do for other stuff.

It's not  big deal on nRF52, but on nRF51 when you only have less than 4k of RAM of JS code, I figure anything we can do to reduce usage is worth it :)

urish:
Done!

gfwilliams:
Thanks!

PR #1132, opened 2017-04-10
change jsError() to jsExceptionHere(). Suppress jsiConsolePrintf() for HTTPS connections

nkrkv:
Currently, there are places in `network.c` which raise errors that we have no chance to handle with JavaScript. Also, there are places when the console is polluted with a forced debug/trace output to it. The PR fixes that.

### Rationale

We’ve tried to make a device that could work with HTTPS API for a long time without restarts. In our case, it’s Telegram Messenger API. Works fine, but fails forever at some point when a problem that otherwise could be solved by retrying lead to a unhandled error.

With this patch applied we can handle such cases by using trivial `try { ... } catch` blocks.

### DIFF
--- a/libs/network/network.c
+++ b/libs/network/network.c
@@ -246,7 +246,7 @@ bool networkGetFromVar(JsNetwork *net) {
 #endif
   case JSNETWORKTYPE_JS : netSetCallbacks_js(net); break;
   default:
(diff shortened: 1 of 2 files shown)


gfwilliams:
Thanks!

PR #1144, opened 2017-04-19
Also build espruino_esp32.bin for tgz

wilberforce:


gfwilliams:
Are these not the changes I just made?

I think you might be doing something odd with the way you're merging and creating PRs.

Ideally you want to do:

```
git checkout ESP32
git merge master
# ESP32 will now have all the code from master in it, as well as your changes
... Now you can issue a PR from it
```

You might just be manually copying files or something?

wilberforce:
I had added a couple of lines to Make file, so it was your changes plus a couple of lines the copied to the binary so that tgz would have a consistent filename.

I have reverted this, and merged with master with the current build so that ESP32 will be up to date.

gfwilliams:
Ok, thanks!

PR #1147, opened 2017-04-21
suppress warnings, FatFS docs, suppress logging

wilberforce:
Remove esp-idf logging - stops wifi and booting output confusing web ide
clean up FatFS - fix compile warnings
remove unimplemented readonly mode (it was a compile not runtime option)


targets/esp32/jshardware.c
```diff
@@ -354,8 +341,9 @@ void jshSetOutputValue(JshPinFunction func, int value) {
   }
   else{
     pin = ((func >> JSH_SHIFT_INFO) << 4) + ((func >> JSH_SHIFT_TYPE) & 15);
-    value >> (16 - PWMTimerBit);
-    setPWM(pin,value);
+    // Was warning: statement with no effect [-Wunused-value] - changed to value - needs testing!
```
wilberforce: @jumjum123 

I have commented this out as it has no effect and was a compiler warning - the results was thrown away so I'm not sure this code is working as intended?

Can you recheck  it working as intended?

targets/esp32/main.c
```diff
@@ -51,7 +53,8 @@ static void espruinoTask(void *data) {
   jshInit();     // Initialize the hardware
   jswrap_ESP32_wifi_restore();
   jsvInit();     // Initialize the variables
-  vTaskDelay(1000 / portTICK_PERIOD_MS);
+  // not sure why this delay is needed?
+  vTaskDelay(200 / portTICK_PERIOD_MS);
```
wilberforce: @jumjum123 
I'm not sure the intention of the the delays? The one at line 38 does not need to be needed.

I could not remove the one at line 57 - it seems some delay was still required - so made it 200 ms

targets/esp32/jshardware.c
jumjum123: The value for jshSetOutputValue is given as 16 bit.
DAC is set to handle 10 bit.
Therefore this line should convert the 16 bit value to a 10 bit value.
Oh, should the line be: value = value >> (16 - PWMTimerBit); ?

targets/esp32/main.c
jumjum123: As far as I remember this delay already existed in first version of main.c, written by nkolban

targets/esp32/jshardware.c
wilberforce: Ok. I can make the change and you test it - Is that the best way.

targets/esp32/main.c
wilberforce: Yes - I think you are correct. I'll remove the comments then and leave only the small delay.

targets/esp32/jshardware.c
jumjum123: Thats ok for me

wilberforce:
@gfwilliams Please can you merge into master. Thanks!

gfwilliams:
Great, thanks!

PR #1170, opened 2017-05-25
nRF52: Addendum to 2deb5f97

drandreas:
The patch I posted earlier was part of a series.

For this reason the second safety check is off by
several lines. The safety check should be performed
prior to any other branch.

Sorry for the inconvenience


gfwilliams:
Ahh - thanks!

drandreas:
I think you also added `else rx_data_size = 0;` below the first safety check. This filters out single byte responses without CRC.  I'm not sure if this every happens.
Just something to keep in mind.

gfwilliams:
Thanks - yeah, maybe it was ill-advised - I just figured if it didn't have a CRC it was probably malformed and we wanted to fail.

PR #1180, opened 2017-06-09
L4 LL fix

VVESTM:
Hello Gordon, 

Here are two fix for stm32_ll part. It will fix some issues I have seen on ADC and I2C.

Regards
Vincent


gfwilliams:
Nice, thanks!

PR #1184, opened 2017-06-16
Fix building on newer Linux systems with Python3 default

daveshah1:
Newer systems such as Arch have `python` defaulting to Python 3 if it is
installed. This commit fixes building in this case, by detecting it and
using the `python2.7` command instead. It also allows the Python command
used to be overridden with the PYTHON variable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,14 +54,6 @@ ifndef SINGLETHREAD
 MAKEFLAGS=-j5 # multicore
 endif
 
-ifndef PYTHON
-ifneq "$(findstring Python 3, $(shell python --version 2>&1))" ""
-PYTHON=python2.7
-else
-PYTHON=python
-endif
-endif
-
 INCLUDE=-I$(ROOT) -I$(ROOT)/targets -I$(ROOT)/src -I$(GENDIR)
 LIBS=
 DEFINES=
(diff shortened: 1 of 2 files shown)


gfwilliams:
Thanks - does it really not work with Python 3? We had a bit of a push to tweak things a while back and I thought the scripts would run on either version

daveshah1:
Odd - I was getting this error with Python 3.6.1 on Arch Linux, but not 2.7.13

`
Generating platform configs
Exception ignored in: <common.Discarder object at 0x7f3bb71285c0>
AttributeError: 'Discarder' object has no attribute 'flush'
make: *** [Makefile:724: /home/david/Espruino/gen/platform_config.h] Error 120
`

gfwilliams:
Can you run `python3 scripts/build_platform_config.py $(BOARD) gen/platform_config.h` and see what it actually reports in full? (or make with `V=1`)

I don't see any instance of `flush` in the python file - so maybe it's somehow using a python2.7 library in python3? 

I just replaced all instances of `python` with  `python3` in my Makefile here and it seems to work for me - albeit only with Python 3.5.2, but I doubt they would have broken compatibility between 3.5 and 3.6?

If they have I'd really like to fix the scripts though, rather than doing the usual thing of regressing to 2.7 ;)

daveshah1:
Investigating further it seems the `Discarder` class used in `common.py` to suppress `stdout` output when `SILENT` is set (which means the error doesn't occur with `V=1`) needs a `flush` dummy method for the code to run properly on Python 3.6. I'll fix it properly that way and create another pull request.

gfwilliams:
That's awesome - thanks!

gfwilliams:
Great, thanks for this!

PR #1189, opened 2017-06-23
No timeout argument setTimeout() calls work

opichals:
It behaves as if zero (0) was passed in. Compliant to the spec: see
https://www.w3.org/TR/2011/WD-html5-20110525/timers.html#get-the-timeout


gfwilliams:
Thanks!

Issue #1062, opened 2017-02-11
UDP Support


opichals:
# UDP Support

The purpose of this ticket is to track Espruino native UDP support.
The intention is to loosely follow the node.js `dgram` (https://nodejs.org/api/dgram.html) module API .

The development is currently in progress and can be reviewed at https://github.com/espruino/Espruino/compare/master...opichals:espruino-udp

## Phase 1

Minimal effort to get UDP working.

- [x] Introduce ST_UDP flag
- [x] Drop NetCreateFlags in favor of SocketType
- [x] Add SocketType argument to net->createsocket()
- [x] Implement ST_UDP handling (network_linux.c)
- [x] Validate feasibility for ESP8266 (network_esp8266.c)
- [x] Validate this being enough for JS mDNS implementation
- [x] Multicast support
- [x] Made 'on:message' the same as 'on:data' (node.js compat)

```js
// server
let srv = dgram.createSocket('udp4');
srv.bind(41234, bsrv => {
  bsrv.on('message', function(msg, info) {
    // echo 1
    srv.send('echo1: '+msg, info.port, info.host);
    // echo 2 // TBD: using setTimeout() to separate the message packets
    setTimeout(() => srv.send('echo2: '+msg, info.port, info.host));
  });
});

// client
let sckt = dgram.createSocket('udp4');
sckt.on('message', function(msg, info) {
  console.log('c:message', msg, info.host, info.port);
});
sckt.send('Espruino UDP Message!', 41234, '127.0.0.1');
```

**Issues**

Messages greater than net->chunkSize seem to sometimes freeze the whole esp8266 board. This could be because of heap shortage (wasn't able to verify)
- [x] Drop messages bigger than net->chunkSize (536/2 - quite small, many osx Bonjour/mDNS messages are way bigger).

**Cleanup**
- [x] Working test_dgram_socket.js test
- [ ] Fix cross-layer interfacing and avoid hacks
- [ ] Add property name constants ("receiveData", "multicastGroup", ...) similarly to e.g. HTTP_NAME_RECEIVE_DATA
- [ ] Add doc comments to the added function parameters in network.h/c files

## Phase 2

Separating send message packets (separate pull request).

- [ ] Combine the socketserver:receiveInfo/receiveData, create a message data object {data, host, port} and pass to clientRequestWrite
- [ ] Adapt to make it send every send() call separately, not concatenating to a stream (pushData)

```js
// server
let srv = dgram.createSocket('udp4');
[...]

PR #1190, opened 2017-06-23
net->createsocket() SocketType argument added

opichals:
First steps for UDP support (https://github.com/espruino/Espruino/issues/1062)


gfwilliams:
Looks awesome, thanks!

PR #1195, opened 2017-06-30
Align getting sendData with its jsvUnLock

opichals:
Both the sendData jsvObjectGetChild() and jsvUnLock() are in the
same `if (!connectionCloseNow)` code block.


gfwilliams:
Wow, great - thanks for spotting that! I'm surprised that didn't cause more issues to be honest.

opichals:
It was actually the added [UDP test](https://github.com/opichals/Espruino/blob/espruino-udp/tests/test_dgram_socket.js) that was leaking and made me debug this.

Issue #1037, opened 2017-01-26
load() causes endless loops if "No code in flash!"


MaBecker:
@gfwilliams - please set label as ESP8266

    >load();
    =undefined
    No code in flash!
    ERROR: Not connected to the internet
    ERROR: Not connected to the internet
    ......

gfwilliams:
Wow, odd. Does it come back after a power cycle? `load()` shouldn't write anything to flash so it shouldn't be able to do anything permanent

MaBecker:
A hardware reset  or power off/on brings it back

MaBecker:
I have no idea where to start digging to fix this, do you have any hints ?

gfwilliams:
Not really - but maybe add some prints to `jswrap_flash.c` to see if you can tell where it ends up looping?

MaBecker:
Hi @tve,

is ist possible to force a stack dump in function [networkGetFromVarIfOnline](https://github.com/espruino/Espruino/blob/master/libs/network/network.c#L253) to analyse the call stack ?

I tried a div by 0 : causes a reset without stack dump

```
>load();
=undefined
No code in flash!
ERROR: Not connected to the internet
 ets Jan  8 2013,rst cause:2, boot mode:(3,6)
```

wilberforce:
https://github.com/espruino/Espruino/blob/9ec75f55ce4391550f8895128be9f22d77c6f9c1/src/jswrap_flash.c#L502

gfwilliams:
But does that explain why it keeps happening over and over?

Surely if it crashed it'd just reboot as if power was re-applied?

Perhaps it's an uninitialised variable somewhere. If the ESP8266 does a 'warm' reboot then I guess RAM contents will no longer be 0, and that might throw something off.

It still doesn't really explain why 'save()' would work, and a 'load()' doesn't. They should really be identical, since 'save()' writes to flash but doesn't change the RAM contents, and `load()` isn't changing RAM contents at this point either

MaBecker:
This version still works:

```
>load()
=undefined
No code in flash!
process.env.VERSION
="1v90.2079"
process.env.GIT_COMMIT
="f39dd3b75668f059487d6cbe1ba8fa6463885668"
```

commit f39dd3b75668f059487d6cbe1ba8fa6463885668
Author: Gordon Williams <gw@pur3.co.uk>
Date:   Mon Dec 19 18:52:27 2016 +0000

[...]

PR #1197, opened 2017-07-13
some changes

iotong:
1.fixed telnet recv function bug
2.addition shell function in linux system ,that allow excute shell cmd in espruino
3.addtion xtea crypto

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,4 @@ targetlibs/nrf5x/examples
 /CURRENT_BOARD.make
 /topreadonly
 /topstrings
+/libs/network/mqtt
--- a/ChangeLog
+++ b/ChangeLog
@@ -43,6 +43,8 @@
             Add `E.set/getFlags` to allow interpreter state to be controlled from one place
             Add `E.setFlags({pretokenise:1})` to dynamically turn on pretokenisation (fix #1178)
 
+
+
      1v92 : nRF5x: Fix issue where Espruino could crash during save() if the flash got into a strange state
             Added Pin.toggle() function
             Fix implicit casting to bool/number on ArrayBuffers (fix #1030)
--- a/Makefile
+++ b/Makefile
@@ -550,6 +550,12 @@ libs/crypto/mbedtls/library/sha1.c \
 libs/crypto/mbedtls/library/sha256.c \
 libs/crypto/mbedtls/library/sha512.c
 
+ifdef USE_XTEA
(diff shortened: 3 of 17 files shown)


gfwilliams:
Thanks - why do you think `netRecv`is in error? As far as I can tell the data isn't zero-terminated so there's no problem using all 256 chars of the buffer.

Also you seem to have committed an extra bash script, and not added the `jswrap_mqtt` files? Out of interest, what did you find wrong with the existing JS MQTT implementation that you wanted a C based one for?

iotong:
When I send a larger amount of data, telnet will lose a character, I did not carefully look for reasons. But when I changed 256 to 255, the problem was not happening, and I really wanted to use the mqtt protocol to rewrite it because I used the esp8266 module when using js mqtt when i could write the amount of code It will be much less, there is no transplant is completed, so there is no upload mqtt related code

gfwilliams:
Hi - thanks for the changes, but you still haven't committed the missing files. For some reason `/libs/network/mqtt` is now in gitignore so you're not committing your jswrap files. If you check the 'travis' note next to your commit you can see that the changes you've sent don't compile: https://travis-ci.org/espruino/Espruino/builds/254742334

MaBecker:
HI @iotong, 

does this fix issue #1037?

gfwilliams:
Just to add, this is still broken because you haven't committed the files that are needed: https://travis-ci.org/espruino/Espruino/jobs/255994548

MaBecker:
tried ca67ee6, works on my side.

```
>E.setBootCode()
Erasing Flash..................
Writing...
Compressed 25600 bytes to 8
Checking...
Done!
=undefined
>load();
=undefined
Loading 8 bytes from flash...
> 
```

@wilberforce  @jumjum123 this might work for ESP32 as well 

@gfwilliams added a fix for  #1037

iotong:
@MaBecker yes ,it work!are you chinese?

gfwilliams:
Please see: https://travis-ci.org/espruino/Espruino/jobs/256128299

Please can you commit `libs/network/mqtt/jswrap_mqtt.c` and `libs/network/mqtt/jswrap_mqtt.h`? Without this, I will have to close the PR as it doesn't build!

iotong:
The transplant has not been completed, and I will restart the pull request after the completion of the transplant

MaBecker:
@iotong I am german

gfwilliams:
Great, thanks!

PR #1198, opened 2017-07-15
UDP support

opichals:
This PR is intended for code review and further guidance purposes. I have reached a working state of the https://github.com/espruino/Espruino/issues/1062 work at which point I would appreciate a deeper review of my changes and further guidance.

Brief changes summary:
* new ST_UDP SocketType flag
* netSend and netRecv receive SocketType argument just as netCreateSocket
* Board-specific UDP support is optional (only the linux & esp8266 are supported)

What to look for:
* external usage can be seen in the tests/test_dgram_socket.js
* added FIXME in the diffs where I deem the solution not nice (though working)
* The socketserver.c:receivedData is treated separately for UDP providing the remoteinfo address/port/size as the second argument to the on('message') callback
* the sendData handling as noted in https://github.com/espruino/Espruino/issues/1062 Phase 2 section should be reworked somehow to handle different host/port send is still TBD


gfwilliams:
Great, thanks! I'm on holiday this week but I'll try and give it a proper look through next week

opichals:
@gfwilliams Sure, enjoy your spare time! Till then I might get some time to work on some of the points I mentioned.

gfwilliams:
Sorry for the delay - I'm just looking through this a bit now - I'll try and mention some stuff inline...

libs/network/network.h
```diff
@@ -72,17 +72,17 @@ typedef struct JsNetwork {
...
+  int (*createsocket)(struct JsNetwork *net, uint32_t host, unsigned short port, SocketType socketType, JsVar *options);
   /// destroys the given socket
   void (*closesocket)(struct JsNetwork *net, int sckt);
   /// If the given server socket can accept a connection, return it (or return < 0)
   int (*accept)(struct JsNetwork *net, int sckt);
   /// Get an IP address from a name
   void (*gethostbyname)(struct JsNetwork *net, char * hostName, uint32_t* out_ip_addr);
   /// Receive data if possible. returns nBytes on success, 0 on no data, or -1 on failure
-  int (*recv)(struct JsNetwork *net, int sckt, void *buf, size_t len);
+  int (*recv)(struct JsNetwork *net, int sckt, void *buf, size_t len, uint32_t *host, unsigned short *port);
```
gfwilliams: Passing host/port to everything feels a bit ugly... I guess it works, but somewhere else you mentioned the need to be able to receive multiple packets?

What if the data returned/sent for UDP was just an array of bytes: 

`[PKTLEN_HI, PKTLEN_LO, PORT_HI, PORT_LO, HOST4, HOST3, HOST2, HOST1, packet data ... ]`

Just a thought - but it'd really reduce the amount of changes needed throughout the code?

libs/network/esp8266/network_esp8266.c
```diff
@@ -1007,14 +1052,42 @@ int net_ESP8266_BOARD_createSocket(
...
-  pEspconn->type      = ESPCONN_TCP;
   pEspconn->state     = ESPCONN_NONE;
   pEspconn->proto.tcp = tcp;
   tcp->remote_port    = port;
   tcp->local_port     = espconn_port(); // using 0 doesn't work
   pEspconn->reverse   = pSocketData;
-  espconn_set_opt(pEspconn, ESPCONN_NODELAY); // disable nagle, don't need the extra delay
+
+  // multicast support
+  // FIXME: perhaps extend the JsNetwork with addmembership/removemembership instead of using options
```
gfwilliams: Is this something you think would be available on all platforms? I'm struggling to see how I'd implement this for ESP8266... I think maybe we could leave it as-is for now?

libs/network/socketserver.c
```diff
@@ -317,47 +339,61 @@ bool socketServerConnectionsIdle(JsNetwork *net) {
...
-                    ));
+            if ((socketType&ST_TYPE_MASK)==ST_UDP) {
+              JsVar *address = jsvNewFromEmptyString(); // inet_ntoa replacement
+              JsVar *receiveInfo = jsvNewObject();
+              if (receiveInfo && address) { // could be out of memory
+                jsvAppendPrintf(address, "%d.%d.%d.%d", ((uint8_t*)&host)[0], ((uint8_t*)&host)[1], ((uint8_t*)&host)[2], ((uint8_t*)&host)[3]);
+                jsvObjectSetChildAndUnLock(receiveInfo, "address", address);
+                jsvObjectSetChildAndUnLock(receiveInfo, "port", jsvNewFromInteger(port));
+                jsvObjectSetChildAndUnLock(receiveInfo, "size", jsvNewFromInteger(num));
+                jswrap_dgram_messageCallback(connection, receiveData, receiveInfo);
```
gfwilliams: You could queue an event here instead I guess - that's what we seem to do elsewhere in the socket library. I'm not sure if it'd make much difference, but I guess the user could change the socket state under out feet this way round.

libs/network/socketserver.c
```diff
@@ -366,7 +402,9 @@ bool socketServerConnectionsIdle(JsNetwork *net) {
       // send data if possible
       JsVar *sendData = jsvObjectGetChild(socket,HTTP_NAME_SEND_DATA,0);
       if (sendData && !jsvIsEmptyString(sendData)) {
-        int sent = socketSendData(net, socket, sckt, &sendData);
+        JsVar *hostInfo = jsvObjectGetChild(connection, HTTP_NAME_HOST_INFO, 0);
```
gfwilliams: Maybe just check socketType==UDP here to save the minor slowdown for non-UDP connections?

libs/network/socketserver.c
```diff
@@ -479,7 +519,7 @@ bool socketClientConnectionsIdle(JsNetwork *net) {
       if (isHttp)
         hadHeaders = jsvGetBoolAndUnLock(jsvObjectGetChild(connection,HTTP_NAME_HAD_HEADERS,0));
       else
-        hadHeaders = true;
+        hadHeaders = (socketType&ST_TYPE_MASK)!=ST_UDP;
```
gfwilliams: I haven't run this, but I'd have thought hadHeaders should be true even for UDP? If there aren't any headers it's expecting it should be true I think.

libs/network/socketserver.c
```diff
@@ -534,20 +579,35 @@ bool socketClientConnectionsIdle(JsNetwork *net) {
...
-                if ((socketType&ST_TYPE_MASK)==ST_HTTP && !hadHeaders) {
-                  // for HTTP see whether we now have full response headers
-                  JsVar *resVar = jsvObjectGetChild(connection,HTTP_NAME_RESPONSE_VAR,0);
-                  if (httpParseHeaders(&receiveData, resVar, false)) {
-                    hadHeaders = true;
-                    jsvObjectSetChildAndUnLock(connection, HTTP_NAME_HAD_HEADERS, jsvNewFromBool(hadHeaders));
-                    jsiQueueObjectCallbacks(connection, HTTP_NAME_ON_CONNECT, &resVar, 1);
+
+                if ((socketType&ST_TYPE_MASK)==ST_UDP) {
+                  JsVar *address = jsvNewFromEmptyString(); // inet_ntoa replacement
```
gfwilliams: The client/server code was pretty badly duplicated before - did you come across any reason why we couldn't un-duplicate it? It's probably a bit too much to add to the PR, but I could file an issue for it afterwards?

gfwilliams:
Thanks - this looks really nicely done!

I mentioned the host/port thing, but maybe it's fine - it just feels like overkill.

in `jswrap_dgram_socket_send` you have the data, then have to pack it into an object, and then unpack it, then pass it as arguments. Potentially it wouldn't be much worse to just send everything in the `data` argument?

I don't know though - maybe it's just more messy.

opichals:
Thank you for the review. I will get back on individual items later hopefully over the weekend.

I get the notes about the host/port in the interface but I wanted rather not to introduce additional data buffer copying. In this context the receiveInfo and hostInfo creation sounds better to me. Or do you have any buffer no-copy-concat capabilities built-in? What do you think?

gfwilliams:
>  do you have any buffer no-copy-concat capabilities built-in?

Well, for send, you can just access that part of the array with `&buf[8]` or similar? clientRequestWrite appends a lot of the time anyway, so if you were using that you're not really making anything slower. It'll be much faster than allocating/deallocating objects.

For receive, you have a buffer of data available that you can put data in as well, so I think you're fine there too. There wouldn't actually be any significant amount more copying than before (no more copying of the data at least).

The real bonus though is that right now I think if you call `jswrap_dgram_socket_send` twice in quick succession then both packets get merged into one. It'd be a real pain to fix with the current solution. By having a packet_length header in the actual data you send, you can actually fix that really easily since the 'send' implementation can just take however many bytes make up one packet.

opichals:
Thanks for the description! You are right, it sounds like the data buffer copies are already there so it should be easy to put the port:host ahead of the payload and be fine. I will have a look at that.

opichals:
@gfwilliams I have updated the code so that the host:port is embedded within the send/recv buffer. I am not really convinced it looks any better. But I guess like you mentioned before it will perform better.

opichals:
> socketserver.c: The client/server code was pretty badly duplicated before - did you come across
> any reason why we couldn't un-duplicate it?

I think it could be deduplicated. I would prefer not to combine the UDP and the refactor work.

gfwilliams:
Wow, thanks! As you say I'm not sure the changes themselves look that much better, but:

* You can now send multiple UDP packets to different addresses and they don't get confused
* It should perform a little better
* On devices that don't support UDP all the arguments to send/recv still make sense
* We can now remove the TLS bitfield from netSend

I'm pretty happy with that really.

> I would prefer not to combine the UDP and the refactor work.

Totally agree :) Do you want me to look at doing that refactor before or after we merge this in? I guess before it could cause some merge headaches :)

opichals:
>  I guess before it could cause some merge headaches :)

Yes, my preference would be to do it after this is in master.

opichals:
@gfwilliams I guess adjusting the network_esp8266 would be nice as this was working before I moved the host:port to buf. As you say other platforms should just work as they were without the network.h interface changes once they are made to match the send/recv signature changes and UDP support can be then added incrementally to those.

I should find time for this today/tomorrow the evening, but if you feel like working on this feel free. :)

opichals:
@gfwilliams Went through the network_* stuff and synced the createsocket/recv/send signatures and removed inappropriate jsWarn() calls in favor of a DBG macro.

It seems to be compiling. I changed the network_esp8266 to work with the buf embedded host:port though have only tested on linux so far.

MaBecker:
@opichals I can jump in and test  on ESP8266. 

Can you please share you linux test code?

MaBecker:
There are some helpful tests created by @tve

https://github.com/espruino/Espruino/tree/master/libs/network/esp8266/tests/projects

opichals:
@MaBecker I created a basic test here https://github.com/espruino/Espruino/pull/1198/files#diff-bf1f203d99c827faf97580d231351b4e.  There are other network related tests (tests/*_net*.js).

Then I usually test the [mdns module I crafted so far](https://gist.github.com/opichals/1be27a74571b6ca2e1cc02b3587b78b8).

opichals:
Added size to the UDP buf header in f3378c4 so multiple socket.send() calls are possible.

The mdns module mentioned above working as expected on both OSX (network_linux) and ESP8266 d1 mini board.

opichals:
@gfwilliams After the esp8266 startMDNS() decision this is _merge at will_ from my point of view.

wilberforce:
@opichals 

Will need to update the esp32 code to the same as the Linux network updates after this pull request is done?

opichals:
@wilberforce Yes, the behavior should not change by merging this. The board specific UDP support is to be added incrementally in future PRs.

gfwilliams:
This looks great to me - thanks for all your work on this!

PR #1202, opened 2017-07-23
fix ESP32 'Not connected to the internet'

wilberforce:
Fold ESP32 fixes to ESP32

https://github.com/espruino/Espruino/pull/1197/commits/ca67ee67bc881bd4a507a867ed5de4cada932e66

Thanks to @iotong


gfwilliams:
Thanks!

PR #1222, opened 2017-08-18
Add package python-serial

wilberforce:


gfwilliams:
Great, thanks! Shame we have to pull it in for all builds, but hey - it's probably pretty quick.

PR #1223, opened 2017-08-19
Fix ESP32 travis build

wilberforce:
Sorry - no way of testing this without a pull request.


wilberforce:
Working now - please merge.

There where 2 issues - ADC count causing undefined, and pyserial import failing.

Failed on this board  BOARD=NUCLEOL476RG
https://travis-ci.org/espruino/Espruino/jobs/266390674

boards/NUCLEOL476RG.py
```diff
@@ -50,7 +50,7 @@
   'usart' : 3,
   'spi' : 4,
   'i2c' : 3,
-  'adc' : 1,
+  'adc' : 2,
```
wilberforce: According to this - there are 3 adc:
https://developer.mbed.org/platforms/ST-Nucleo-L476RG/

however unsure what the knock on effect of changing to 3 would be, so changed to 2 so the compile will work!

gfwilliams: Ahh - thanks - yes, this should be 3. I'll change it.

Sorry about breaking that bit of the build - I guess I didn't notice because of the other thing.

Now we know what the fix it I'll move the pyserial install into `provision.sh` though - then it's only installed for ESP32 where it's needed and the other builds stay a bit faster

PR #1234, opened 2017-08-29
Date timezone handling fixes.

opichals:
No tz-containing date strings are considered of local timezone.
No time-containing date strings are considered GMT.


gfwilliams:
Nice - thanks!

PR #1235, opened 2017-08-29
Introduced sockerserver:serverAddMembership

opichals:
This avoids the HTTP_NAME_OPTIONS_VAR server property leaking the
socketserver.c local abstraction.


gfwilliams:
Nice idea - thanks!

Issue #1206, opened 2017-07-26
NodeMCU.D0 value is undefined on ESP8266


fedy2:
I'm using a NodeMCU Amica board with Espruino. I'm printing the D0 equivalent pin name:

`console.log(NodeMCU.D0);`

But I get undefined as value:

`>undefined`

MaBecker:
It is defined, but returns undefined 

https://github.com/espruino/Espruino/blob/master/targets/esp8266/jswrap_nodemcu.c#L31

board definition names that pin B16

https://github.com/espruino/Espruino/blob/master/boards/ESP8266_BOARD.py#L59

MaBecker:
GPIO 16 is used to  wakeup the ESP from deep sleep and doesn't have an interrupt associated with it, so  this is why this port is not available. 

https://github.com/espruino/Espruino/blob/master/boards/ESP8266_4MB.py#L114

Please share what you like to use NodeMCU.D0 : GPIO16 for  to figure out if it makes sense to enable GPIO16

fedy2:
I'm using the GPIO16 to connect a DHT22 sensor. Why disable it a priori?

MaBecker:
just tested 
```
>console.log(NodeMCU.D0);
D16
=undefined
```
so lets create a pull request

MaBecker:
Running some test - It does not work ......

GPIO16 can be INPUT, OUTPUT, or INPUT_PULLDOWN_16 but not input_pullup, like in DHT22 module

    setTimeout(function() {pinMode(ht.pin,'input_pullup');},1);

code:

```
var dht = require("DHT22").connect(NodeMCU.D0);

setInterval(function() { 
  dht.read( function(a){
    console.log("Temp is "+a.temp.toFixed(1).toString()+
                " and RH is "+a.rh.toFixed(1).toString());
  });},3000);

```

using NodeMCU.D0 or D16 produces a stack dump and reboot

```
ASSERT(channel>=EV_EXTI0 && channel<=EV_EXTI_MAX) FAILED AT src/jsdevices.c:626
  #1[r3,l2] Object {
    #2[r1,l2] Name String [1 blocks] "\xFF"      #3[r1,l2] Object {
        #6[r1,l2] Name String [2 blocks] "timers"          #8[r2,l2] Array(2) [
            #59[r1,l3] Name Integer 1              #47[r1,l2] Object {
                #50[r1,l2] Name String [1 blocks] "time"= int 83
[...]

PR #1243, opened 2017-09-14
ESP8266: Add GPIO16 as D16 without watch (#1206)

MaBecker:


gfwilliams:
This looks great - thanks!

gfwilliams:
Just to add - you can do soft PWM/I2C/SPI/etc with this too now, so it's a pretty powerful addition.

MaBecker:
Yes, will add this to Cange Log as well

Issue #837, opened 2016-04-08
FlashEEPROM.readMem() causes reset on esp8266


ceremcem:
While following [the documentation](http://www.espruino.com/FlashEEPROM), I faced with the following situation: 

```
getFile = function(){
  var f;
  f = new (require("FlashEEPROM"))(0x076000);
  f.endAddr = f.addr + 1024;
  return f;
};
```

```
>f = getFile()
={ 
  "flash": function () { [native code] }, 
  "addr": 483328, "endAddr": 484352 }
>f.read(0)
=new Uint8Array([104, 101, 108, 108, 111])
>E.toString(f.read(0))
="hello"
>f.readMem(0)
=" ets Jan  8 2013,rst cause:4, boot mode:(3,7)
wdt reset
load 0x40100000, len 1396, room 16 
tail 4
chksum 0x89
load 0x3ffe8000, len 776, room 4 
tail 4
chksum 0xe8
load 0x3ffe8308, len 540, room 4 
tail 8
chksum 0xc0
csum 0xc0
2nd boot version : 1.4(b1)
  SPI Speed      : 40MHz
  SPI Mode       : QIO
  SPI Flash Size & Map: 4Mbit(256KB+256KB)
jump to run user1 @ 1000
rlrl��Loading 6525 bytes from flash...
Running onInit()...
connected? err= null info { 
  "ip": "192.168.2.193", 
  "netmask": "255.255.255.0", 
  "gw": "192.168.2.101", 
  "mac": "18:fe:34:d6:72:f6"
 }
```

gfwilliams:
Thanks - I guess this might be something to do with how flash memory is mapped into the address space. Presumably `E.memoryArea` should add some offset to addresses that are in flash.

For now, just use `f.read()` - which should work fine? `readMem` is only really useful when you're storing massive amounts of data.

ceremcem:
Okay, `f.read()` works fine and suffices right now.

MaBecker:
@ceremcem - can this be closed ?

gulfaraz:
@MaBecker readMem still resets the esp8266

MaBecker:
@gulfaraz  - can you add sample code  ?

ceremcem:
@MaBecker  Is my sample code passing your tests?

MaBecker:
@ceremcem - did some testing

 `E.memoryArea()`  is the reboot cause for this reboot on ESPs

MaBecker:
[...]

PR #1244, opened 2017-09-14
ESP8266: Enable unaligned reads for ESP8266_4MB (fix #1240,#837)

### LINKED ISSUE #1240: E.setBootCode not working for ESP8266_4MB
looks like code cannot be found during dump() and ESP8266.reboot()

````
>E.setBootCode("console.log('Hi.....');");
Erasing Flash..................
Writing...
Compressed 25600 bytes to 32
Checking...
Done!
=undefined
>dump();
pinMode(D0, "input_pullup", true);
pinMode(D4, "input_pullup", true);
pinMode(D5, "input_pullup", true);
pinMode(D12, "input_pullup", true);
pinMode(D13, "input_pullup", true);
pinMode(D14, "input_pullup", true);
pinMode(D15, "input_pullup", true);
// Code saved with E.setBootCode
=undefined
>require('ESP8266').reboot()
=undefined
>
 ets Jan  8 2013,rst cause:2, boot mode:(3,6)
load 0x40100000, len 2592, room 16
tail 0
chksum 0xf3
load 0x3ffe8000, len 764, room 8
tail 4
chksum 0x92
load 0x3ffe82fc, len 676, room 4
tail 0
chksum 0x22
csum 0x22
2nd boot version : 1.7(5d6f877)
SPI Speed : 80MHz
SPI Mode : QIO
SPI Flash Size & Map: 32Mbit(1024KB+1024KB)
jump to run user1 @ 1000
âìnìròn|ìll`b|rlnànâl`ârllþLoading 32 bytes from flash...
> 

````
maybe this is not working for a Flash map 4MB:1024/1024?

found this: https://github.com/espruino/Espruino/blob/master/src/jswrap_flash.c#L568-L572

@tve  do you remember if this value only works for ES8266-01?

### PR DESCRIPTION
**Pull request by MaBecker**

### CODE REVIEW

**Comment by MaBecker** [comment]
MFORCE32 = `xtensa-lx106-elf-gcc --help=target | grep mforce-l32`

is breaking the build 

    make: xtensa-lx106-elf-gcc: Command not found

File: make/family/ESP8266.make [hunk 1]
```diff
@@ -66,6 +66,14 @@ CFLAGS+= -fno-builtin \
 -Wno-parentheses -Wno-type-limits -Wno-unused-function -Wno-unused-value \
 -Wl,EL -Wl,--gc-sections -nostdlib -mlongcalls -mtext-section-literals
 
+# only use mfore-l32 if 4MB board for now
+ifdef FLASH_4MB 
+MFORCE32 = `xtensa-lx106-elf-gcc --help=target | grep mforce-l32`
+ifneq ($(MFORCE32),)
+CFLAGS += -mforce-l32
```
wilberforce: @MaBecker 

Does this need to be conditional?  Since the C code *relies* on MFORCE32 - if at this point the compiler does not have the flag you are up sh*t creek anyway!

gfwilliams:
Is this ready to go?

MaBecker:
Yes

PR #1245, opened 2017-09-14
Drop socketIsHTTPS in favor of SocketType

opichals:


gfwilliams:
Thanks! That's a great cleanup.

PR #1246, opened 2017-09-14
Support for dgram.createSocket({ reuseAddr: true })

opichals:


opichals:
The question is whether Espruino needs to follow the nodejs API in this particular thing.

The libs/crypto/mbedtls/library/net.c does always use SO_REUSEADDR.

gfwilliams:
Not sure I understand - you mean the nodejs API defaults to `reuseAddr: true`?

opichals:
nodejs implements the `reuseAddr` option to control the use of SO_REUSEADDR.

The [mbedtls uses it unconditionally](https://github.com/espruino/Espruino/blob/master/libs/crypto/mbedtls/library/net.c#L205).

After actually submitting this PR it occured to me that it could be better if Espruino behaves consistently with the mbedtls (always use SO_REUSEADDR, just as it is now in master) rather than introducing the same reuseAddr option as nodejs does. Perhaps other boards' SDKs/network stacks would just not care or support this option and therefore it might end up in an inconsistent support across boards?

gfwilliams:
I don't think we're using any of that code in mbedtls are we?

At the moment mbedtls only works though the `network.c` file, where we manually specify what commands we want to use for send and receive, and initiate everything ourselves: https://github.com/espruino/Espruino/blob/master/libs/network/network.c#L608

There's no support for UDP in it whatsoever.

opichals:
You are right Espruino doesn't use that mbedtls code...

I was just wondering whether other stacks would have this option supported (or rather if it is possible to turn using SO_REUSEADDR off in all cases) but I guess we can decide or document such exceptions later.

Merge/close at will.

PR #1253, opened 2017-09-26
Stm32 l496 family

tbistm:
L496 family with its pins description.

It is done through 2 new boards introduction.
Both having USB OTG FS, until 256 KB of RAM currently used.
Last STM32cube FW L4 (V1.9.0) is also included.


gfwilliams:
Hi - just wondering, what was the difference between this and https://github.com/espruino/Espruino/pull/1250, https://github.com/espruino/Espruino/pull/1251 and https://github.com/espruino/Espruino/pull/1252? Were those just accidental?

Do we need the `icf` files included? I'd hoped that most of the linking for STM32 stuff would be done by the linker script generator.

tbistm:
Hi,

#1250 , #1251 , #1252 could not pass because of compilation errors with Travis on last commit
(commit called : “More RAM for USART exchanges when it is possible”)
I could not reproduce these errors in my development environment. It was not systematic and occurred on different boards.
As this commit can be viewed as optional (optimization better to have for performance) , I finally produce #1253 without this commit.

Another Pull Request will be proposed later.

About icf files, they are used for IAR and they are not needed in #1253 since it is compiled with gcc only.
They are here because they are within the cube delivery by default and also because I saw them on the espruino master.

Thierry

gfwilliams:
Ah - thanks! I'll merge then.

Do you see any problem if I took the icf files out? I must have just missed them when the other bit of the SDK went in.

gfwilliams:
Just to add - the build errors *may* have just been Travis - occasionally the VM instances they spin up have problems downloading the files they need.

tbistm:
Hi,

The icf files can be removed, no problem with that.

Thierry

gfwilliams:
Ok, just done! In future, please can you avoid committing manuals too? I just found out that about 1/2 the size of Espruino's total repository is made out of STM32L reference manuals :)

PR #1255, opened 2017-10-09
Added Support for Arduino Due

ThomasChr:
This PR includes the full Environment to make Code for Atmel SAMD Devices (as used by Arduino), and also a working Espruino implementation for the Arduino Due.
The Arduino Due has a working REPL, Timing accurate to 10 Microseconds, all Pins can be used and the save()-Function does it's job and saves the code in the second flash bank.
What's left:
- The other UARTs
- ADC
- DAC
- Pin watch
- I2C
- PWM
.
.
.

And much more I can't think of right now. But it should be very straightforward to implement those features.

I personally think the code is ready to merge, but I'm quite sure I've made some mistakes. So please see this PR as a discussion about my bad coding style which will eventually lead to merging. Hopefully :-)


gfwilliams:
Merged! It just doesn't show as such because I did a filter-branch to remove some of the extra files.

Thanks! This is looks great - and it builds easily using all the existing ARM toolchain

PR #1278, opened 2017-11-16
socketSendData() to use dataLen buffer for UDP

### LINKED ISSUE #1275: socketSendData() cutting UDP messages
The `socketSendData()` allocates only the `net->chunkSize` and relies on the fact that additional data will be send during a another call (TCP). That is however not true for UDP where the message is cut (only `net->chunkSize` bytes is sent).

opichals:
Fixes #1275

Note that this only changes the `dgramSocket.send()` method buffer handling where the available stack-1024 bytes is the limit for the data size.

The receiving side is still hardcoded board/net implementation specific (the `net->chunkSize` sized buffer is used).


gfwilliams:
Looks awesome - thanks!

Issue #1276, opened 2017-11-16
http.createServer().listen() should return a Server instance


opichals:
server.listen() is void unlike the [node.js implementation](https://nodejs.org/dist/latest-v8.x/docs/api/net.html#net_server_listen_port_host_backlog_callback) where it returns an instance of the server.

As the fix is trivial it appears to be worth fixing for better developer compatibility.

gfwilliams:
Just fixed :)

Also I noticed that dgram.bind returned the parent without locking it again - but fixing it causes memory leaks - so I think there might be something iffy going on with locks in dgram - perhaps an extra lock somewhere else where there shouldn't be one... It might be worth looking into as it could cause problems later on?

PR #1279, opened 2017-11-16
dgramSocket.bind() returns the dgramSocket instance

opichals:
Added the required jsvLockAgain(parent).

The tests/test_dgram_socket.js now also works directly
in node.jsl (tested in v6.9.4).


opichals:
> from https://github.com/espruino/Espruino/issues/1276#issuecomment-344870411:
>
> Also I noticed that dgram.bind returned the parent without locking it again - but fixing it causes memory leaks - so I think there might be something iffy going on with locks in dgram - perhaps an extra lock somewhere else where there shouldn't be one... It might be worth looking into as it could cause problems later on?

After adding the `"return"` binding annotation it does not exhibit any leaks in any of the `tests/test_dgram_*` tests

gfwilliams:
Thanks! I'd forgotten about the `return` in JSON :)

I think you need to put ` parent = jsvLockAgain(parent);` right at the start though? It looks like if the function's called when you're offline, it'll return an unlocked instance of the parent?

opichals:
You are right with the beginning. PR updated.

gfwilliams:
Thanks!

PR #1285, opened 2017-12-08
ESP32  jshGetSystemTime() make interrupt tolerant

wilberforce:
Pull @jumjum123 changes from ESP32 branch into master

### DIFF
--- a/targets/esp32/jshardware.c
+++ b/targets/esp32/jshardware.c
@@ -553,9 +553,12 @@ JsVarFloat jshGetMillisecondsFromTime(JsSysTime time) {
 /**
  * Return the current time in microseconds.
  */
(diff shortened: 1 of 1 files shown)


opichals:
Perhaps there should be an initializer like this instead of just undefined content?
```
   portMUX_TYPE mux = portMUX_INITIALIZER_UNLOCKED;
```

Adding an interesting article link about the ESP32 synchronization stuff https://techtutorialsx.com/2017/09/30/esp32-arduino-external-interrupts/

gfwilliams:
Just added that for you.

wilberforce:
@gfwilliams 

Thanks Gordon - that was unexpected!

gfwilliams:
Just saves everyone some time if it's a one-liner :)

PR #1286, opened 2017-12-12
Lazy dgram (UDP) createSocket

### LINKED ISSUE #1281: dgram (UDP) `createSocket` behaviour
Hi @opichals,

I'm just looking at implementing UDP support on Espruino WiFi/ESP8266/etc and I've noticed that `createSocket` seems to try and create a socket with `netCreateSocket` immediately (for port 0).

If using `bind` it'll then close that socket and reopen at a different port.

The [Node.js docs](https://nodejs.org/api/dgram.html#dgram_socket_send_msg_offset_length_port_address_callback) for `send` seem to say:

> If the socket has not been previously bound with a call to bind, the socket is assigned a random port number and is bound to the "all interfaces" address

And it looks like `jswrap_net_connect` doesn't actually *have* to connect immediately (for instance it doesn't for HTTP). Is there something subtle that I'm missing, or do you think it would make sense to change this around a bit so that it only automatically connected on `send` or `bind`?

opichals:
Fixes #1281

### DIFF
--- a/libs/network/jswrap_net.c
+++ b/libs/network/jswrap_net.c
@@ -407,7 +407,7 @@ JsVar *jswrap_net_connect(JsVar *options, JsVar *callback, SocketType socketType
   JsVar *rq = clientRequestNew(socketType, options, callback);
   if (unlockOptions) jsvUnLock(options);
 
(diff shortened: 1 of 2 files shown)


gfwilliams:
Neat - thanks!

Issue #1271, opened 2017-11-05
dgram.createSocket('udp4').send() signature to support offset and length


opichals:
As there is bunch of code probably going to be ported from node.js it would be very nice to conform to its complete [send() signature](https://nodejs.org/api/dgram.html#dgram_socket_send_msg_offset_length_port_address_callback).

```js
socket.send(msg, [offset, length,] port [, address] [, callback])
```

I would have created a PR but I struggle to find a way to express all this in the C comments-based binding language properly (the /**JSON... */).

opichals:
With [these changes](https://github.com/espruino/Espruino/compare/master...opichals:dgramsocket-5args-send?expand=1) it appears that the 5th (address) argument is not passed in:

```sh
$ ./espruino  ./tests/test_dgram_multi.js 

41234,"localhost", 41234 "localhost"
3,2, 51234 Segmentation fault: 11
```

Perhaps due to the amount of argument info that can fit into JsnArgumentType's size as noted at https://github.com/espruino/Espruino/blob/master/src/jsnative.c#L106?

Any hints on how to approach this?

gfwilliams:
Sure - just put one argument as `JsVarArray` (grep the source to see how it's done). You can then access that as a standard array, and can have however many args you want.

To be honest the build *should* really fail if something like that happens. I'll see what I can do about that.

opichals:
@gfwilliams Thank you, PR submitted.

The callback argument support would need a bit more work there so I left it unsupported for now.

PR #1287, opened 2017-12-15
Added dgramSocket.send() offset + len support

opichals:
Both of the following #send() signatures supported:
  client.send('msg', port, address);
  client.send('  msg  ', 2, 3, port, address); // sends 'msg' message

Partial fix for #1271


gfwilliams:
Thanks! I think you might need to jsvUnLock  to port and address though or you'll get a memory leak - but then you also need to be careful when you do `port = offset;`.

opichals:
Thanks for the review. Should be fixed now.

gfwilliams:
Great - thanks!

PR #1289, opened 2017-12-16
infor-link.com bare NRF52832 board

vbolshakov:
Please add bare NRF52832 board (like infor-link.com one).


gfwilliams:
Hi - How is this different to the NRF52832DK image? It just has the LEDs and Button defs removed?

vbolshakov:
Right pin definitions and no LEDs and buttons.

PR #1316, opened 2018-01-21
Wiznet W5100 support

opichals:
I defined the \_WIZCHIP_ to 5100 I got some compile time errors in Ethernet/socket.c so I went to upgrade the ./libs/network/wiznet/Ethernet folder to the latest [Wiznet/ioLibrary_Driver v3.1.3](https://github.com/Wiznet/ioLibrary_Driver/tree/master/Ethernet).

It seems that the code is now too large to fit the flash on the original Espruino boards?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,6 +177,9 @@ ifdef USE_NET
 ifndef LINUX
 ifdef WIZNET
 USE_WIZNET=1
+ifdef W5100
+USE_WIZNET_W5100=1
+endif
 else ifeq ($(FAMILY),ESP8266)
 USE_ESP8266=1
 else ifeq ($(FAMILY),ESP32)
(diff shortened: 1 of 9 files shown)


opichals:
The default W5500 build:

```sh
WIZNET=1 RELEASE=1 BOARD=ESPRUINOBOARD make
...
bash scripts/check_size.sh espruino_1v95.1392_espruino_1r3.bin
FAIL - size of 231460 is over 231424 bytes
```

The new W5100 build:
```sh
WIZNET=1 W5100=1 RELEASE=1 BOARD=ESPRUINOBOARD make
...
bash scripts/check_size.sh espruino_1v95.1392_espruino_1r3.bin
FAIL - size of 231884 is over 231424 bytes```

**Comment by opichals** [comment]
Also please note that this PR is so far just _blind software only work_ without any HW testing done.

**Comment by gfwilliams** [comment]
That's very cool - thanks! I'll try and give this a go on some actual hardware when I get a chance.

It's always been really tight trying to get everything to fit in the original board, so I'm not super surprised about this - but I wonder what got added to the driver that bulked it out so much :(

**Comment by gfwilliams** [comment]
Not having much luck here either. It looks like `ctlwizchip(CW_GET_PHYLINK` doesn't work for W5100 so can be commented out, but even after that I'm not able to set/get the IP with:

```
eth.setIP({
ip:"192.168.1.123",
subnet:"255.255.255.0",
gateway:"192.168.1.1",
dns:"8.8.8.8",
mac:"00:08:dc:ab:cd:ef"});
eth.getIP();
```

This should just be a matter of writing to SPI registers, so I'm not sure what's up. It might even be electrical since the shield is meant for 5v inputs... I wonder if it works so well off 3.3v data lines.

gfwilliams:
It works! It was because the shield I used actually had SPI on the tiny little header at the end - not on D11,12,13 like I was expecting.

gfwilliams:
All merged in - thanks!

PR #1317, opened 2018-01-21
change baud

MaBecker:


gfwilliams:
Was there a reason for this? The higher rates seemed reasonably reliable.

MaBecker:
ok, do keep it as it is

Issue #1333, opened 2018-02-12
ESP32: DeepSleep is missing ?


jvvood:
While I found the deepsleep function for ESP8266, but  not for ESP32.
Do You plan to implement the deepsleep for ESP32 ?

wilberforce:
There have been only two people working on the esp32 port to date, and just in spare time.

There are no current plans at present.

PR #1334, opened 2018-02-12
Deepsleep for ESP32

jvvood:
Only timer based wakeup is working.

ESP32.deepsleep(time_in_milliseconds)

#1333

### DIFF
--- a/targets/esp32/jswrap_esp32.c
+++ b/targets/esp32/jswrap_esp32.c
@@ -50,6 +50,22 @@ void jswrap_ESP32_reboot() {
 /*JSON{
   "type"     : "staticmethod",
   "class"    : "ESP32",
+  "name"     : "deepsleep",
+  "generate" : "jswrap_ESP32_deepsleep",
+  "params"   : [ ["ms", "int", "Sleeptime in ms"] ]	
+}
+Put device in deepsleep state for "ms" milliseconds.
+*/
+void jswrap_ESP32_deepsleep(int ms) {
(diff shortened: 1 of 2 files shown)


gfwilliams:
Looks great, thanks!

@jumjum123 @wilberforce are you guys happy with this? If so I'll merge.

wilberforce:
@jrwood. Thanks for this.

There looks like there is a debug printf that could be removed.

@gfwilliams 
At this stage do we want to add a new esp32 deep sleep or should try to have a common `setDeepSleep`. I understand there are some esp8266 specific sleep functions too and the functionality is different....

gfwilliams:
Yeah, it's difficult. The standard `setDeepSleep` for Espruino is totally different - it just enables a proper low power sleep mode where the device wakes up *with the entire contents of RAM still there* - STM32 has a different set of sleep modes to ESP8266. There's no timeout because it's all done automatically.

... so use of the whole `deepSleep` thing of the ESP8266/ESP32 is super confusing. Personally I'm happy with the current solution for the moment since the functionality is really pretty EspressIf-specific - but I'd bring it in-line with the ESP8266 one to avoid even more confusion:

* Change capitalisation so it's `deepSleep`
* Change milliseconds to microseconds (even though IMO milliseconds makes way more sense here)

jvvood:
@gfwilliams
I've changed them according to your suggestions.

gfwilliams:
Great - thanks for doing those changes so quickly - merging!

PR #1341, opened 2018-02-14
ESP32: Fix jshFlashGetMemMapAddress offset

wilberforce:
The romdata_jscode point is mapped by the MMU to the real address of the jscode partition. The offset from the start of flash for the code areas is 0x100000 so this has be removed in the direct access of memory case.

### DIFF
--- a/targets/esp32/jshardware.c
+++ b/targets/esp32/jshardware.c
@@ -718,13 +718,16 @@ void jshFlashErasePage(
 }
 
 size_t jshFlashGetMemMapAddress(size_t ptr) {
-  // Treat 0-based as external flash
-  // 0x3F400000 - 0x3F7FFFFF = external Flash
-  if (ptr < 0x400000)
-    return 0x3F400000 + ptr;
-
-  // Otherwise lets assume we know what we should be accessing
-  return ptr;
+   // if ptr is high already, assume we know what we're accessing
+  if (ptr > 0x10000000) return ptr;
+  // romdata_jscode is memory mapped address from the js_code partition in rom - targets/esp32/main.c
+  extern char* romdata_jscode;
+  if (romdata_jscode==0) {
+    jsError("Couldn't find js_code partition - update with partition_espruino.bin\n");
+    return 0;
+  }
+  // Flash memory access is offset to 0, so remove starting location as already accounted for
+  return &romdata_jscode[ptr - 0x100000 ];
 }
 
 unsigned int jshSetSystemClock(JsVar *options) {


wilberforce:
```
>var f=require("Flash");
=function () { [native code] }
>var st=require("Storage");
=function () { [native code] }
>st.eraseAll();
[Flash] EraseAll
=undefined
DEBUG FILES 0 live
=undefined
>st.list();
=[  ]
>st.read('test1');
=undefined
>st.write('test1','ABCD');
[Flash] CreateFile
[Flash] CreateFile new 0x00100010
=true
>st.read('test1');
="ABCD"
>st.debug();
DEBUG FILES 20 live
PAGE 0x00100000 (4096 bytes) - 20 live 4076 free
0x00100010	test1	(4 bytes)
=undefined
>f.read(4,0x00100010);
=new Uint8Array([65, 66, 67, 68])
> 
```

wilberforce:
FYI @gfwilliams 

`/home/travis/build/espruino/Espruino/src/jsutils.h:283:0: warning: "assert" redefined`

https://travis-ci.org/espruino/Espruino/jobs/341679050#L1451

```
/home/travis/build/espruino/Espruino/src/jsutils.h: At top level:
/home/travis/build/espruino/Espruino/src/jsutils.h:283:0: warning: "assert" redefined
  #define assert(X) do { } while(0)
 ^
In file included from /home/travis/build/espruino/Espruino/esp-idf/components/esp32/include/esp_assert.h:17:0,
                 from /home/travis/build/espruino/Espruino/esp-idf/components/soc/esp32/include/soc/soc.h:20,
                 from /home/travis/build/espruino/Espruino/esp-idf/components/soc/esp32/include/soc/gpio_reg.h:17,
                 from /home/travis/build/espruino/Espruino/esp-idf/components/driver/include/driver/gpio.h:19,
                 from /home/travis/build/espruino/Espruino/esp-idf/components/esp32/include/esp_deep_sleep.h:19,
                 from /home/travis/build/espruino/Espruino/esp-idf/components/esp32/include/esp_system.h:21,
                 from targets/esp32/jshardwareTimer.c:20:
/home/travis/build/espruino/Espruino/esp-idf/components/newlib/include/assert.h:16:0: note: this is the location of the previous definition
 # define assert(__e) ((__e) ? (void)0 : __assert_func (__FILE__, __LINE__, \
```

Due to the global effect of adding  a check width `#ifdef` to see if the assert macro already exists I've raised the issue rather than doing a pull request. Let me know if you want me to a raise a new issue for this.

wilberforce:
If you commit above, I can then clean up the code. I have worked out how I can can get the parameters from platform.h and remove the constant 0x100000 and use `#define FLASH_SAVED_CODE_START            1048576` instead.

I can then move the code to main.c and adjust `&romdata_jscode` so that the offset can be:
`return &romdata_jscode[ptr];`  as you first had it.

perhaps rename romdata_jscode to `flash_saved_code_start` or `mmu_flash_saved_code_start`?

gfwilliams:
With the assert, I don't see it's an issue to check if it already exists and only include if it doesn't.

It seems this PR still doesn't solve the underlying issue, it just makes the memory mapping even more insane. Anything below 0x100000 now accesses some totally random bit of memory, 0x100000 to 0x200000 is jscode(ish), and then above that but below 0x10000000 is random too.

However it doesn't tally even remotely with what is given to `jshFlash*` as far as I can tell? Sure, jshFlashGetFree returns the same 0x100000 (which isn't free, because it's now used for saved code), but when you write with jshFlashWrite it writes to the *actual SPI flash address of 0x100000*, and when you read using memoryArea it reads from the `jscode` partition which is somewhere else

The magic 0x100000 you're seeing is just randomly defined here: https://github.com/espruino/Espruino/blob/master/boards/ESP32.py#L56

Because that seems to be an *actual unmapped address in SPI flash that was considered 'safe' to use*.

So I think we want to:

Either:

* Make our own magic Espruino memory space where 0->0x10000 *only* accesses the 64k jscode partition, with E.memoryArea, peek, *and* jshFlashWrite. I this case it's nice and safe but none of the rest of the 4MB of memory can be accessed.
* We use a proper flat memory space, where 0->0x400000 maps to SPI flash, however we actually define `FLASH_SAVED_CODE_START` to be `romdata_jscode` so it writes data directly into the jscode page. If we can't actually access the flash memory as one flat 4MB memory-mapped chunk (I'm still unsure about this, but that fact that the current code in master works after `eraseAll` implies we can) then we remap just the area the corresponds to where the jscode partition is to the romdata_jscode address.

gfwilliams:
https://dl.espressif.com/doc/esp-idf/latest/api-guides/partition-tables.html is an interesting reference. So it looks like 0x100000 is just overwriting the 'factory' app, which is in this case Espruino :(

I think it'd be good to be able to read all the flash with the flash API from JS, including the partition table - so we do really want a flat memory space... So it's just a matter of 'patching' the flash address to a memory mapped address.

When `romdata_jscode` is set up, I think ideally we need:

* The memory mapped address (this is romdata_jscode)
* The size of the partition (`p->size`?)
* The actual offset of the partition in flash memory (this'll be in there somewhere)

Then we can change the flash memory map to:

```
if (addr >= jscode_offs_in_flash && addr < jscode_offs_in_flash+jscode_len)
  return romdata_jscode + addr - jscode_offs_in_flash;
```

And we should remove the existing free area from FlashGetFree unless we're adding a second partition or are able to check the partition table and point to an unallocated area.

wilberforce:
0x100000 is not random - it's what I defined in the custom partition table for that I made for espurino:

https://github.com/espruino/EspruinoBuildTools/blob/3ac785e699e7f472b5439ff3b0a4cf1dcfd80ef0/esp32/build/app/partitions_espruino.csv

```
#Name,      Type, SubType, Offset,    Size, Flags
#boot,      data, ,        0x1000,    4K
partition,  data, ,        0x8000,    0x1000
nvs,        data, nvs,     0x9000,    16K,
otadata,    data, ota,     0xd000,    8K,
factory,    app,  factory, 0x10000,   960K,
js_code,    data, ,        0x100000,  64K,
ota_0,      app,  ota_0,   0x200000,  960K,
storage,    data, ,        0x300000,  1M,
```

on start up, in main.c it looks up this partition by name, and then stores the physical address in `romdata_jscode`. 

I have tested the code and it working as it was before the storage changes.

The  flashFs filessytem is working as is  flash reads and writes.

The flashFS code uses the area at 0x300000.

If anything ota_0 should be removed as the ota stuff has not been implemented. This area could then be used for the `flash.getFreeArea()`.

gfwilliams:
Ahh - thanks for clearing that up. 

Yes, I totally agree on ota2 and `flash.getFreeArea()`.

So you're saying that with the current (master) build, `E.memoryArea(0x100000,16)` doesn't work and actually causes the device to crash? Because I got the impression from the reports on the other issue that it was ok?

Let's merge this anyway then, but it'd be nice to at the very least use FLASH_SAVED_CODE_START and length, and ideally use the values from the partition table.

PR #1353, opened 2018-02-26
New target: ET-STM32 Stamp module - initial port

ramangopalan:


ramangopalan:
Gordon, will you please consider merging the update for the Stamp module?

R

gfwilliams:
Thanks! I'm afraid I'm not pulling new board files into the master branch any more unless it's for a board that a company is actually paying maintenance for.

It's just getting too painful to maintain since every time the Makefile changes I now have to make changes to 50+ board files to keep them compiling (nobody else maintains them once they have been committed).

However I'll leave this open as I do want to find a way to keep ports like yours as part of Espruino. I'll come up with something soon and will merge this in.

gfwilliams:
I have now merged this into [`UNMAINTAINED_BOARDS`](https://github.com/espruino/Espruino/tree/UNMAINTAINED_BOARDS)

PR #1377, opened 2018-04-22
HTTP POST & Transfer-Encoding: chunked refactor

opichals:
Fixed the http request/response  'data', 'end' and 'close' event semantics.

Added two added tests for http now pass in espruino as well in node.js
```
./tests/test_http_post.js
./tests/test_http_post_chunked.js
```

### DIFF
--- a/libs/network/socketserver.c
+++ b/libs/network/socketserver.c
@@ -18,11 +18,14 @@
 #include "jshardware.h"
 #include "jswrap_net.h"
 #include "jswrap_stream.h"
+#include "jswrap_string.h"
+#include "jswrap_functions.h"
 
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks - looks really good. Are we sure this doesn't break any of the TCP/UDP stuff?

opichals:
@gfwilliams I tried to go through the related tests and none of them failed:

I tested
```
tests/test_net*
tests/test_http*
tests/test_dgram*
```

Not aware of anything more there is?

gfwilliams:
Ok, great! No, I don't think there are other network tests in there...

I wonder whether the network portion of the tests would run on Travis (seems scary if they'd allow people to start their own servers on the machines), but yes, it'd make sense to make the Linux build automatically run through all the tests.

PR #1381, opened 2018-04-25
capsulate gdb with ifdef DEBUG (fix #1378)

### LINKED ISSUE #1378: user_main.c and gdb stuff
Does anyone know if this lines can be removed?

https://github.com/espruino/Espruino/blob/master/targets/esp8266/user_main.c#L244-L245

https://github.com/espruino/Espruino/blob/master/make/family/ESP8266.make#L99-L100

MaBecker:


MaBecker:
well old story, too much stuff

PR #1384, opened 2018-04-26
ESP8266 build specific stuff to make SHA256/512 switchable

MaBecker:


gfwilliams:
Ahh - thanks - so does the build fail if that objcopy fails?

MaBecker:
that what you get without

LD espruino_esp8266_partial.o
xtensa-lx106-elf-objcopy: 'libs/crypto/mbedtls/library/sha256.o': No such file
make: *** [espruino_esp8266_partial.o] Error 1

wilberforce:
I have seen in other linker scripts that the sections for libs can be done in the linker script rather than a separate pass in the make process, so could look at this at some stage.

PR #1386, opened 2018-04-26
Telnet - prevent New interpreter error: FIFO_FULL

wilberforce:
Allow TCP/IP in Telnet have flow control in the firmware itself.

http://forum.espruino.com/conversations/317981/#comment14206992


gfwilliams:
Thanks!

PR #1390, opened 2018-05-01
Compare Transfer-Encoding value case-insensitively

opichals:
By RFC 2616: All transfer-coding values are case-insensitive.


gfwilliams:
Thanks - that's an annoying bit of the spec :)

PR #1393, opened 2018-05-02
 Implement ignore-case RegExp flag, /something/i

opichals:
Examples from tests that match:
```
/world/i.exec("Hello World")
/[bac]*d/i.exec("Hello aBcD")
```


gfwilliams:
Very nice - thanks for tweaking String.toUpper/LowerCase to use the same functions as well :)

PR #1394, opened 2018-05-03
Add ignoreCase string comparison helper

opichals:
Add jsvIsStringIEqualAndUnLock helper. No extra allocation solution
for cases where an ignoreCase comparison is necessary.


gfwilliams:
Thanks!

PR #1396, opened 2018-05-03
Managed to improve image creation on MacOS

ukaoskid:
Added MacOS detection inside the image creation sh scripts.


gfwilliams:
Thanks! could you just alter the original line to use `IMGSIZE=$(expr $BOOTLOADERSIZE + $(du "$ESPRUINOFILE" | cut -f1))`? That works on Linux too, so then you can avoid the OS-specific ifdef?

ukaoskid:
Sure! Got to do that. Makes sense.

gfwilliams:
Thanks!

PR #1399, opened 2018-05-05
Implement RegExp character set ranges, /[a-z]/

opichals:
A simple implementation with a known issue that the range's first character must not be a quoted.


gfwilliams:
Thanks!

Issue #1397, opened 2018-05-04
parseInt() not capable of parsing from large strings


opichals:
The issue seems to be that the `parseInt/Float` functions do not work on large then JS_NUMBER_BUFFER_SIZE strings because of [this condition](https://github.com/espruino/Espruino/blob/master/src/jswrap_functions.c#L150-L153).

As a consequence of this, the recently implemented [http chunked receive code](https://github.com/espruino/Espruino/blob/master/libs/network/socketserver.c#L318) doesn't work on larger packets (released in 1v97).

For the test extended as follows:

```diff
diff --git a/tests/test_parseint.js b/tests/test_parseint.js
index 1f4cfdb..1977303 100644
--- a/tests/test_parseint.js
+++ b/tests/test_parseint.js
@@ -20,6 +20,8 @@ var a = [
          parseFloat("1.e+1"), 10.0,
          parseFloat("1.2e+2"), 120.0,
          parseFloat("1200e-3"), 1.2,
+         parseFloat("123 with a further non-numeric string"), 123,
+         parseFloat("456 plus a string longer than 66chars which is the jsvGetString() size limit"), 456,
 ];
```

The test output:

```sh
$ ./espruino --test tests/test_parseint.js 
...
Uncaught Error: String too big to convert to float

 at line 24 col 99
... jsvGetString() size limit"), 456,
```

gfwilliams:
Yeah - short of writing separate string to int/float functions that work on JsVars I don't know how we avoid that. I guess we could make it allocate a large enough string on the stack if it was required? Or stringToInt could return the character at which it gave up, and if that character was the string end *then* we could error.

I wonder whether in reality many people hit this though? Using stringToInt to automatically stop parsing is reasonably rare behaviour.

opichals:
Thanks! :)

PR #1400, opened 2018-05-06
Do parseInt() on only pre-CRLF http chunked data

opichals:
This avoids the HTTP chunks decoding hitting the parseInt string
length issue https://github.com/espruino/Espruino/issues/1397


gfwilliams:
Thanks!

Issue #1380, opened 2018-04-25
rewrite wifi.save and restore to use the store lib 


MaBecker:
at the moment wifi.save is using a 4kB page to store some credentials and settings.

With the storage library it should be possible to store and restore this data and keep the wifi reconnect at boot time.

@gfwilliams  What do you think, is this  possible?

flash usage now for 512KB boards without FOTA

```
0x7B000		one wifi save page

0x7AFFF
...
0x78000		save code address, 3 pages

0x77000		one flash  area page

0x76FFF
...
0x01000		espruino_esp8266_user1.bin, max 117 pages  = max size 468 KB
```

after moving wifi connection data to storage

```
0x7BFFF	
...
0x78000		save code address, 4 pages

0x77FFF
...
0x01000		espruino_esp8266_user1.bin, max 118 pages  = max size 472 KB
```

gfwilliams:
Yes, I'd say that was a great idea - and very possible. Potentially it could be saved as JSON such that it could be parsed and then fed right into the `wifi.connect` function - it'd probably really tidy up the code and maybe even save some code space as well.

wilberforce:
If we are going to this trouble, may I suggest a further enhancement?
Store the list of pairs in an array, so that credentials for more than one ssid can be stored.
Potentially we could store the last connect to ssid, and attempt to connect that on restart.

Do we have encryption/decryption - if so we could stored encrypted password rather than plain text password in flash.

MaBecker:
Good Point @wilberforce 

Like to start with the config structure and replace  jshFlashWrite and  jshFlashRead with jswrap_storage_read and jswrap_storage_write

Are there some samples how to handle c structures  with storage functions?

gfwilliams:
There's no easy encrypt/decrypt we could rely on - after all, if you can read flash you could just read the decryption key as well.

I'd skip trying to save the config structure as-is - looking at the code it's full of hard-coded lengths and things, with strcpys all over the place.

Instead, just do something like this to save:

```
#define WIFI_CONFIG_STORAGE_NAME "wificonf"

JsVar *o = jsvNewObject();
if (!o) return;
jsvSetChildAndUnLock(o, "ssid", jsvNewFromString((char *)sta_config.ssid));
jsvSetChildAndUnLock(o, "password", jsvNewFromString((char *)sta_config.password));
// ...
JsVar *name = jsvNewFromString(WIFI_CONFIG_STORAGE_NAME);
[...]

PR #1401, opened 2018-05-08
rewrite wifi.save and restore to use the storage lib (imp #1380)

MaBecker:
* use .wificfg to store wifi credentials
* free wifi save page and add to save_code
* remove flash page 0x77000 and add to flash_available
* move wifi restore after JavaScript environment is up and running

### DIFF
--- a/boards/ESP8266_BOARD.py
+++ b/boards/ESP8266_BOARD.py
@@ -46,8 +46,8 @@ chip = {
   'saved_code' : {
     'address' : 0x78000,
     'page_size' : 4096,
-    'pages' : 3,
-    'flash_available' : 468, # firmware can be up to this size
+    'pages' : 4,
+    'flash_available' : 472, # firmware can be up to this size
   },
 };
 devices = {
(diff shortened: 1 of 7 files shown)


MaBecker:
Possible to merge?

gfwilliams:
Looks good - thanks :)

boards/ESP8266_BOARD.py
```diff
@@ -46,8 +46,8 @@
   'saved_code' : {
     'address' : 0x78000,
     'page_size' : 4096,
-    'pages' : 3,
-    'flash_available' : 468, # firmware can be up to this size
+    'pages' : 4,
+    'flash_available' : 472, # firmware can be up to this size
```
wilberforce: I understand the code size has increased by a page, however don't understand why the firmware can be larger?

MaBecker: Gave up one Wifi page and one flash, for details see #1380

wilberforce: > espruino_esp8266_user1.bin, max 118 pages  = max size 472 KB

Where did the one for flash come from? Isn't the phy data stored in the last block by the operating system?

PR #1404, opened 2018-05-08
Add String#match() implementation

opichals:


gfwilliams:
Thanks! We'll see how this works out memory-wise but I might make the whole thing ifndef SAVE_ON_FLASH because a few boards are struggling for free memory now

Issue #1358, opened 2018-03-18
Add missing option ssid_hidden for Wifi.startAP()


MaBecker:
Extend list of possible option with "hidden" to start a AP  visible or hidden.

AP is visible 

    "hidden" : 0          default value

AP is hidden

    "hidden" : 1         

Gordon where do I find the API reference for Wifi.startAP(), because  I like to add this to the reference too.

wilberforce:
@MaBecker 

https://github.com/espruino/Espruino/blob/0ad87619ddbbeefa91fc6e9ff712c34ed2e98c2a/libs/network/jswrap_wifi.c#L274-L276

MaBecker:
Thanks

what about ESP32 and hidden option?

MaBecker:
How do you create this nice looking source  code links?

wilberforce:
I did a search to find `startAP`
In the results page clicked on the line number 261 here libs/network/jswrap_wifi.c

Then clicked on line 272 and shift-clicked line 278. This updates the URL to have this line range.
Then cut and paste into the comment. I have added a single back tick to make this appear:

`https://github.com/espruino/Espruino/blob/0ad87619ddbbeefa91fc6e9ff712c34ed2e98c2a/libs/network/jswrap_wifi.c#L272-L278`

wilberforce:
> what about ESP32 and hidden option?

Here is an example
https://github.com/espruino/Espruino/blob/0ad87619ddbbeefa91fc6e9ff712c34ed2e98c2a/libs/network/jswrap_wifi.c#L291

Not sure you can use a condtition for part of the options list - you just say ESP32 specfic

gfwilliams:
Yes, there's no way to have one argument only for one platform. You'd just have to document it manually in the `/*JSON` comment

wilberforce:
Oh I think your asking if esp32 supports hidden  AP - yes!

Set as part of the struct here:

https://github.com/espruino/Espruino/blob/9ffc5162047e93200e1356c3334b05f5ab67db53/libs/network/esp32/jswrap_esp32_network.c#L851

MaBecker:
Cool Wilberforce

So adding option hidden has to take place in ESP8266,ESP32 and ESP8266WiFi.js - correct?

Just checked AT Instructions and figured out how to handle this

    AT+CWSAP=<ssid>,<pwd>,<chl>,<ecn>[,<max	conn>][,<ssid	hidden>]

[...]

PR #1407, opened 2018-05-09
add ESP8266 #1380, #1358

MaBecker:


MaBecker:
Wow - that was quick - Thanks.

PR #1408, opened 2018-05-09
Avoid double on('connect') for non-HTTP connects

### LINKED ISSUE #1405: Websockets now broken.
Hi @opichals,

It looks like the recent [HTTP chunked PR](46e4bc69d73631b63f62b11febb4f2c07ec1acf3) has broken websocket support. There's a post on it here: http://forum.espruino.com/conversations/320553/

Basically run this on node.js:

```
var WebSocketServer = require('ws').Server,
    wss = new WebSocketServer({
        port: 8080
    });

console.log("Server on 8080");

wss.on('connection', function connection(ws) {
    ws.room = [];
    ws.send("User Joined");

    ws.on('message', function(message) {
        message = JSON.parse(message);
        if (message.join) {
            ws.room.push(message.join);
        }
        if (message.room) {
            broadcast(message);
        }
        if (message.msg) {
            console.log("Server got: " + message.msg);
        }
    });

    ws.on('error', function(er) {
        console.log(er);
    })


    ws.on('close', function() {
        console.log('Connection closed')
    })
});

function broadcast(message) {
    wss.clients.forEach(function each(client) {
        if (client.room.indexOf(message.room) > -1 || message.room == 'all') {
            client.send(message.msg);
        }
    });
}
```

and this on Espruino (you can do it on the Linux build if you copy ws.js/ws.min.js to a directory called node_modules):

```
var host = "localhost";
var WebSocket = require("ws");
    var ws = new WebSocket(host,{
      path: '/',
      port: 8080, // default is 80
      protocol : "echo-protocol", // websocket protocol name (default is none)
      protocolVersion: 13, // websocket protocol version, default is 13
      origin: 'Espruino',
      keepAlive: 60,
      headers:{ some:'header', 'ultimate-question':42 } // websocket headers to be used e.g. for auth (default is none)
    });

ws.on('open', function() {
  console.log("Connected to server");
});

ws.on('message', function(msg) {
  console.log("MSG: " + msg);
});
```

Then you get `RangeError: Invalid WebSocket frame: RSV1 must be clear` from node.js

`ws.js` doesn't actually get around to sending anything, and it looks like `require("net").connect` calls back *twice* now, which causes `onConnect`/`handshake` in `ws.js` to be called twice, which breaks everything.

Could it be that it's now trying to parse headers even on connections that aren't HTTP?

It'd be good to get a fix out for this soon as it looks like it's broken in the 1v97 release :(

opichals:
Fixes #1405

### DIFF
--- a/libs/network/socketserver.c
+++ b/libs/network/socketserver.c
@@ -375,18 +375,23 @@ void socketReceived(JsVar *connection, JsVar *socket, SocketType socketType, JsV
   JsVar *reader = isServer ? connection : socket;
   bool isHttp = (socketType&ST_TYPE_MASK)==ST_HTTP;
   bool hadHeaders = jsvGetBoolAndUnLock(jsvObjectGetChild(reader,HTTP_NAME_HAD_HEADERS,0));
-  if (!hadHeaders && (!isHttp || httpParseHeaders(receiveData, reader, isServer))) {
-    hadHeaders = true;
-    jsvObjectSetChildAndUnLock(reader, HTTP_NAME_HAD_HEADERS, jsvNewFromBool(hadHeaders));
+  if (!hadHeaders) {
+    if (isHttp && httpParseHeaders(receiveData, reader, isServer)) {
+      hadHeaders = true;
 
-    if (isServer) {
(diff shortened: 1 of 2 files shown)


gfwilliams:
Thanks for fixing this so quickly!

PR #1410, opened 2018-05-09
Make on('message') a queued callback

opichals:
And optimize the socketReceived()/socketReceivedUDP() calls


gfwilliams:
Thanks - that really tidies it up!

PR #1431, opened 2018-05-20
Add shiftOut clkPinsClk option

opichals:
Allowing e.g. the [tm1640.js:send()](https://github.com/espruino/EspruinoDocs/blob/0a349f329c566b0147377737cc54704aa61294f5/devices/TM1640.js#L24) function to be replaced with a single `shiftOut()` call.

### DIFF
--- a/src/jswrap_io.c
+++ b/src/jswrap_io.c
@@ -447,6 +447,7 @@ typedef struct {
   volatile uint32_t *clkAddr;
 #endif
   bool clkPol; // clock polarity
+  bool clkPinsClk; // clock pin clock ordering
 
   int cnt; // number of pins
   int repeat; // iterations to perform per array item
@@ -456,6 +457,15 @@ void jswrap_io_shiftOutCallback(int val, void *data) {
   jswrap_io_shiftOutData *d = (jswrap_io_shiftOutData*)data;
   int n, i;
   for (i=0;i<d->repeat;i++) {
+#ifdef STM32
+    if (d->clkPinsClk && d->clkAddr) {
+      *d->clkAddr = d->clkPol;
+    }
+#else
+    if (d->clkPinsClk && jshIsPinValid(d->clk)) {
+      jshPinSetValue(d->clk, d->clkPol);
+    }
+#endif
     for (n=d->cnt-1; n>=0; n--) {
   #ifdef STM32
       if (d->addrs[n])
@@ -468,12 +478,12 @@ void jswrap_io_shiftOutCallback(int val, void *data) {
     }
 #ifdef STM32
     if (d->clkAddr) {
-        *d->clkAddr = d->clkPol;
+        if (!d->clkPinsClk) *d->clkAddr = d->clkPol;
         *d->clkAddr = !d->clkPol;
     }
 #else
     if (jshIsPinValid(d->clk)) {
-      jshPinSetValue(d->clk, d->clkPol);
+      if (!d->clkPinsClk) jshPinSetValue(d->clk, d->clkPol);
       jshPinSetValue(d->clk, !d->clkPol);
     }
 #endif
@@ -514,6 +524,7 @@ shiftOut([A3,A2,A1,A0], { clk : A4 }, [1,2,3,4]);
 {
   clk : pin, // a pin to use as the clock (undefined = no pin)
   clkPol : bool, // clock polarity - default is 0 (so 1 normally, pulsing to 0 to clock data in)
+  clkPinsClk : bool, // clock pin ordering - default is 0 (normally the order is pins, clk, clk, if 1 the order is clk, pins, clk)
   repeat : int, // number of clocks per array item
 }
 ```
@@ -531,11 +542,13 @@ void jswrap_io_shiftOut(JsVar *pins, JsVar *options, JsVar *data) {
   d.cnt = 0;
   d.clk = PIN_UNDEFINED;
   d.clkPol = 0;
+  d.clkPinsClk = 0;
   d.repeat = 1;
 
   jsvConfigObject configs[] = {
       {"clk", JSV_PIN, &d.clk},
       {"clkPol", JSV_BOOLEAN, &d.clkPol},
+      {"clkPinsClk", JSV_BOOLEAN, &d.clkPinsClk},
       {"repeat", JSV_INTEGER, &d.repeat}
   };
   if (!jsvReadConfigObject(options, configs, sizeof(configs) / sizeof(jsvConfigObject))) {


gfwilliams:
I don't really understand what's going on here - does just flipping the clock polarity not do what you want?

opichals:
Yes, the current shiftOut() implementation didn't get me anything displayed.

opichals:
Arduino libs actually do clk, data, clk as well.

E.g. here https://github.com/ghalfacree/Arduino-Sketches/blob/master/libraries/TM1638/TM16XX.cpp#L138

gfwilliams:
Except for the actual `shiftOut` function in Arduino's API:

http://forum.arduino.cc/index.php?topic=45529.0

```
void shiftOut(uint8_t dataPin, uint8_t clockPin, uint8_t bitOrder, uint8_t val)
{
     uint8_t i;

     for (i = 0; i < 8; i++)  {
           if (bitOrder == LSBFIRST)
                 digitalWrite(dataPin, !!(val & (1 << i)));
           else      
                 digitalWrite(dataPin, !!(val & (1 << (7 - i))));
                 
           digitalWrite(clockPin, HIGH);
           digitalWrite(clockPin, LOW);            
     }
}
```

Honestly I'm struggling with this. As far as the hardware is concerned the data should be changed as soon as possible.

What actually happens when the TM1640 doesn't work? Do you get the wrong data, or just just flat-out refuses to do anything?

What about a second time when the clock has already been set to the right polarity.

opichals:
Yes, Arduino's shiftOut() does the same thing as Espruino's does. But that's probably the reason the Arduino libs contain the send() method itself and are not using the `shiftOut()`.

> What actually happens when the TM1640 doesn't work? Do you get the wrong data, or just just flat-out
> refuses to do anything?

It doesn't display anything. No activity.

> What about a second time when the clock has already been set to the right polarity.

Nothing I could see from here.

opichals:
I tried sw SPI, but no improvement.

Have a look at this post pinpointing the fact it is not SPI. http://dangerousprototypes.com/forum/viewtopic.php?f=4&t=4450#p43661

gfwilliams:
Thanks - this does just seem horribly complex though - I mean it's got its own little parser. In an ideal world we'd be using the JS parser that's part of Espruino for that and then it'd be infinitely flexible!

I just don't see this getting used apart from in this one case.

What about:

```
var DATA = D0;
var SCLK = D1;

var s = new SPI();
s.setup({mosi:DATA, sck:SCLK});

console.log("Start Output");
DATA.reset();
SCLK.reset();
s.write(0b10101010);
DATA.reset();
SCLK.set();
DATA.set();
```

I tested that and it would appear to output the data in exactly the form that the post above mentioned.

Not only that, but you can use the fact that `digitalWrite` outputs everything in LSB order (http://www.espruino.com/Reference#l__global_digitalWrite):

```
digitalWrite([SCLK,DATA],0b00);
s.write(0b10101010);
digitalWrite([DATA,SCLK,DATA],0b110);
```

You can pre-bind those as well for even more speed.

opichals:
@gfwilliams I struggle to see how the SPI.write would generate the required pattern.

I tried to draw it in wevadrom [here](https://wavedrom.com/editor.html?%7B%20signal%3A%20%5B%0A%20%20%7B%20name%3A%20%27CLK%27%2C%20%20wave%3A%20%270..1.0...1.0...1.0..%27%20%20%7D%2C%0A%20%20%7B%20name%3A%20%27MOSI%27%2C%20wave%3A%20%270.1...0.......1...0.%27%2C%20%7D%2C%0A%20%20%7B%7D%2C%0A%20%20%7B%20name%3A%20%27DIN%27%2C%20%20wave%3A%20%270...1...0.......1...0.%27%2C%20%7D%2C%0A%5D%2C%0A%7D%0A) where the CLK & MOSI are what I imagine is SPI, and the DIN what I think I need with respect to the CLK signal.

I am a complete newbie in this field so bare with me please.

gfwilliams:
Looking at your code: https://github.com/espruino/EspruinoDocs/blob/0a349f329c566b0147377737cc54704aa61294f5/devices/TM1640.js#L24

`c` and `d` are normally high, and when you call sendCommand the waveform looks like:

```
d(0)
// i=0
c(0)
d(bit0)
c(1)
// i=1
c(0)
d(bit1)
c(1)
// ...
// i=7
c(0)
d(bit7)
c(1)
// loop ends
d(1)
```

You're actually not doing what's mentioned at that datasheet diagram at http://dangerousprototypes.com/forum/viewtopic.php?f=4&t=4450#p43661

Since you're only sending 8 rising clock edges, not 9. Maybe that's a red herring though.

But you're looking at it and thinking it does this:

* Sets the state of some pins
* Loops, setting clock to 0, setting data, and setting it to 1
* Sets the state of some pins

But that's kind of the same as what you can do with SPI:

* Sets the state of some pins
* Loop, setting data, then setting clock to 0 then to 1
* Sets the state of some pins

It's just a matter of what goes on in the start and end. I'm pretty sure they're saying in the link that you can't use *hardware SPI* because you need control of the pins for the start and end state, and that's about right - but we're using software SPI so we can do whatever we want with the pins.

Anyway, I'm pretty sure the code I posted would work - although you need `order:'lsb'` in `s.setup` since the bit order is different.

Or even if you don't want to do that, this code is 3 times faster:

```
var C = pins.clk.set.bind(pins.clk);
var c = pins.clk.reset.bind(pins.clk);
var D = pins.din.set.bind(pins.din);
var d = pins.din.reset.bind(pins.din);
var o = pins.din.write.bind(pins.din);

function send(x) {
  c();d(x&1);C();c();o(x&2);C();c();
  d(x&4);C();c();o(x&8);C();c();
  d(x&16);C();c();o(x&32);C();c();
  d(x&64);C();c();o(x&128);C();
}

  g.flip = function() {
    var b = g.buffer;
    for(var i=0;i<8;i++) {
      d();send(0x44);D();
      d();send(0xC0|i);
      send(b[i]);D();
      d();c();C();D();
    }
    d();send(0x88|intensity);D();
  };
```

opichals:
@gfwilliams Thank you for your patience with me.

I actually went and read what the SPI.write method does internally and figured this to work the way the tm1640 requires https://github.com/espruino/EspruinoDocs/pull/436

gfwilliams:
No problem at all - it was interesting to have a play with it. Thanks for taking the time to make the TM1640 driver better :)

Just so you know, `pins.din.set.bind(pins.din);` will definitely be faster/more efficient than `digitalWrite.bind(null, pins.din, 1);` - maybe not much, but a little :)

opichals:
> `pins.din.set.bind(pins.din);` will definitely be faster/more efficient

Sure, will create a PR for that too. :)

PR #1447, opened 2018-05-25
Add shiftOut pattern option

opichals:
A more versatile alternative proposal to https://github.com/espruino/Espruino/pull/1431. The clock/pin ordering can be programmed via the added `pattern` option as follows:

```
*0 // set clk to 0
*1 // set clk to 1
-0 // set pins[0] to val & 1
-4 // set pins[4] to val & 1
```

Variant examples:
```
// 1 bit, single data pin+clk
shiftOut(D0, { clk : A1, pattern: "-0*0*1" }, [1,0,1,0]);

// 4 bits, LSB via 4 data pins
shiftOut([D0, D1, D2, D3], {clk: A1, pattern: "-3-2-1-0*0*1"}, [1, 2, 3, 4]);

// 4 bit, MSB via 4 data pins
shiftOut([D0, D1, D2, D3], {clk: A1, pattern: "-0-1-2-3*0*1"}, [1, 2, 3, 4]);

// a whole byte, single data pin (like software SPI)
shiftOut(D0, {clk: A1, pattern: new Array(8).fill("-0*0*1").join('')}, [1, 2, 3, 4]);

// 8bit, tm1640 pattern (clk,data,clk)
shiftOut(D0, {clk: A1, pattern: new Array(8).fill("*0-0*1").join('')}, [1, 2, 3, 4]);
```

### DIFF
--- a/src/jswrap_io.c
+++ b/src/jswrap_io.c
@@ -447,6 +447,7 @@ typedef struct {
   volatile uint32_t *clkAddr;
 #endif
   bool clkPol; // clock polarity
+  char pattern[64]; // pin ordering pattern
 
   int cnt; // number of pins
   int repeat; // iterations to perform per array item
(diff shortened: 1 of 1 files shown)


opichals:
No need for this, sw SPI works as suggested by @gfwilliams. https://github.com/espruino/EspruinoDocs/pull/436

Issue #1432, opened 2018-05-20
Missing require('Storage').readStream('file')


opichals:
It'd be nice way to stream data out in WebServer via `pipe()` call.

Also this can allow processing larger than 65kB contents (as strings are limited to this).

A first step may be adding `offset`+`len` support to `Storage.read()` i.e. `jsfReadFile()`

opichals:
> With the 65k limit, I think we can actually get away without that on all but the most confined (<1k vars) platforms. I've just tweaked it.

I don't understand. How can the data beyond the 65k limit /that can be written by require('Storage').write('file', "test", 66000, 100)/ be accessed?

opichals:
And yes, the JS module is completely fine for this, not sure what would be more space-saving.

gfwilliams:
Closing this as you can now do `E.pipe(require('Storage').read(...), dst)` and this should work fine

PR #1453, opened 2018-05-29
Add E.pipe as fs.pipe alternative

opichals:
For boards not having the 'fs' module it is still useful to have the built-in pipe() functionality.

They could stream for example Storage contents via https://github.com/espruino/Espruino/issues/1432

### DIFF
--- a/src/jswrap_espruino.c
+++ b/src/jswrap_espruino.c
@@ -720,6 +720,19 @@ Run `E.getFlags()` and check its description for a list of available flags and t
 /*JSON{
   "type" : "staticmethod",
   "class" : "E",
+  "name" : "pipe",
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks! Which boards are we talking about that don't have SAVE_ON_FLASH or filesystem defined? It's just the ESP8266 512k?

It just seems a bit strange, because if you have the flash available then you'd have filesystem, but if you don't have the flash then you probably can't store anything big enough to warrant using pipe?

opichals:
For the ESP8266 (even for the 4MB flash version) the filesystem doesn't make sense because of lack of RAM, not flash.

gfwilliams:
Ahh, interesting - thanks.

PR #1456, opened 2018-06-04
Ble/wifi switchable - jsvars calculated 

wilberforce:
We might as well merge this into master now...


wilberforce:
> Please could you check? I don't want to merge if it's got a known buffer overflow in it :)

I had code that was saved with a normal `save()`. When saved with `save on send` in the IDE it has come up fine - so there does not appears to be an issue.

PR #1469, opened 2018-06-13
Add docs on how to get the list of built-in modules

### LINKED ISSUE #1464: Any way to get the list of built-in modules?
`Modules.getCached()` only returns modules that were already required.
Is there a way to get all built-in modules that I can `require()`?

vshymanskyy:
Fix #1464


gfwilliams:
Nice - thanks!

PR #1471, opened 2018-06-16
Disable TIME_WAIT socket state for Linux

opichals:
Linux sockets to set SO_LINGER sockopt to 0.

This allowes running the network tests immediately
one after another without blocking the listening port
after process exit.


gfwilliams:
That's great - thanks! I always wondered why they failed when run together.

opichals:
@wilberforce Not sure what are you after exactly. The PR added it to [network_linux.c:199](https://github.com/espruino/Espruino/blob/1ceb47bb562b8890d6a4e25468b6530715781433/libs/network/linux/network_linux.c#L199).

wilberforce:
So #define SO_LINGER must get set in a header  as part of the linux network code.
 I was expecting it to defined in the Esprunio code somewhere to turn this on.

opichals:
@wilberforce It is a standard http://man7.org/linux/man-pages/man7/socket.7.html thing.

PR #1472, opened 2018-06-17
Add dgram.createSocket({recvBufferSize: 1024}) option

opichals:
Introduced JsNetwork->rcvBufferSize updated at socket
creation/closure time

### DIFF
--- a/libs/network/esp32/network_esp32.c
+++ b/libs/network/esp32/network_esp32.c
@@ -71,7 +71,6 @@ bool net_esp32_checkError(JsNetwork *net) {
 
 /// if host=0, creates a server otherwise creates a client (and automatically connects). Returns >=0 on success
 int net_esp32_createsocket(JsNetwork *net, SocketType socketType, uint32_t host, unsigned short port, JsVar *options) {
-  NOT_USED(net);
   int ippProto = socketType & ST_UDP ? IPPROTO_UDP : IPPROTO_TCP;
   int scktType = socketType & ST_UDP ? SOCK_DGRAM : SOCK_STREAM;
   int sckt = -1;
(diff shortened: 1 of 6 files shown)


opichals:
@gfwilliams Currently any board is limited to only be able to receive UDP packets of max size of `chunkSize` and it is due to the fact that `recvfrom()` simply drops the bytes that don't fit the provided buffer for UDP sockets.

> Could you not just have a if (bufSize>chunkSize) chunkSize=bufSize in dgram.createsocket?

So that it would increase the `netRecv()` globally for all connections? I guess I could live with something as simple as that.

opichals:
@gfwilliams What about now?

gfwilliams:
Thanks - that looks good. Is there a reason you do `jsvGetIntegerAndUnLock(jsvObjectGetChild(options, "recvBufferSize", 0))` rather than using `net->data.recvBufferSize` in the linux/esp32 network files?

Sorry, I know it seems I'm being needlessly picky about this. Pretty much every release recently I've had to spend ages trying to find ways to reduce the build size for the Original Espruino board and some others enough to get them building. I've spent a long time trying to compact code and now I'm starting to have to pull out features for every release - so now there's this balance of trying to add features people will use while not removing other things that actually are being used.

opichals:
> using net->data.recvBufferSize in the linux/esp32 network files

No reason, I had the `options` based pattern there already so I used it. I am going to amend the PR as you suggest (it is less work and shorter compiled code).

No worries, I understand the size battle. It would be nice to put those functionality reductions to the ChangeLog explicitly for everyone to spot immediately.

opichals:
@gfwilliams adjusted the code to use `net->data.recvBufferSize`

gfwilliams:
Great - thanks!

PR #1493, opened 2018-08-29
ESP332: update buildtools sdk 3.03, set CONFIG_ESP32_XTAL_FREQ_AUTO

wilberforce:


gfwilliams:
Seems it isn't finding the libraries now?

PR #1505, opened 2018-09-10
fix warning: array subscript is above array bounds

MaBecker:
src/jsdevices.c:93:27: warning: array subscript is above array bounds [-Warray-bounds]
         jshSerialDeviceCTSPins[i] = PIN_UNDEFINED;
                                ^


gfwilliams:
Good spot!

PR #1520, opened 2018-10-08
UI adjustement for overlay legends on board view

stephaneAG:
Prevents overlapping of legends for boards that have many functions on the same pins.
This fix allows the viewer to hover parts he wishes to see:
- not hovering: dispay everything
- hovering #board: display board without overlays
- hovering a pin row: overlay only the legends corresponding to that row of pins

### DIFF
--- a/boards/STM32F3DISCOVERY.py
+++ b/boards/STM32F3DISCOVERY.py
@@ -134,6 +134,15 @@ board["_css"] = """
   top: 320px;
   left: 530px;
 }
+/* prevent pins overlay legends from overlapping when hovering */
+#boardcontainer:hover #left { opacity: 0.1; }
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks!

stephaneAG:
my pleasure ;)

nb: I'm currently re-digging the stm32f4discovery pinouts & it seems the merged version isn't updated on Espruino.com/Reference.. wonder why ? ..

cheers ++

gfwilliams:
Ahh, I just looked into this. Very few people were using the STM32F3DISCOVERY port so I took it out of the distribution zip for 2.00 since it takes me time to maintain and I don't think it's very stable, but that means that the board docs don't get built for it automatically any more.

I've just tweaked the scripts so the docs for the F3 do get rebuilt, even though it's not in the distribution zip.

PR #1521, opened 2018-10-12
remove  SHA256 SHA512 (fix #1517)

### LINKED ISSUE #1517: Was there a reason for adding SHA256 and SHA512 on ESP8266_4MB
As far as I know there is only module `ws` that is using sha1? Or do I miss some modules or functions 
that use SHA256 or even SHA512?

This is all about remove not needed stuff to have space for important things.

Like to remove it if not needed.

MaBecker:


MaBecker:
Thanks

PR #1522, opened 2018-10-12
release heap used by logDebug(true) (fix #1508)

### LINKED ISSUE #1508: release heap used by logDebug(true)
Espruino starts by default with logDebug(true) on ESPs which consumes about 1.2k of heap.

So if RELEASE=1 lets call `jswrap_ESP8266_logDebug(false);`  in `initDone()`.

MaBecker:


MaBecker:
Thanks

PR #1526, opened 2018-10-15
All server net.Socket are set to SO_REUSEADDR

opichals:
Same as in node.js https://nodejs.org/api/net.html#net_server_listen

### DIFF
--- a/libs/network/http/jswrap_http.c
+++ b/libs/network/http/jswrap_http.c
@@ -238,7 +238,12 @@ JsVar *jswrap_http_createServer(JsVar *callback) {
     return 0;
   }
   jsvUnLock(skippedCallback);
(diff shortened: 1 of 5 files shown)


MaBecker:
wow, does this fix the socket problem?

opichals:
It appears so. The --test-all works properly. At least on MacOS for me here.

wilberforce:
Sorry - I'm missing something - where has `SO_REUSEADDR` been added to the code?

I see there is a new options parameter - do you need to set somehow with this?

opichals:
ESP8266 doesn't have that option in its sdk I think.

opichals:
Closing this in favor https://github.com/espruino/Espruino/pull/1527 with minimal changes.

PR #1527, opened 2018-10-16
net.Server to set SO_REUSEADDR by default

opichals:
Same as in node.js https://nodejs.org/api/net.html#net_server_listen

i.e. non-UDP servers

### DIFF
--- a/libs/network/esp32/network_esp32.c
+++ b/libs/network/esp32/network_esp32.c
@@ -125,7 +125,8 @@ int net_esp32_createsocket(JsNetwork *net, SocketType socketType, uint32_t host,
       return 0;
     }
 
-    if (jsvGetBoolAndUnLock(jsvObjectGetChild(options, "reuseAddr", 0))) {
+    if (scktType != SOCK_DGRAM ||
+        jsvGetBoolAndUnLock(jsvObjectGetChild(options, "reuseAddr", 0))) {
       int optval = 1;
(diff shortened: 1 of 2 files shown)


gfwilliams:
Thanks - and we're sure that this doesn't break HTTP/TCP/UDP servers?

Last time --test-all worked but it seems that running any TCP server on its own was broken.

gfwilliams:
I believe there are some tests in --test-all, but yeah - they just use Espruino, and a single instance at that.

So you're saying you haven't tested this? Having just spent yesterday tracking down a bug from another of your commits that wasn't tested, I'm not about to merge this one unless I know server/etc still works on Linux and ESP32.

opichals:
I am using espruino on MacOS and there I saw no problems even with the SO_LINGER in there. So for MacOS I tested it being able to run a web server and able to do websocket as well.

With no automatic PR tests this anxiety will get worse the bigger community there is...

gfwilliams:
Ok, looks promising on Linux. No idea about ESP32 though. Did you test that?

I do have some plans for testing each commit on actual hardware, but finding time to implement that is difficult and I think it would be amazingly bad for Espruino if people just committed stuff without testing and used the test harness as a way to see if it was ok.

opichals:
I went for consistency of the server.listen() call to set the reuseAddr for non-UDP servers. The fact is that I haven't tested it physically on ESP32.

> if people just committed stuff without testing and used the test harness as a way to see if it was ok

I would not worry about it too much till there is such a thing. I'd say HW testlab utilization could be controlled by you in the PR by say adding a tag or something.

For the rest of the things travis is cool IMO.

PR #1531, opened 2018-10-19
remove getBuildHash and add mqtt library

iotong:
remove getBuildHash and add mqtt library

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,4 +53,6 @@ targetlibs/nrf5x_mesh
 /CURRENT_BOARD.make
 /topreadonly
 /topstrings
-
+/sdk
+/projects
+*.lod
--- a/Makefile
+++ b/Makefile
@@ -511,6 +511,12 @@ ifdef USE_NET
   WRAPPERSOURCES += libs/network/telnet/jswrap_telnet.c
   INCLUDE += -I$(ROOT)/libs/network/telnet
  endif
+
+ ifdef USE_MQTT
+  DEFINES += -DUSE_MQTT
+  WRAPPERSOURCES += libs/network/mqtt/jswrap_mqtt.c
+  INCLUDE += -I$(ROOT)/libs/network/mqtt
+ endif
 endif # USE_NET
 
 ifdef USE_TV
--- a/boards/ESP8266_4MB.py
+++ b/boards/ESP8266_4MB.py
@@ -28,6 +28,7 @@ info = {
      'GRAPHICS',
      'CRYPTO',
      'NEOPIXEL',
+     'MQTT'
      #'FILESYSTEM',
      #'FLASHFS'
    ],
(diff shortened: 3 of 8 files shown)


gfwilliams:
Thanks for this - but just wondering:

* Why was getBuildHash removed? It was added for a good reason - any code saved with 'save()' in an older version of Espruino won't work
* Why do you want to build mqtt in? What is wrong with the JS version of it that can't be fixed by changing the JS? Many Espruino boards just dont have the available flash to have this built in all the time.

iotong:
1. The old version of the code can't run because the native function reference has an error. So just modify the pointer to the native function to solve the problem! 2. Create native mqtt because it will consume less variable

gfwilliams:
If the JS code does something like `a=print; setInterval('print("Hello")',100)` then your changes will just make the board crash as soon as it loads the code. I don't think that's an improvement.

I can see you might want to add a parameter to `load()` to *force* it to load saved code from an older version, but just making the board load by default is a bad idea.

Sorry, but I'm not willing to merge this - you could just add `JSMODULESOURCES += libs/js/mqtt.min.js` to the `BUILD.py` file and include the JS which will instantly save loads of variables. Even if your library would fit on all the Espruino boards (which it won't at the moment), it appears to have at least one memory leak when saving, the API is subtly different to the existing MQTT library, and the code doesn't appear to be able to cope with MQTT commands that are spread across multiple network packets, which the JS version does - it would just end up being a complete pain to maintain.

gfwilliams:
If the JS code does something like `a=print; setInterval('print("Hello")',100)` then your changes will just make the board crash as soon as it loads the code. I don't think that's an improvement.

I can see you might want to add a parameter to `load()` to *force* it to load saved code from an older version, but just making the board load by default is a bad idea.

Sorry, but I'm not willing to merge this - you could just add `JSMODULESOURCES += libs/js/mqtt.min.js` to the `BUILD.py` file and include the JS which will instantly save loads of variables. Even if your library would fit on all the Espruino boards (which it won't at the moment), it appears to have at least one memory leak when saving, the API is subtly different to the existing MQTT library, and the code doesn't appear to be able to cope with MQTT commands that are spread across multiple network packets, which the JS version does - it would just end up being a complete pain to maintain.

gfwilliams:
If the JS code does something like `a=print; setInterval('print("Hello")',100)` then your changes will just make the board crash as soon as it loads the code. I don't think that's an improvement.

I can see you might want to add a parameter to `load()` to *force* it to load saved code from an older version, but just making the board load by default is a bad idea.

Sorry, but I'm not willing to merge this - you could just add `JSMODULESOURCES += libs/js/mqtt.min.js` to the `BUILD.py` file and include the JS which will instantly save loads of variables. Even if your library would fit on all the Espruino boards (which it won't at the moment), it appears to have at least one memory leak when saving, the API is subtly different to the existing MQTT library, and the code doesn't appear to be able to cope with MQTT commands that are spread across multiple network packets, which the JS version does - it would just end up being a complete pain to maintain.

PR #1541, opened 2018-10-25
update micro-seconds when updating time

jbgagnon:
(first pull request: not sure what i am doing)

only 1 commit must be taken 
targets/esp32/jshardware.c


gfwilliams:
Neat - thanks!

PR #1542, opened 2018-10-25
fix types lacks a cast

MaBecker:
warning: comparison of distinct pointer types lacks a cast [enabled by default]


gfwilliams:
Thanks!

PR #1543, opened 2018-10-26
Dockerfile that just works

Fxlr8:
Hi, I updated your Dockerfile to make it work. It is far from perfect right now, but it works.
Here are some ways to improve it:
1. Transfer all logic from provision.sh so all the dependencies are handled by docker. This will let docker's cache work and speed up the container build
2. Mount source code and result folders to the container, so you don't need to rebuild it every time you change the code. Also you will not need to copy the results from the container every time.
3. Pass the board name as an env variable to the container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,21 @@
-FROM ubuntu:14.04
+# Usage:
+# 1: Buld the container image 
+# docker build . -t img_name
+# 2: Run container image so it builds espruino
+# docker run img_name --name container_name
+# this will run the container and save build results into the container's filesystem
+# 3: Copy build results from the container into your filesystem
+# docker cp container_name:espruino/espruino_1v99.4185_stm32f100lqfp48.elf ./
 
-RUN git clone https://github.com/espruino/Espruino espruino
+FROM python:3
+
+COPY . /espruino
 WORKDIR /espruino
 
 # Change these to provision and compile for a different board
-RUN source scripts/provision.sh PICO_R1_3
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks - but does `FROM python:3` really pull in everything you need? How does it know to pull in a version of Linux for the build?

Nice ideas for improving it - although I'm totally against pulling in `provision.sh` just for a bit of speed. Very few people seem to have bothered maintaining the Docker file thus far - so if we did move the code there's no way it'd stay working past a few weeks :(

Since we used `ENV BOARD`, can the rest of the commands be made to use `$BOARD`?

Fxlr8:
When i first tried to run your Dockerfile the first error i got was `python`

gfwilliams:
Ahh, interesting - it looks from `Shared Tags` like `python:3` may actually use different OS images on some platforms? I guess we'd have to use `3-stretch`.

Very odd about Ubuntu images though - pretty sure Ubuntu always comes with Python by default (it did when the Docker file was made by the look of it), but maybe they've changed it now and stripped it right back for Docker.

You can't get `provision.sh` to install everything right now, but yeah, it'd be nice to allow `provision.sh ALL` - and not too painful to do inside `provision.sh`. Nothing should conflict - however it would make it quite a bit slower to set up the image. We could just run `./provision` several times for the ESP8266/32 and ARM - but it's not great for maintainability :)

Ideally people would just be able to point Docker at github without ever having to clone the Repo to change the Docker file -  if it enabled that then it'd be well worth it.

PR #1548, opened 2018-11-03
ESP8266 cleanup

MaBecker:


MaBecker:
close because it contains progress work

PR #1550, opened 2018-11-12
ESP8266 deepSleep invalid microseconds (fix #1547)

### LINKED ISSUE #1547: ESP8266 deepSleep invalid microseconds
Its not possbile to set deepsleep for a long time
for example this 10 digit number returns an error:
6000000000
while 9 digits works

since deepsleep is in micro seconds it really limits the range.
```
>ESP8266.deepSleep(9071992547)
Uncaught Error: Invalid microseconds.
 at line 1 col 29
ESP8266.deepSleep(9071992547)
                            ^
>ESP8266.deepSleep(907199254)
=undefine
```

MaBecker:
will update ChangeLog later


gfwilliams:
Thanks :)

PR #1561, opened 2018-11-14
fix error String.length value for native string

hellprototypes:
If a native string size greater than 65535B, the String.length was wrong.


gfwilliams:
Thanks. Looks like this may cause warnings on some builds (with less than 1024 vars) but that's not such a big deal.

PR #1566, opened 2018-11-14
add ESP8266 (fix #1563)

### LINKED ISSUE #1563: Missing variable Flash Mode (--flash_mode, -fm) in ESP8266.make files
Flash mode set fixed to qio, what works for many boards, but not for all.

Some vendors of ESP-boards decided to change this from qio to dio. 

So lets add EF_FM to ESP8266.make file in family and targets to allow:

    ET_FM=dio ESP_FLASH make flash

MaBecker:


gfwilliams:
In the future, is it possible to try and merge these 3 PRs into one? They all seem to be related

MaBecker:
Sure, that's what I do next time.

PR #1569, opened 2018-11-15
Remove HACK to get around SDK 1.4 bug (fix #1568)

### LINKED ISSUE #1568: Remove HACK to get around SDK 1.4 bug
There is still some [old code alive](https://github.com/espruino/Espruino/blob/master/targets/esp8266/user_main.c#L229-L231) which has been set for SDK 1.4

MaBecker:


gfwilliams:
Thanks :)

PR #1575, opened 2018-11-29
Remove insecure (redundant) jQuery reference

IrregularShed:
I've been getting a 'Mixed Content' error in the Chrome console where jQuery was trying to be loaded over HTTP when the page was loaded over HTTPS. jQuery is already getting loaded on espruino.com pages from a local copy - there's two references in fact, one in `<head>` and the other at the bottom of the page - so we can get rid of this part all together. (I couldn't find the page template that is doubling-up the jQuery request but I'd expect the browser cache to handle that well enough.)

### DIFF
--- a/scripts/build_board_docs.py
+++ b/scripts/build_board_docs.py
@@ -265,7 +265,7 @@ writeHTML("""  <STYLE>
 }
 
 """);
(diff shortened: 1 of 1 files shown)


gfwilliams:
Neat - thanks! I imagine this was done so files could be viewed on their own, but honestly I'm not sure there's much jQuery in there anyway, and if there is we should be using normal browser stuff now.

Issue #1489, opened 2018-08-10
small numbers a convert to zero if smaller than 1e-9, big numbers are strip to 1e68 when convert to string


jbgagnon:
so we can't do any physics ...

to test:
console.log(1e-9) write 0
'x:" + 1e-9 write "x:0"
JSON.stringify({x:1e-9}) return ="{\"a\":0}"
Math.pow(2,-32) says 0
('' + 1e72).length says 69
JSON.stringify({x:1e80}) is also wrong so can't be communicate to an external service
console.log(1e100/3) write something not enough long ( or two long 3.333333333333333e+99 is the good answer)

the bug is in
file: jsutils.c
function: ftoa_bounded_extra

it must test if exponent must be added for example if 
val < 1e-9 or val > 1e64 (I don't really know the actual limits)
but double value have  17 chars of precision so probably test val > 1e17 (it also depends of buffer length, normaly a buffer of 32 chars is good (sign + dot + 17 chars + e + sign + exponent max 3 char  = 24 chars)

bug probably or some platform (esp32, esp8266, ...) we can use native library and smaller this code ... no ?

gfwilliams:
I think parsing floating point numbers works just fine. `0==1e-20` is `false`. It's just converting to Strings.

Custom double-to-string code was originally added because platform-specific versions pulled in large amounts of extra code - on some platforms it caused 1kB of RAM to be statically allocated. I believe JS may have some standards on how double to string code should behave as well (not that we currently adhere to those).

So I'd like to keep the implementation inside Espruino if only because it aids testing to always use the same implementation - but you're welcome to contribute an updated `ftoa_bounded_extra` function that'll use `e` for larger/smaller values if you want to.

jbgagnon:
hello Gordon,
I go en hollidays tomorrow. But when i come back at end of august i will be
happy to contribute to this project.
JB Gagnon


2018-08-10 15:49 GMT+02:00 Gordon Williams <notifications@github.com>:

gfwilliams:
Any chance of doing this?

jbgagnon:
Ok I can compile now !
(making some change in provision.sh for path to access get_board_info.py,
and do a pacman for tar command ....)

so to train myself I correct this issue (the easy one)
https://github.com/espruino/Espruino/issues/1488

I test it on my board, it's ok ! no "back to the future" at all

what is the good way to proceed (is-there a guide ?)
must a need to create a branch ?
must somebody make a code review before commiting ?
[...]

PR #1577, opened 2018-11-30
print big numbers

jbgagnon:
take only  src/jsutils.c
it is only for big numbers.
not small numbers because small numbers are round near integer so small number are always round to zero ...

this is for #1489

sorry for the delay


jbgagnon:
Is there any big endian module ?
Because, I have to make some changes if big endian is supported
(just a line two change if someone give me the good #ifdef )

gfwilliams:
Thanks for removing the other files. If you're distributing your builds to others, the removal of the "maybe donate something" text is a bit unreasonable though.

Is it possible to do a simpler version without the bitwise math? It's cool, but if Espruino were on big endian systems it'd be an annoying bug.

Also on some of these boards (ESP8266 512k, some of the ARM boards) I'm really struggling with flash memory usage, and I feel like reimplementing pow10 and doing bit hacks rather than just doing `log` and `jswrap_math_pow(10,...)` is just increasing the code side when I imagine if you're converting massive double numbers into strings it's kindof unlikely that speed is of the essence?

jbgagnon:
for big endian it's just a line replacing
       int p2 = (((unsigned short *)& val)[3] << 1 >> 5) - 0x3ff;
by
       int p2 = (((unsigned short *)& val)[0] << 1 >> 5) - 0x3ff;
(I think)
but it need to be tested

you can also use my pow10int function for replacing at line 549 comment

for testing 
for(var i=0,j=1;i<=308;i++) { console.log(i, j); j *=10; }
for(var i=0,j=10;i<=308;i++) { console.log(i, j/7); j *=10; }
is a start

jbgagnon:
sorry for the "donate ..." it's all what  I find to make difference from my version of the original  one.
Certainly stupid 
and I discovering git and comminting bad  things ...

gfwilliams:
Sorry for the delay - closing this finally - just pulled this in manually with some changes to handle small numbers too

PR #1581, opened 2018-12-06
Invert TX/RX pins for Quectel BC95-G

conalllaverty:


gfwilliams:
Does changing this actually fix your problems? It's come up before and I'm pretty sure these are correct.

conalllaverty:
Yes. We've just tried on an RAK8211-NB board. Using the following snippet:
```
var buff = "";
var iTracker = require("iTracker");

var nbiot_serial = function(){
 buff = "";
 Serial1.removeAllListeners();
 Serial1.setup(9600, { tx:D20, rx:D12 });
 Serial1.on('data', function (data) {
  buff = buff + data;
   var OK = buff.includes("\nOK");
   var ERROR = buff.includes("\nERROR");
   if (OK === true || ERROR === true) {
     console.log(buff);
     buff = "";
   }
 });
};

iTracker.setCellOn(true, function(usart) {
 console.log("cell now on");
 nbiot_serial();

 Serial1.println("ATI");

});
```

gfwilliams:
Thanks - I'm just waiting to hear back from Rak Wireless about this. I'm pretty convinced the 8211-G does have the pins as they were before, so adding this PR would break that one while making the 8211-NB work.

It may be that I end up adding a `setNBCellOn` function or something like that.

conalllaverty:
Cool. An additional function like `setNBCellOn` should work. Happy to test any changes you make around this.

PR #1585, opened 2018-12-16
ESP32: save/restore wifi config to Storage. Add set/getHostname, setSNTP

wilberforce:
Implement all the remaining fuctions to bring inline with ESP8266 port


wilberforce:
@gfwilliams 
Not sure why some of the push checks failed - when the same pr checks passed?

Please merge!

gfwilliams:
Thanks! Sometimes Travis builds do just randomly fail (usually connection issues) so it's nothing to worry about.

PR #1586, opened 2018-12-18
WIP: a better dockerfile

Fxlr8:
Hi, there is still a lot of improvements to be made with the Dockerfile. Currently, it is not cache-friendly and container image rebuild takes a lot of time. We should run the `provision.sh` script first and then copy all the source code into the container. This way docker will only re-run the last steps of the Dockerfile and save a lot of time. 

I tried to do this, and also I changed the last line to
```Dockerfile
CMD "make"
```
instead of
```Dockerfile
CMD ["bash", "-c", "source scripts/provision.sh ALL && make"]
```

and when I try to run the container I get this error:
```
make: arm-none-eabi-gcc: Command not found
```

Why is that happening? `arm-none-eabi-gcc` should already be installed during the build of a docker image, isn't it?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,8 +22,8 @@
 
 FROM python:3
 
-COPY . /espruino
-WORKDIR /espruino
+COPY ./scripts/provision.sh provision.sh
+COPY ./targetlibs /targetlibs
 
 # Workaround add some stuff that the provision script uses
 # in here so it doesn't have to use sudo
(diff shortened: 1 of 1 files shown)


gfwilliams:
`scripts/provision.sh` installs all the tools required, but ALSO sets up environment variables - hence having to be called via `source`. I wonder if there's much of a way around it.

Another option might be to detect is the `BOARD` env variable was set and to just to `provision BOARD` if it was?

Fxlr8:
Maybe a noob question, but is it possible to have all the gcc's installed at once? To make the image more cache-friendly we have to install the least changing things first.

gfwilliams:
>  is it possible to have all the gcc's installed at once? 

Yes! that's what I spent hours on doing with this current Docker image that uses `source scripts/provision.sh ALL` and I guess why you're finding the build is slow.

gfwilliams:
As I understand it this PR currently replaces something that was working perfectly (albeit a bit slowly) with something broken?

Any extra work on this? Or shall I close it?

From my point of view I have received absolutely zero feedback that the Docker image is actually being used or is helpful to Espruino users, so working to speed the build up by 50% isn't super high priority.

wilberforce:
Time to close?

gfwilliams:
Yes :)

PR #1592, opened 2018-12-29
No exception on clearTimeout(invalidTimeoutId)

opichals:
Make `clearTimeout/Interval()` behave closer to other JS environments where it doesn't throw exceptions on invalid or stale timeout id is passed.


gfwilliams:
Thanks!

PR #1609, opened 2019-01-31
Update esp-idf to 3.1.2

wilberforce:


gfwilliams:
Is this good to merge? Any reason we couldn't just update the master branch of the build tools?

wilberforce:
Hi, because I want to test that it works first - I need to know the library links and works.

At this stage it's reporting as  3.1 rather than 3.1.2 so I have further testing to do.

I update the master branch in build tools when I know all is working. Sorry - it's a bit convoluted as two repositories have to be updated - I don't know if it can be simplified.

PR #1611, opened 2019-02-02
ESP32: update EspruinoBuildTools to esp-idf V3.1.2

wilberforce:
fix Wifi connect isssues (multiple ssid)


wilberforce:
Please merge

gfwilliams:
This appears to be literally *only* a modification to the Changelog - is that all you intended?

wilberforce:
Yes- as I have updated build tools master and the change log is enough to trigger the build

PR #1621, opened 2019-02-22
Esp32 neopixel debug

ExCyb:
Hi! Here is some fixes to esp32 neopixel from SergeP. Improved slightly from forum version to be more readable.

### DIFF
--- a/targets/esp32/esp32_neopixel.c
+++ b/targets/esp32/esp32_neopixel.c
@@ -48,7 +48,7 @@ typedef union {
 } rmtPulsePair;
 
 static uint8_t *neopixel_buffer = NULL;
-static unsigned int neopixel_pos, neopixel_len, neopixel_half;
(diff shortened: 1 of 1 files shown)


gfwilliams:
Forum post about this here: http://forum.espruino.com/conversations/330748/#comment14644546

gfwilliams:
@MaBecker should I merge this? You seem to do quite a bit of neopixel stuff, so does it improve life for you?

MaBecker:
Well the existing Neopixel implementation is working as expected.  

>  so does it improve life for you?

NO

gfwilliams:
Ok, thanks! Actually looking at the code again and at http://forum.espruino.com/conversations/330748/#comment14644546 it looks entirely likely that the existing ESP32 code had issues with longer strings of neopixels (also this code is a bit tidier than before) so merging anyway.

MaBecker:
Did some tests with up to 300 pixel,  still works as expected :)

PR #1626, opened 2019-03-06
Fixed fs.statSync() to report in localtime and corrected Month offset.

TomWS1:
Previous fs.statSync() reported file modification timestamp in UTC time, rather than localtime.  This update fixes that.
Also, previous statSync() didn't use 0 based month code.  This was fixed as well.


gfwilliams:
Thanks for this - if there's stuff like this that comes up in future with a static function in one file, it's worth making it non-static and exporting it so you can re-use it. It saves on code space and also makes it less likely things will break in the future :)

I'll quickly tweak that now though.

PR #1632, opened 2019-03-16
Fix bug where, upon asking for an open drain pin without a pullup, it…

swineone:
… is configured as having the pullup.


gfwilliams:
Thanks!

PR #1633, opened 2019-03-17
Workaround for building on gcc versions > 6.

swineone:


MaBecker:
Hi, @swineone,
what is the benefit of using a higher gcc version?

swineone:
> Hi, @swineone,
> what is the benefit of using a higher gcc version?

No idea, really. I'm just used to updating when new versions coming out -- I assume optimizations are better, etc.

Hopefully this doesn't dissuade anyone from merging this fix, as it's as simple and harmless as it gets.

gfwilliams:
Thanks for this - newer GCCs do sometimes get better code density and speed, but I've had issues with some (sometimes the manufacturer-provided HALs do non-compliant stuff and end up getting broken) so I don't tend to change compiler unless there's a good reason.

Do you have any background as to why swapping the order of object files around actually helps? What error does it fix? It just seems like something that shouldn't really have any effect at all?

gfwilliams:
Aha - I just spotted your forum post - thanks! Loads of info there :)

http://forum.espruino.com/conversations/331066/#comment14658810

PR #1639, opened 2019-04-01
fix compiler warning

MaBecker:
src/jswrap_string.c:512:13: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
     if (last<=jsvGetStringLength(parent))


gfwilliams:
Thanks!

PR #1641, opened 2019-04-09
Added wildcard for blacklist

jumjum123:
Used to blacklist full class.
Helpful to create customized firmware. For example if you don't need SPI in your project


gfwilliams:
Thanks!

PR #1648, opened 2019-04-15
Fixing issue with BLE serial on NRF52840

the1laz:
Fixing issue with BLE serial tx buffer not clearing on on NRF52840.


gfwilliams:
Awesome - thanks!

gfwilliams:
Referenced here: http://forum.espruino.com/conversations/323658/?offset=25#comment14698063

PR #1649, opened 2019-04-16
esp8266 - update  building doc (fix #1608)

### LINKED ISSUE #1608: Update building doc
Section Esp8266 is outdated

MaBecker:


gfwilliams:
Thanks! Is it worth noting that if you do the  `source scripts/provision.sh BOARDNAME` at the top of the readme, the majority of that is done automatically? I think it's just the com port that needs setting...

MaBecker:
> source scripts/provision.sh BOARDNAME

Yes, this  is the easiest way to setup a build environment  for Linux.

So if someone ist not on Linux then they have to setup it up manually.

PR #1652, opened 2019-04-23
Fix broken GPIO1 on nRF52840 chips

urish:
See http://forum.espruino.com/conversations/323658/?offset=25#comment14707531 for discussion


gfwilliams:
Nice - thanks!

PR #1653, opened 2019-04-23
Fix advertising on nRF52840

urish:
Also refactored `advertising_init()`, `jsble_advertising_start()` to eliminate the duplicate call to `sd_ble_gap_adv_set_configure()`.

### DIFF
--- a/targets/nrf5x/bluetooth.c
+++ b/targets/nrf5x/bluetooth.c
@@ -163,6 +163,8 @@ static bool                             m_in_boot_mode = false;
 
 #if NRF_SD_BLE_API_VERSION > 5
 static uint8_t m_adv_handle = BLE_GAP_ADV_SET_HANDLE_NOT_SET;                   /**< Advertising handle used to identify an advertising set. */
+static ble_advdata_t advdata;
+static ble_advdata_t scanrsp;
 #endif
 
 volatile uint16_t                       m_peripheral_conn_handle = BLE_CONN_HANDLE_INVALID;    /**< Handle of the current connection. */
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks - but I'm a bit iffy about doing this with static vars. I know it seems dumb but when we're targeting nRF51 we have 6kB of RAM in total (including stack and JS), so having two of the `ble_advdata_t` (which are reasonably big) is going to hurt.

I've just committed some code that should fix this by removing `advertising_init` completely. Also found that setAdvertising didn't work on 52840 so have fixed that too.

urish:
Actually, the static vars are only compiled in when targeting SDK >= 15, so I'm not sure if they are relevant for nRF51. But as long as advertising works on nRF52840 - I'm happy :)

gfwilliams:
Ahh, I think it might have been a copy/paste issue then?

```
#if NRF_SD_BLE_API_VERSION<=5
    static ble_advdata_t advdata;
    static ble_advdata_t scanrsp;
#endif
```

Still static, just defined in a different place :) Anyway, as you say it should be working now - and the code is hopefully a bit simpler, which is always good.

urish:
Yup, good spot - it was a copy-paste issue indeed. Always happy for simpler code!

gfwilliams:
Just so you know, it seems that updating the advertising data causes the scan response to be cleared - so chances are we'll need a static array for the scan response eventually on SDK15 anyway (although I imagine that'll take the form of `m_enc_scan_response_data` to ensure that we're not recomputing everything all the time).

gfwilliams:
But it's no big deal - right now it just means that when you update advertising it stops advertising the Nordic UART service

PR #1657, opened 2019-04-27
Feature characteristic security

vkolotov:
This PR is to add support for configuring security attributes for characteristics. NRF5X devices only as I don't have any ESP device handy (although it is possible to add support for ESP as well).

I've tested this with a BT sniffer, works ok. Once you marked a characteristic rear/write with a security attribute, a secure links gets established automatically when you are trying to read/write for android (I assume for ios as well) and Bluez with generic USB adapters. Bluegiga serial adapter does not establish secure link automatically, a separate pairing procedure has to be run before reading/writing protected characteristics.


gfwilliams:
Looks great - thanks!

PR #1662, opened 2019-05-16
fix src/jsvar.c:1859:5 warning

MaBecker:
src/jsvar.c:1859:5: warning: passing argument 1 of 'jshGetPinFromVar' discards 'const' qualifier from pointer target type [enabled by default]
     return (bool)jshIsPinValid(jshGetPinFromVar(v));
     ^
In file included from src/jsdevices.h:168:0,
                 from src/jshardware.h:25,
                 from src/jsinteractive.h:18,
                 from src/jsvar.c:18:


gfwilliams:
Thanks!

PR #1664, opened 2019-05-17
set variables to 1700

MaBecker:
freeHeap: 10856


gfwilliams:
Are you sure that's enough? the freeHeap might include stack too - so if you open eg. two TCPIP connections it might drop a lot

MaBecker:
ok, drop it for now and do some more tests

PR #1678, opened 2019-06-24
use espruino_snprintf, freeHeap +192byte (fix 1677)

MaBecker:


MaBecker:
Ups, compile for ESP8266_4MB failed and ESP8266_BOARD build without erros?!

checked: failed for code I have not touched?

Ok travis build is using gcc 5.4,I am using gcc 4.8.5, this this the reason

4.8.5 is the recommended version by espressif for SDK 2.2.1

MaBecker:
looks like SDK could not installed ....

==== ESP8266
ESP8266_NONOS_SDK-2.2.1
curl: (7) Failed to connect to github.com port 443: Connection timed out
gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now

gfwilliams:
Could you try making a minor commit to your branch? Like change some whitespace? 

It was probably just a one-off failure in Travis and that'll kick off a new build

gfwilliams:
Thanks - however it's a shame to have to have more tweaks in `jsutils` for `espruino_snprintf` - as now it's doing a copy of the format string every single time it's called. It's also strangely different to what's already done in jsutils to errors/etc: https://github.com/espruino/Espruino/blob/master/src/jsutils.h#L399-L416

So as I understand it, the issue is just accessing single bytes from flash storage? Your solution actually pulls the whole string out of flash into RAM for every string before running.

It looks like `vcbprintf` is the culprit as it just accesses chars directly:  https://github.com/espruino/Espruino/blob/master/src/jsutils.c#L705

so you could just change every instance of `*fmt` in `vcbprintf` to `READ_FLASH_UINT8(fmt)` and I think it'd work fine without changes? The ESP8266 specific hacks could probably also be improved (https://github.com/espruino/Espruino/blob/master/src/jsutils.h#L399-L416), albeit not removed since you still need to have to use FLASH_STR to force the strings into flash.

MaBecker:
>Thanks - however it's a shame to have to have more tweaks in jsutils for espruino_snprintf - as now it's doing a copy of the format string every single time it's called. It's also strangely different to what's already done in jsutils to errors/etc: 

Ok, so do not merge this, as 192 bytes are not a big win.

> So as I understand it, the issue is just accessing single bytes from flash storage? Your solution actually pulls the whole string out of flash into RAM for every string before running.

No sure about this

PR #1680, opened 2019-06-25
fix lcdGetPixel_SDL reading wrong pixels

fanoush:
Multiply by width was missing in GetPixel, now using same code as in SetPixel.
Current bug causes Graphics.scroll(x, y) to produce very strange patterns, after fix it works as expected.


gfwilliams:
Nice, thanks!

PR #1681, opened 2019-06-28
ESP8266 more heap (fix #1679)

### LINKED ISSUE #1679: add one more section to flash  with ld option --rename, only ESP8266_4MB
add rename .rodata=.irom.literal, because CCFLAG -mforce-l32 is used for ESP8266_4MB boards.

freeHeap +2064 byte

MaBecker:
- a win of +2064 byte for freeHeap, only for ESP8266_4MB
- print out header details for  section 0. - 5.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -16,6 +16,7 @@
             SDK15: Writing to flash now works
             nRF52840: USB Serial works even for big sends, and device swaps to USB automatically
             nRF52840: Allow Serial2 to be used
+            ESP8266: optimize rename-section for ESP8266_4MB board, freeHeap +2064 byte (fix #1679)
             
      2v03 : nRF5x: Fix issue when calling NRF.setAdvertising while connected via BLE (fix #1659)
             nRF5x: 'dump()' not outputs `NRF.setSecurity` line if it has been called.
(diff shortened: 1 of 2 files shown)


gfwilliams:
Nice! Does this actually use more flash memory then, so it can't be used on the 512k one?

Generally moving from RAM to Flash doesn't use more flash, since even when the data is in RAM, it has to be loaded from flash at boot time anyway.

MaBecker:
> so it can't be used on the 512k one?

True, because CFLAG `-mforce-l32` is needed for this rename, but this blows up the firmware size for ESP8266_BOARD builds.

```
  8 -rwxr-xr-x  1 markbecker  staff    2188 28 Jun 09:20 eagle.app.v6.data.bin
872 -rwxr-xr-x  1 markbecker  staff  444228 28 Jun 09:20 eagle.app.v6.irom0text.bin
 16 -rwxr-xr-x  1 markbecker  staff    7012 28 Jun 09:20 eagle.app.v6.rodata.bin
 56 -rwxr-xr-x  1 markbecker  staff   28576 28 Jun 09:20 eagle.app.v6.text.bin
** user1.bin uses  489172 bytes of 479232 available
espruino_esp8266_user1.bin too big!
```

gfwilliams:
Ahh, gotcha - so just to check `-mforce-l32` is actually applied to the 4MB already? I didn't see it in this change.

MaBecker:
not sure when this was added :)

https://github.com/espruino/Espruino/blob/master/make/family/ESP8266.make#L71-L74

gfwilliams:
Neat - thanks :)

PR #1683, opened 2019-06-29
add setOptions description (fix #1682)

### LINKED ISSUE #1682: setOptions description for TelnerServer library is missing
doc:

http://www.espruino.com/Reference#t_l_TelnetServer_setOptions

source code:

https://github.com/espruino/Espruino/blob/master/libs/network/telnet/jswrap_telnet.c#L79-L87

MaBecker:


gfwilliams:
Thanks!

PR #1686, opened 2019-07-08
Added remove blacklist for class and instanceof in build_jswrapper.py…

jumjum123:
… (function removeBlacklistForWrapper)


gfwilliams:
Thanks!

PR #1692, opened 2019-08-27
reverse  (fix #1679)

### LINKED ISSUE #1679: add one more section to flash  with ld option --rename, only ESP8266_4MB

MaBecker:


gfwilliams:
Thanks! Information on why is in https://github.com/espruino/Espruino/issues/1679

PR #1697, opened 2019-10-10
Does this fix the Puck build?

AkosLukacs:
Don't over bangle yourself Gordon :)


AkosLukacs:
Yes, this fixes the Puck build. Altho I'm just guessing :)

gfwilliams:
Thanks! I merged the bangle branch in a bit of a hurry when I found the DFU update wasn't going :)

Issue #1689, opened 2019-08-21
Building ESP8266_4MB in Docker fails


ipepe:
I'm trying to build ESP8266_4MB in Docker but I fail with error:
```
espruino-builder_1  | Installing dev tools for all boards
espruino-builder_1  | Provision BOARDNAME = ALL
espruino-builder_1  | Provision FAMILY =
espruino-builder_1  | ===== ESP32
espruino-builder_1  | pip 19.1.1 from /usr/local/lib/python3.7/site-packages/pip (python 3.7)
espruino-builder_1  | python/pip installed
espruino-builder_1  | pyserial installed
espruino-builder_1  | installing xtensa-esp32-elf-gcc
espruino-builder_1  | Folder found
espruino-builder_1  | GCC is /espruino/xtensa-esp32-elf/bin//xtensa-esp32-elf-gcc
espruino-builder_1  | ===== ESP8266
espruino-builder_1  | installing xtensa-lx106-elf-gcc
espruino-builder_1  | Folder found
espruino-builder_1  | GCC is /espruino/xtensa-lx106-elf/bin//xtensa-lx106-elf-gcc
espruino-builder_1  | ===== LINUX
espruino-builder_1  | ===== RASPBERRYPI
espruino-builder_1  | ===== NRF52
espruino-builder_1  | ===== ARM
espruino-builder_1  | installing gcc-arm-embedded
espruino-builder_1  | Folder found
espruino-builder_1  | Generating pin info
espruino-builder_1  | Generating platform configs
espruino-builder_1  | Generating JS wrappers
[...]

PR #1699, opened 2019-10-20
Added docker-compose.yml for easier building of Espruino

ipepe:
Regarding #1689 I fixed building Espruino in Docker, also added docker-compose so it's much easier to build yourself an image.


gfwilliams:
No reply...

ipepe:
What reply was needed @gfwilliams ?

gfwilliams:
I'd asked: `Does this line overwrite what was proviously specified in docker-compose.yml?` - you should see it above?

ipepe:
That's how it looks on my side: Screenshot 2020-07-27 at 11 29 55

I'm not sure what do You mean with that question. Which line?

gfwilliams:
Sorry - looks like that was my fault - I assumed review comments appeared automatically but it looks like it needed submitting...

This is what I had anyway...

image

So my concern was: Does it *always* build *ALL* boards now?

Also, looks like you removed the `RELEASE=1` line. The vast majority of builds for boards will fail without that in.

ipepe:
> Sorry - looks like that was my fault - I assumed review comments appeared automatically but it looks like it needed submitting...
> 
> This is what I had anyway...
> 
> image
> 
> So my concern was: Does it _always_ build _ALL_ boards now?

Yes. But You can override it by env variable in `docker-compose.yml`

gfwilliams:
Ok, great!

Looks like there are some merge conflicts now I'm afraid. Do you think you'd be able to help sort those out as I might break it doing so myself?

ipepe:
I will take a look this week, I will refresh this PR and see what cloud be done better. Do You want me to make any specific changes?

gfwilliams:
Thanks! I don't have any specific requests, no - it looks good

PR #1714, opened 2019-12-06
fail build if initialized data+code does not fit ito bootloader area

fanoush:
Initialized data is stored after code into FLASH  memory and is copied to RAM at startup, so testing only for code size is not enough.
Taken basically from https://stackoverflow.com/questions/23017522/deteaction-of-memory-section-overflow-in-ld


gfwilliams:
Awesome - thanks!

PR #1719, opened 2019-12-07
fix build error BOARD=LINUX on MACOSX (fix #1718)

### LINKED ISSUE #1718: Build BOARD=LINUX cause error
The Linux board has library `TENSORFLOW` which causes some build errors

On Mac OS X the gcc complains `c++` and cflag `-std=gnu99`:

* error: invalid argument '-std=gnu99' not allowed with 'C++'

* fix: remove that  flag for building on MACOSX

After removing that cflag a include is missing

* error:  /Volumes/Espruino/repos/Espruino/libs/tensorflow/tensorflow/lite/kernels/internal/optimized/neon_check.h:30:10: fatal error: 'NEON_2_SSE.h' file not found
#include "NEON_2_SSE.h"
         ^~~~~~~~~~~~~~

  https://github.com/tensorflow/tensorflow/issues/23440#issuecomment-436445860

* fix: add  `CCFLAGS += -DTF_LITE_DISABLE_X86_NEON` when buildung on MACOSX

MaBecker:


gfwilliams:
Thanks - but have you tried this when building for a microcontoller?

`MACOSX` gets set whenever building ON MacOS as far as I can see, regardless of whether you're building FOR it.

MaBecker:
> Thanks - but have you tried this when building for a microcontoller? 

Checked build for

* LINUX
* ESP8266_4MB 
* PICO
* PIXL
* BANGLEJS

gfwilliams:
Ahh, sorry - didn't realise this was under the ifdef for LINUX

PR #1720, opened 2019-12-09
new ellipse

MaBecker:
might be a bit slower
has no spilkes


gfwilliams:
Great, thanks! Would it be possible to translate the `fehlerhafter Abbruch bei flachen Ellipsen` comment to english?

MaBecker:
done

PR #1725, opened 2019-12-17
Improper decoding of string BLE UUIDs

### LINKED ISSUE #1724: Improper decoding of string BLE UUIDs
If you [setServices](https://www.espruino.com/Reference#l_NRF_setServices) using strings for BLE service / characteristic UUIDs on with master (2v04.294), you will get an error regarding hex characters, perhaps due to the revisions in https://github.com/espruino/Espruino/commit/d1675bb8f8412eb0817a219938765a209d71ee5c:

```
>NRF.setServices({
:  "BCDE" : {
:    "ABCD" : {
:      value : "Hello",
:      readable : true
:    }
:  }
:});
ERROR: empty UUID type
ERROR: empty UUID type
Uncaught Error: invalid Service UUID:UUID string should only contain hex characters and dashes
 at line 8 col 2
});
 ^
```

mfikes:
Fixes #1724


gfwilliams:
Great - thanks for this!

PR #1726, opened 2019-12-17
ESP32 mishandling of string BLE UUIDs in NRF.setServices()

### LINKED ISSUE #1723: ESP32 mishandling of string BLE UUIDs in NRF.setServices()
If you [setServices](https://www.espruino.com/Reference#l_NRF_setServices) using strings for BLE service / characteristic UUIDs on an ESP32 you will get errors.

For example, with 2v04:

```
>NRF.setServices({
:  "BCDE" : {
:    "ABCD" : {
:      value : "Hello",
:      readable : true
:    }
:  }
:});
ERROR: empty UUID type
ERROR: empty UUID type
=undefined
```

This results in the service being advertised, but under the wrong UUIDs. For me, instead of being advertised under the service 0xBCDE, I see 80084259-FFFF-FFFD-0000-00003FFC0000 and instead of the characteristic 0xABCD, I see 0xFEBC.

Note that if instead you use integer UUIDs, as in the following you get a correct advertisement.

```
>NRF.setServices({
:  0xBCDE : {
:    0xABCD : {
:      value : "Hello",
:      readable : true
:    }
:  }
:});
=undefined
```

Note there is a separate issue #1724 on master.

mfikes:
Fixes #1723

Note that, in the `BLE_UUID_TYPE_128` branch, the patch preserves this questionable assignment setting the `uuid` field:

```c
uuid->uuid = ((data[13]<<8) | data[12]);
```

With this patch, you can emperically see that 16-byte UUID characteristics are not yet properly honored on the ESP32, with the UUID depending on the 2-byte value assigned above.

My hunch is that, if the BLE characteristic [initialization code](https://github.com/espruino/Espruino/blob/3251c043d6afad2e44f219b6c003ed4555faa4c9/targets/esp32/BLE/esp32_gatts_func.c#L438) on the ESP32 is fixed to address this (so that 16-byte characteristic UUIDs work), then the preserved assignment mentioned above should be removed.

(On the other hand, service UUIDs empirically appear to be [properly initialized](https://github.com/espruino/Espruino/blob/3251c043d6afad2e44f219b6c003ed4555faa4c9/targets/esp32/BLE/esp32_gatts_func.c#L495) on the ESP32 and can handle 16-byte UUIDs.)

### DIFF
--- a/libs/bluetooth/bluetooth_utils.c
+++ b/libs/bluetooth/bluetooth_utils.c
@@ -185,9 +185,15 @@ const char *bleVarToUUID(ble_uuid_t *uuid, JsVar *v) {
   }
   return err_code ? "BLE device error adding UUID" : 0;
 #else
-  uuid->uuid = ((data[13]<<8) | data[12]);
-  for(int i = 0; i < 16; i++){
-	uuid->uuid128[i] = data[i];
+  if (expectedLength == 2) {
+    uuid->type = BLE_UUID_TYPE_BLE;
+    uuid->uuid = ((data[1]<<8) | data[0]);
+  } else {
+    uuid->type = BLE_UUID_TYPE_128;
+    uuid->uuid = ((data[13]<<8) | data[12]);
+    for(int i = 0; i < 16; i++){
+      uuid->uuid128[i] = data[i];
+    }
   }
   return 0;
 #endif


gfwilliams:
Thanks for your work on this - ESP32 BLE definitely needs some love (and looks like you already tracked down an issue that affects other devices as well).

The 12/13 assignment does (I believe) fit with the BLE spec - so if for instance the 128 bit UUID `0000xxxx-0000-1000-8000-00805F9B34FB` was specified (which is what a 16 bit UUID actually is) then the 16 bit UUID field would then contain the correct UUID.

But yeah, if the ESP32 code correctly used something other than ESP_UUID_LEN_16 then I guess it wouldn't matter anyway,

In some respects it might be safer to actually remove `->uuid` and just use `->uuidf128` for everything.

PR #1741, opened 2020-01-09
MDBT42Q - change neopixel clock pin to D23

AkosLukacs:
Moved the neopixel clock pin to the unrouted D23 pin.
Before: D22 couldn't be used. When I tried to check it with my DMM's frequency counter mode, that messed up neopixel timings. Apparently that is not a high-impedance input....
After: D22 can be used, and neopixel still works :)


gfwilliams:
Neat - thanks!

Issue #1721, opened 2019-12-10
New vector font


gfwilliams:
`g.fillPoly` now supports irregular polys, so there's no need for the regular polygons inside vector_font.h

A new font with irregular polygons should be significantly smaller

MaBecker:
Do you like to share what’s ob you mind?

gfwilliams:
The existing font generator (which I haven't actually published) leans on a bunch of stuff from another project and just mashes triangles together, so I can't just 'redo' the font.

So either:

* A script to mash the regular polygons in the existing font together
* A script to make a new vector font. I tried to see if I could do something quick, but no luck :) I think right now the best bet would be:
  * a script to make an SVG with all the characters in it
  * Open the SVG in Inkscape, do Object->Path and 'Flatten Beziers'
  * another script to read the flattened SVG and make a font file

MaBecker:
What about Hershey fonts you named in an earlier chat?

See https://github.com/espruino/Espruino/issues/1702#issuecomment-562940360

gfwilliams:
Hershey test code here: http://forum.espruino.com/conversations/321520/

I guess I could drop the vector font and use hershey fonts instead. My concern is really that I will get complaints if the font changes, but I guess it's not like the vector font is amazing at the moment anyway.

Also, the dumb way of doing wider lines for hershey would be to just draw each line as a separate polygon - but that's going to be really slow so I guess we need to make something that'll output a polygon (with smooth edges) from a standard polyline.

It'd be a neat thing to have, but it's easier said than done :)

MaBecker:
> It'd be a neat thing to have, but it's easier said than done :)

I totally agree, spend a few hours with line thickness and trying to draw some Hershey characters .....

MaBecker:
Have you thought about using bezier curves for drawing chart?

Any hints why g.fillPoly() is not filling this mirrored C as expected?

```
function quadraticBezier(x0,y0,x1,y1,x2,y2,drawCircle){
  var x,y,t,t2;
  var pp = [];
  var sn = 3;

if (drawCircle){
  g.drawCircle(x0,y0,3);
  g.drawCircle(x1,y1,3);
  g.drawCircle(x2,y2,3);
}
  pp.push(x0,y0);
  var s = Math.min(sn/Math.abs(x0-x2),sn/Math.abs(y0-y2));
  for( t = s; t <= 1; t += s) {
    t2 = t*t;
    tp2 = (1 - t) * (1 - t);
[...]

PR #1743, opened 2020-01-10
add squared bezier curve

MaBecker:
as discussed in `New vector font #1721`


MaBecker:
Not sure any more if this is really needed.....

MaBecker:
Not sure any more if this is really needed.....

gfwilliams:
I don't know - seems like a good utility function to have available. Could be useful for other stuff (like stepper motor control) too.

Although maybe in most cases it'd be fine to have the JS version. Anyone else have any thoughts?

MaBecker:
It's not lost, it will return at leased as  JS module ;-)

But for now it's just needed as a helper for creating fonts.

[quadraticBezier.c.txt](https://github.com/espruino/Espruino/files/4072516/quadraticBezier.c.txt)

PR #1755, opened 2020-02-10
support 1MB ESPs boards

MaBecker:
- add make flags FLASH_1MB and NO_FOTA
- add SPI1 labels to ESP8266
- fix some compiler warnings


gfwilliams:
Would it be possible to revert the warning fixes, and the removal of `urlize` I'd mentioned above? Then I'll happily merge this.

There's definitely at least one case where I'm pretty sure the removal of warnings breaks stuff that did work, and I don't have time right now to go through all of them and check.

MaBecker:
Just did, sorry for mixing things....

MaBecker:
Thanks

alievrenkut:
Thanks for 1MB support. Will you put this version at https://www.espruino.com/Download?

MaBecker:
No, you have to build your own.

alievrenkut:
> No, you have to build your own.

I read that we can flash 4MB builtd from file README_flash.txt. Is there any difference between flashing built 1MB version or flashing 4MB built using below info from readme:

For 1MB flash use 0xFC000/0xFE000 and for 2MB flash use 0x1FC000/0x1FE000 on the last line.

PR #1757, opened 2020-02-12
add g.quadraticBezier

MaBecker:


MaBecker:
Let me know I you like to double  the number of possible vertices from 64 to 128.

gfwilliams:
Thanks - does it make sense to make `sn` an optional thing, so the accuracy can be configured?

Just something like:

```
if (jsvIteratorHasValue(&it)) sn = jsvIteratorGetIntegerValue...
```

at the end of the iterator stuff would do it nicely.

MaBecker:
> does it make sense to make sn an optional thing, so the accuracy can be configured?

Yes, this allow full control and wipe out all the calc's with dx,dy ......

this is what it does now
```
# cacl 10 vertices
g.quadraticBezier([x0, y0, x1, y1, x2, y2],{"count":10})

# or calc how many vertices should be used max 10 min 1
g.quadraticBezier([x0,y0,x1,y1,x2,y2])
``` 

you mean changing it to something like this?

```
#  number of vertices to calc 
g.quadraticBezier(nvc,[x0,y0,x1,y1,x2,y2])
```

gfwilliams:
Oh right - sorry, not sure why I missed that! Looks great then!

Any chance of just removing the `jsiConsolePrintf`?

ahh... and by 64 verts you meant the drawPoly stuff? Yes, let's bump that up.

MaBecker:
> Any chance of just removing the jsiConsolePrintf?

done

gfwilliams:
Thanks!

PR #1772, opened 2020-03-07
Add Makefile to easy esptool-based flashing

opichals:
Try `make help` to see available targets and their descriptions.

### DIFF
--- a/make/targets/ESP8266.make
+++ b/make/targets/ESP8266.make
@@ -99,7 +99,9 @@ $(ESP_ZIP): $(USER1_BIN) $(USER2_BIN)
 	$(Q)rm -rf build/$(basename $(ESP_ZIP))
 	$(Q)mkdir -p build/$(basename $(ESP_ZIP))
 	$(Q)cp $(USER1_BIN) $(USER2_BIN) scripts/wiflash.sh $(BLANK) \
-	  $(INIT_DATA) $(BOOTLOADER) targets/esp8266/README_flash.txt \
+	  $(INIT_DATA) $(BOOTLOADER) \
+	  targets/esp8266/README_flash.txt \
+	  targets/esp8266/Makefile \
 	  build/$(basename $(ESP_ZIP))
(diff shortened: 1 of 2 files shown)


MaBecker:
Do you know that Makefile for ESP have some flash options?

MaBecker:
Most of them download a ready to go firmware and then they struggle because they bought cheap boards with slow flash modes.

MaBecker:
What about maintain the page and the Readme file too.

https://github.com/espruino/EspruinoDocs/blob/master/tutorials/ESP8266_Flashing.md

https://github.com/espruino/Espruino/blob/master/targets/esp8266/README_flash.txt

gfwilliams:
Thanks! Would it be possible to change the location of the Makefile though? And update https://github.com/espruino/Espruino/blob/master/targets/esp8266/README_flash.txt ? All the other boards have the makefiles in the `make` directory - for example:

* nRF52 flashing : https://github.com/espruino/Espruino/blob/master/make/common/NRF5X.make#L461
* STM32 flashing: https://github.com/espruino/Espruino/blob/master/make/common/STM32.make#L38

opichals:
The intention of my addition was that the release .zip files would contain [the Makefile](https://github.com/espruino/Espruino/pull/1772/files#diff-382db1665e3e7df44f93aec61c63bf8aR104) for people (me ;)) to avoid constructing the `esptool` command line arguments over and over when a firmware update is due. The `common/XYZ.make` files are not distributed along IIRC. There is already [ESP8266 flashing present in the makefiles](https://github.com/espruino/Espruino/blob/master/make/targets/ESP8266.make#L147).

The Makefile It is basically a copy of the README_flash.txt but for some people that still can be a valid source of information (those not using `make`) so I am not sure on that one.

gfwilliams:
Ahh, sorry - that makes way more sense. Thanks!

MaBecker:
@opichals Ok - got it.

Issue #1779, opened 2020-03-16
How to compile a firmware for ESP32 that can handle SD Cards


MaBecker:
```
/* jshint esversion: 6 */
//File_IO_and_SD_cards.js 

if ( process.env.BOARD == "ESP32") {

    M5Stack = require("M5STACKS");
    SPI1.setup({mosi:M5Stack.MOSI, miso:M5Stack.MISO, sck:M5Stack.SCK});
    E.connectSDCard(SPI1, M5Stack.CS);

    // see what's on the device
    //console.log(require("fs").readdirSync());
}


if ( process.env.BOARD == "PICO_R1_3") {

  // Wire up up MOSI, MISO, SCK and CS pins (along with 3.3v and GND)
  SPI1.setup({mosi:B5, miso:B4, sck:B3});
  E.connectSDCard(SPI1, B6 /*CS*/);
 
  // see what's on the device
  //console.log(require("fs").readdirSync());

}
```

No error on Pico, but this one on ESP32.

```
Uncaught Error: Unimplemented on Linux
 at line 5 col 37
    E.connectSDCard(SPI1, M5Stack.CS);
```

Are there some modules or -DSettings for ESP32 missing?

https://github.com/espruino/Espruino/blob/e568c6c7ac78e81b4ab40bb28bd92e8052f4e7b6/libs/filesystem/jswrap_file.c#L145-L162

MaBecker:
Wow the solution was there in the same file just a few lines above :)

https://github.com/espruino/Espruino/blob/e568c6c7ac78e81b4ab40bb28bd92e8052f4e7b6/libs/filesystem/jswrap_file.c#L22-L24

So the solution is to remove  this line 

https://github.com/espruino/Espruino/blob/e568c6c7ac78e81b4ab40bb28bd92e8052f4e7b6/boards/ESP32.py#L35


and build a new firmware version and it works.

AkosLukacs:
Just a hunch: does writing to internal flash still work?

gfwilliams:
Using Storage it would - FLASHFS is something different (FAT32 on the internal flash)

AkosLukacs:
That's great news, thanks for the clarification.
Btw, is there a list of boards.py or common compiler flags?

PR #1780, opened 2020-03-18
add SDCard support for ESP32

MaBecker:
by removing FLASHFS

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -122,6 +122,7 @@
             Changed 6x8 builtin font to a modified Dina_r400-6 supporting non-ASCII characters
             Bangle.js: Modify E.showMessage/Prompt/Alert to clear the entire screen and redraw widgets (fix #1771)
             Bangle.js: Modify Bangle.drawWidgets to only clear the areas where widgets actually are, not the whole top&bottom bar
+            ESP32: add SDCard support
 
      2v04 : Allow \1..\9 escape codes in RegExp
             ESP8266: reading storage is not working for boot from user2 (fix #1507)
--- a/boards/ESP32.py
+++ b/boards/ESP32.py
@@ -32,7 +32,6 @@ info = {
      'TELNET',
      'NEOPIXEL',
      'FILESYSTEM',
-     'FLASHFS',
      'BLUETOOTH'	 
    ],
    'makefile' : [


MaBecker:
added pr https://github.com/espruino/Espruino/issues/1779

gfwilliams:
FlashFS was there for a reason (to provide easy access to on-chip flash). I'm not sure it should just be removed?

@wilberforce was the one that added it AFAIK.

MaBecker:
Just checked, ESP32 is the one and only that is using FlashFS.

> FlashFS was there for a reason (to provide easy access to on-chip flash). I'm not sure it should just be removed?

Can you please lookup the download stat for ESP32. 

Personally I am a big fan of the Storage lib and try to avoid SDCards where ever it is possible.

There are only very little section added as flash 8k and 64k

https://github.com/espruino/Espruino/blob/e568c6c7ac78e81b4ab40bb28bd92e8052f4e7b6/targets/esp32/jshardware.c#L733-L742

Ups   0x2B0000 has to be changed to 0x310000

MaBecker:
just added https://github.com/espruino/Espruino/pull/1781

wilberforce:
I don't understand why removing this is an improvement to the ESP32 implementation.

All chips have 4MB or more of memory, and 1Mb had been reserved at work as FlashFS.

Why would you remove this to use an external SD card that only some of the boards have?

Worst case - make it switchable to use either - I just don't see what the benefit is here?

MaBecker:
Thanks for your input. Will check if both can be used.

gfwilliams:
Another option is now `Storage` is better, to extend the flash area used for that and remove FlashFS.

MaBecker:
Using both at the same time cause a undefined reference  to `sdSPISetup`

```
libs/filesystem/jswrap_file.o:(.literal.jswrap_file_kill+0x0): undefined reference to `sdSPISetup'
libs/filesystem/jswrap_file.o: In function `jswrap_file_kill':
/Espruino/repos/MaBecker/libs/filesystem/jswrap_file.c:214: undefined reference to `sdSPISetup'
libs/filesystem/jswrap_file.o: In function `jswrap_E_connectSDCard':
/Espruino/repos/MaBecker/libs/filesystem/jswrap_file.c:155: undefined reference to `sdSPISetup'
collect2: error: ld returned 1 exit status
```

MaBecker:
Not sure how to continue.

Fact :  Adding FLASHFS has broken SDCard function.

Suggestion: Remove FLASHFS for now and bring it back when someone has fixed the linker issue.

OR: Close this and remove SDCard support form ESP32 docs.

MaBecker:
Any hints how to handle the conflict named above?

gfwilliams:
I guess I feel a little like given how much flash there is available, having the filesystem built in is probably more useful than SD card support? I haven't seen too many ESPxx boards with SD cards on? Maybe just the ESP32CAM ones?

I think probably just update the docs to say it's not supported? I don't think you can easily have them coexist since FlashFS uses a 4k block size but SD cards will only want 512

MaBecker:
So let's close this.

PR #1783, opened 2020-03-21
add variable COMPORT for easier flashing 

MaBecker:


MaBecker:
hmm, travis failed for microbit and rak8211, i guess this is no issue.

gfwilliams:
Please could you add `COMPORT?=/dev/ttyUSB0` to the Makefile? That way flashing will still work as it did before if no port is supplied

MaBecker:
Sure, this way it is even simpler, because the two if's  can be removed - Thanks.

gfwilliams:
Perfect - thanks!

PR #1785, opened 2020-03-23
Fix bitwise rshift by 0 to be unsigned (fix #1784)

### LINKED ISSUE #1784: Bitwise `shift` by 0 bits doesn't convert to unsigned number (e.g. -7 >>> 0)
Shift by a non-zero number of bits do convert to unsigned. But with 0 it differs from V8:

Chrome
```js
>(-7 >>> 1).toString(16)
"7ffffffc"
>(-7 >>> 0).toString(16)
"fffffff9"
```

Espruino
```js
>(-7 >>> 1).toString(16)
="7ffffffc"
>(-7 >>> 0).toString(16)
="-7"
```

Based on this I would think this should work though...
https://github.com/espruino/Espruino/blob/57da26184a8cc54fa31e0ec5f4b5024026487b1b/src/jsvar.c#L3428

opichals:


gfwilliams:
Nice - thank you!

PR #1789, opened 2020-03-29
Fix ESP8266 bundle Makefile values and add vars

opichals:


MaBecker:
Hmm, did you check content of `make/family/ESP8266.make` and `make/targets/ESP8266.make` because those to file are missing stuff as well, please double check.

Using env COMPORT  so `export COMPORT=<your com port>` 
 
It is also possible to pass ET_FM and ET_FF via command line during make 

`make clean && ET_FM=dio ET_FF=40m` make flash`

MaBecker:
What about [README_flash.txt](https://github.com/espruino/Espruino/blob/master/targets/esp8266/README_flash.txt)?

opichals:
@MaBecker I realize this is redundant to README_flash.txt distributed along the binaries. And to be frank I am not sure this is helping as it may become out-of-sync more easily.

Not sure what you have in mind with regards to the `make/family/ESP8266.make` and `make/targets/ESP8266.make` as those IMO serve different audience - the firmware builders'.

opichals:
> may become out-of-sync more easily

Just want make sure we are on the same page. I am completely fine dropping the `Makefile` I added unless we find an agreeable and beneficial setup.

gfwilliams:
Thanks! 

Well, until we figure out what to do, this feels like an improvement.

While we've got the PR open, should `PORT=/dev/tty.wchusbserial*` be `PORT?=/dev/tty.wchusbserial*` Then at least you can do `PORT=/dev/foo make ...`

opichals:
I normally do `make PORT=/dev/foo` to set the PORT variable.

gfwilliams:
Interesting - I didn't think that'd work when it's being explicitly overwritten.

MaBecker:
Cool, just checked the latest travis build and the Makefile is already part of it.

PR #1792, opened 2020-04-01
change name to lang

MaBecker:
noticed different names, when running a MixedClock in  Emulator


gfwilliams:
Please can we change the JS instead? I've just shipped 1500 watches with this firmware, and realistically many of those won't be updated by users for a long time.

MaBecker:
Sure, I will do so, so let's close this.

gfwilliams:
Thanks!

PR #1808, opened 2020-04-29
calculate free flash space also for SPI flash and for code both in internal and external flash

fanoush:
- fix bug that with SPIFLASH enabled and code in external flash reported free flash size is very large and wrong  (FLASH_SAVED_CODE_START-firmwareEnd)
- add both internal and external flash as separate parts, handle also saved code in the middle of SPI flash

Without this I get something like
```
>f.getFree()
=[
  { addr: 405504, length: 1613877248 }
 ]
```
With this patch and code set in board file as
```
  'saved_code' : {
    'page_size' : 4096,
#    'address' : ((118 - 10) * 4096), # Bootloader takes pages 120-127, FS takes 118-119
#    'pages' : 10,
    'address' : 0x60380000, # P8 empty flash area with no fonts/images or firmware flash space
    'pages' : 32,
#    'flash_available' : 512 - ((28 + 8 + 2 + 10)*4) # Softdevice 2.0 uses 28 pages of flash, bootloader 8, FS 2, code 10. Each page is 4 kb.
#    'flash_available' : 512 - ((31 + 8 + 2 + 10)*4) # Softdevice 3.0 uses 31 pages of flash, bootloader 8, FS 2, code 10. Each page is 4 kb.
#    'flash_available' : 512 - ((35 + 8 + 2 + 10)*4) # Softdevice 5.0  uses 35 pages of flash, bootloader 8, FS 2, code 10. Each page is 4 kb.
    'flash_available' : 512 - ((35 + 8 + 2)*4) # Softdevice 5.0  uses 35 pages of flash, bootloader 8, FS 2, no code. Each page is 4 kb.

```
I get this
```
>f.getFree().forEach((i)=>{console.log("addr 0x"+i.addr.toString(16)+" len 0x"+i.length.toString(16));})
addr 0x63000 len 0x13000
addr 0x60000000 len 0x380000
addr 0x603a0000 len 0x60000
```
BTW I cheated a bit with __isr_vector, there is no linker symbol for start of binary and to calculate free flash space from FLASH_AVAILABLE_FOR_CODE value one needs to know start of binary (=size of softdevice).

There is also this
```
    codeOut("#define BOOTLOADER_SIZE                 "+str(common.get_bootloader_size(board)))
    codeOut("#define ESPRUINO_BINARY_ADDRESS         "+hex(common.get_espruino_binary_address(board)))
```
but it doesn't return anything sensible for nrf5x.

### DIFF
--- a/targets/nrf5x/jshardware.c
+++ b/targets/nrf5x/jshardware.c
@@ -1740,11 +1740,35 @@ JsVar *jshFlashGetFree() {
   /* Try and find pages after the end of firmware but before saved code */
   extern uint32_t LINKER_ETEXT_VAR; // end of flash text (binary) section
   uint32_t firmwareEnd = (uint32_t)&LINKER_ETEXT_VAR;
+  extern uint32_t __isr_vector; // defined in gcc_startup_nrf5x.S, first address of binary
+  uint32_t firmwareStart = (uint32_t)& __isr_vector;
   uint32_t pAddr, pSize;
   jshFlashGetPage(firmwareEnd, &pAddr, &pSize);
   firmwareEnd = pAddr+pSize;
+#ifdef SPIFLASH_BASE
+#if SPIFLASH_BASE <= FLASH_SAVED_CODE_START
+  // we have storage in external flash (so we cannot take is as end of free internal flash)
+  if (firmwareEnd < firmwareStart+FLASH_AVAILABLE_FOR_CODE)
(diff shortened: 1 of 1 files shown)


fanoush:
and BTW linux build fails because of misspelled LENGTH here https://github.com/espruino/Espruino/blob/master/src/jshardware_common.c#L52
was afraid it was me as I just added similar code but that's not my change :-)

gfwilliams:
Thanks!

PR #1817, opened 2020-05-07
allow different board name than board filename

fanoush:
this allows multiple board files for same board with different build options (like e.g. different SDK version or build options)
so now I can have e.g. P8SDK11.py and P8SDK14.py with 
```
info = {
 'name' : "P8 DaFit smartwatch",
 'boardname' : 'P8', # visible in process.env.BOARD
 'default_console' : "EV_BLUETOOTH",
```
and both builds will have proccess.env.BOARD set to P8

BTW the $(BOARD) for makefile is still the board file name since it is build like
```
make -j BOARD=P8SDK14 RELEASE=1
```
but it is OK and expected and does not cause issues.

### DIFF
--- a/scripts/build_platform_config.py
+++ b/scripts/build_platform_config.py
@@ -46,6 +46,9 @@ print("BOARD "+boardname)
 board = importlib.import_module(boardname)
 pins = board.get_pins()
 # -----------------------------------------------------------------------------------------
+if "boardname" in board.info:
+  boardname = board.info["boardname"]
+  print("BOARDNAME "+boardname)
 
 LINUX = board.chip["family"]=="LINUX"
 EMSCRIPTEN = board.chip["family"]=="EMSCRIPTEN"


fanoush:
Also there is one more related change I can include too - I'd prefer omitting `-D$(BOARD)` here
https://github.com/espruino/Espruino/blob/master/make/common/NRF5X.make#L31

For me it causes just troubles and it can be defined in board file if  needed. And often one needs different board name anyway as seen e.g here https://github.com/espruino/Espruino/blob/master/boards/NRF52832DK.py#L41

With  that extra define I cannot name board files with `-` inside, the C compiler complains it is invalid character.

gfwilliams:
Thanks - I'm not too sure about the removal of `-DBOARD` - it's used a lot in the code for board-specific hacks.

> With that extra define I cannot name board files with - inside

I feel like it's not really important enough to justify changing basically every board.py file? Can you not just use underscore?

With this change, as long as you use underscores in your filename you should be able to make the device boardname appear as whatever you want?

fanoush:
>  I'm not too sure about the removal of -DBOARD - it's used a lot in the code for board-specific hacks.

Yes, that's why I did not include this change. And yes, underscores are ok too, actually used them already because of this. The issue was more philosophical - if it is OK to generate C define directly from some file name. Yes it needs adding explicitly e.g.  -DPUCKJS into its board file if code has dependencie on it - well, just like any other define already there.

Actually I found another way that might be doable, there is also `CURRENT_BOARD.make` file generated from board python file too so -D<boardname> can be put there automatically just like e.g. `DEFINES+=-DUSE_BOOTLOADER` now. So for my case above it would add `DEFINES+=-DP8` for both board files instead of -DP8SDK11 or -DP8SDK14 as it is now. and then -DBOARD could be removed.

Would that be OK, should I try?

gfwilliams:
Adding `DEFINES+=-DP8` seems like a good solution for now if it works for you... but doing `-DBOARDNAME` when creating `CURRENT_BOARD.make` and then removing `-DBOARD` might work?

You'd need to remove `-DBOARD` from all the other targets too as well, not just nRF52.

It's quite a bit of work and I'll be fed up if I go to build the next release and end up spending an hour debugging builds just so you could have dashes in filenames :)

fanoush:
>  just so you could have dashes in filenames :)

It was more about making the filename of <board>.py irrelevant and storing the real board name inside the file (and take filename just as fallback). Dash in filename is just sign that it is in a way 'broken'. However I understand what you say, it may be not worth possibly breaking stuff if you feel so.

> You'd need to remove -DBOARD from all the other targets too as well, not just nRF52.

Oh, the makefiles? Yes, that's even better so it is not duplicated everywhere.  I guess the CURRENT_BOARD.make is generated for all targets. So the result will be the same, the definition -D$(BOARD) will just move from this place https://github.com/espruino/Espruino/blob/master/make/common/NRF5X.make#L31 (and other target makefiles as you say) to CURRENT_BOARD.make, however it will possibly use the overriden board name from board file instead of file name. And if not overridden it will be the same as now. And it is just C compiler definition so it won't affect makefiles logic.

So if you have BANGLEJS.py and another BANGLEJS-notensorflow.py with boardname set as BANGLEJS both will have -DBANGLEJS automatically from CURRENT_BOARD.make instead of useless -DBANGLEJS-notensorflow 'board' definition as now. 

Anyway same -Dxxx does not hurt, so it can be in both places defining possibly both (same or different) values to not break stuff. In fact I don't care about dashes that much :-)

gfwilliams:
>  So the result will be the same, the definition -D$(BOARD) will just move

That sounds great! If you can do a PR for that, and `scripts/create_zip.sh` runs without errors (eg all the board files will build) then that would be awesome.

I could be wrong but I'd *hope* that all the board binaries should actually be the same before & after if it's worked properly?

PR #1823, opened 2020-05-16
Add support for testing arbitrary directories

cmF3:
This small change adds --test-dir to the espruino binary
to allow testing against arbitrary directories.

For example, you can now test espruino against the 69000 tests of the ECMAScript Test Suite[1].

Happy testing!

[1] https://github.com/tc39/test262

### DIFF
--- a/targets/linux/main.c
+++ b/targets/linux/main.c
@@ -109,40 +109,37 @@ bool run_test(const char *filename) {
   return pass;
 }
 
-
-bool run_all_tests() {
+bool run_test_dir(const char* testdir) {
   int count = 0;
   int passed = 0;
+  char absdir[PATH_MAX];
+
+  realpath(testdir, absdir);
 
   char *fails = malloc(1);
   fails[0] = 0;
 
-  DIR *dir = opendir(TEST_DIR);
+  DIR *dir = opendir(absdir);
   if(dir) {
     struct dirent *pDir=NULL;
     while((pDir = readdir(dir)) != NULL) {
-      char *fn = (*pDir).d_name;
-      size_t l = strlen(fn);
-      if (l>3 && fn[l-3]=='.' && fn[l-2]=='j' && fn[l-1]=='s') {
-        char *full_fn = (char *)malloc(1+l+strlen(TEST_DIR));
-        strcpy(full_fn, TEST_DIR);
-        strcat(full_fn, fn);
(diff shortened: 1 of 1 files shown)


targets/linux/main.c
```diff
@@ -73,173 +165,168 @@ bool run_test(const char *filename) {
...
-    jsiConsolePrintf("Uncaught %v\n", exception);
+    jsiConsolePrintf("Uncaught %v", exception);
     jsvUnLock(exception);
     e = 1;
   }
 
   if (jspIsInterrupted()) {
     jsiConsoleRemoveInputLine();
-    jsiConsolePrint("Execution Interrupted.\n");
+    jsiConsolePrint("Execution Interrupted.");
```
gfwilliams, line 329: Is there a reason for this? jsiConsolePrint doesn't add a newline so I'd be concerned this would cause issues

targets/linux/main.c
```diff
@@ -376,7 +477,7 @@ int main(int argc, char **argv) {
   while (isRunning) {
     jsiLoop();
   }
-  jsiConsolePrint("\n");
+  jsiConsolePrint("");
```
gfwilliams, line 480: Again I'm pretty sure this line does absolutely nothing now...

gfwilliams:
Hi - thanks - does this PR do anything than add configurable directories? There's a lot of code changed in there.

Did you have any success with test262? When I tried it quite a bit of extra boilerplate was required to get things working.

cmF3:
Hi,

sorry, the later two commits weren't supposed to go into this PR. I forgot to check them into another branch.

Yes, this PR basically just adds configurable tests directories. The later commits expand on that and feature some logging improvements (and formatting). I just find-replaced the EOLs, so some lines make little sense now except for maintaining "backwards compatibility" for how the output looks. I can remove them.  The code changes look like a lot because I ran clang-format over main.c.

The --test-dir argument makes use of ftw(3), which is recursive, so I could to ./espruino -d test262 and just enumerate over all the tests. I have not found the time yet to look at the resulting output in detail but yeah, there were a lot of errors.

This is somewhat of-topic but I noticed that the espruino library exports have global state. Have you considered offering a reentrant API or was there a specific consideration not to?

gfwilliams:
Ok, thanks! Just merging...

> espruino library exports have global state

Not sure what you mean here? The JS library?

PR #1834, opened 2020-05-31
Correct instruction for gcc install

FransM:
Directory after untar was wrong (as also can be seen from the next line with the mv command.


gfwilliams:
Thanks!

PR #1835, opened 2020-06-01
fillPoly tweaks

nebbishhacker:
I fixed an issue where polygons that share edges could end up overlapping slightly, and made a small tweak to the x-axis rounding behaviour.

I'm also starting to rethink my suggestion to add the half-pixel offset - it does make some things look a bit nicer, but it also means that fillPoly and drawPoly no longer line up when given the same points. I've added a couple of tests to illustrate the issue.


gfwilliams:
Thanks - yeah, the big difference between drawPoly/fillPoly looks bad - I don't mind overlap on one side or the other between the two but having a gap is going to annoy people.

It looks like we were already doing a 0.5px addition: https://github.com/espruino/Espruino/blob/master/libs/graphics/jswrap_graphics.c#L1671

So maybe the solution is actually not to do *any* addition at all in either of the two places?

nebbishhacker:
That code you link to is adding 0.5 immediately after multiplying the coordinate by 16, so it's actually effectively a 1/32px addition. I don't think it makes very much difference.

gfwilliams:
Good point, thanks. Brain's not working today :)

PR #1836, opened 2020-06-01
Reduce poly count and fix missing chars in vector font

nebbishhacker:
Now that the scanline bug is fixed, I was able to reduce the number of individual polygons by a lot. I also fixed a bug that was preventing some characters from being embedded in vector_font.c.


gfwilliams:
Wow, that's great - thank you!

You know I mentioned that the accents added ~500 bytes of code, while you reckoned just having the chars would add ~1000 bytes. With these changes do you think it's closer? If it's only a few 100 bytes I think it'd be worth it having things simpler.

nebbishhacker:
It might be a bit closer, yeah. I'll try including the chars directly later so we can see how it compares.

PR #1837, opened 2020-06-01
Use Dockerfile layers to cache dependency install

ikkala:
Before dependency install (provision.sh ALL), copy only required files.
Only after that, copy rest of the source code (COPY . /espruino) that
changes more often.
This way usually only lines starting from COPY . /espruino will
be executed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,8 @@
 # 2: Run container image so it builds espruino
 #
 #   docker run -e BOARD='PICO_R1_3' --name container_name img_name
+# or
+#   docker run -e BOARD='BANGLEJS' -e DFU_UPDATE_BUILD=1 --name container_name img_name
 #
 # This will run the container and save build results into the container's filesystem.
 # Near the end of the build the filename will be displayed, for example espruino_2v00_pico_1r3.bin 
@@ -22,7 +24,6 @@
 
 FROM python:3
 
-COPY . /espruino
 WORKDIR /espruino
 
 # Workaround add some stuff that the provision script uses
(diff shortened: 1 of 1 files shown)


ikkala:
> Before dependency install (provision.sh ALL), copy only required files.

Note, you should review if the selected directories are enough / too much. I went through them by trial and error -- at least the whole build goes ok and I'm able to pull out zip fro Bangle.js.

gfwilliams:
What's the reasoning for this? It just makes it faster by avoiding copying the `espruino` directory over? 

How much time does this save? I just tried here and it takes 2 seconds, so I'm not sure it warrants all the potential breakage that this might introduce?

gfwilliams:
... or this actually allows provision.sh to not be called except right at the start?

ikkala:
Each (almost) Dockerfile line creates new layered filesystem layer. The o[ld Dockerfile copies the source `COPY . /espruino` in very early step](https://github.com/espruino/Espruino/pull/1837/files#diff-3254677a7917c6c01f55212f86c57fbfL25) to a layer and after that the very long running dependency install and provision.sh steps are executed. Those long running steps takes some minutes to run, and what's worst is that they produce always multiple gigas of layer data to disk.

So, in development cycle where we change just one line of code and want to then build all that happen.  The new ordering of the Dockerfile tries to avoid this by moving the `COPY . /espruino` to the end so that docker build doesn't have to trigger the long running early steps at all.

ikkala:
> I just tried here and it takes 2 seconds

Did you try docker build, after changing some source code?

gfwilliams:
> Did you try docker build

Oh no, I mean I literally just tried the copy.

This sounds like a great idea then. I think the directories you've copied should be enough - do you actually need `espruino/targets` though? You may not...

ikkala:
> do you actually need `espruino/targets` though? You may not...

Looks like it's fine without that, so I removed it now.

gfwilliams:
Thanks!

PR #1839, opened 2020-06-01
GPS-raw event: handle UBX protocol

### LINKED ISSUE #1838: Bangle.js: GPS-raw event strings vs. UBX Protocol binary messages
When investigating different approaches to the [A-GPS](https://github.com/espruino/Espruino/issues/1827) realized that the current GPS-raw event, providing messages as strings, works well with NMEA messages introduced in u-blox GPS receiver spec (linked [at page](https://www.espruino.com/Bangle.js+Technical)) but when it comes to u-blox specific UBX protocl, the events are failing to provide all the data.

For example, [the UBX-ACK-NAK (0x05 0x00) messages at GPS-comms.js example](https://github.com/espruino/BangleApps/blob/d8c3b73334c8fdb471f71b9c4a9091873182f752/testing/GPS-comms.js#L16) never come fully through because the zero byte is terminating the string. The message payload coming after the zero is missing totally.

Basically all messages containing zero bytes are truncated.

To fix this, either:
- Change GPS-raw event to provide ArrayBuffer instead of string (ArrayBuffer is trivial to convert to string is JS land when needed)
- or, introduce another event type for u-blox messages

In both options, the underlying implementation must understand which of the protocol it is currently receiving.

ikkala:
Bangle.js: Changed GPS-raw event type from string to ArrayBuffer
and handle u-blox UBX Protocol messages.

Fixes #1838

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -54,7 +54,7 @@
             Puck.js: Fix Puck.magTemp();
             Puck.js v1: Fix magnetometer reading after puck.magOn (2v05 regression) - use old software I2C implementation
             I2C: Improve software I2C (now quickly forces I2C high before pullup, to help improve clock speed)
-            Bangle.js: Changed GPS-raw event type from string to ArrayBuffer and handle u-blox UBX Protocol messages (fix #1838)
+            Bangle.js: Fix GPS-raw events to provde also u-blox UBX Protocol messages (fix #1838)
 
      2v05 : Add Array.includes
             Fix (Number.toFixed) rounding, eg (1234.505).toFixed(2)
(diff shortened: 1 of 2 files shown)


ikkala:
btw, what's the deal with https://github.com/espruino/Espruino/blob/master/libs/banglejs/jswrap_banglef5.c ? I realize my earlier little / cosmetic changes were not done to that file -- for what purpose is that? Something obsolete?

gfwilliams:
Originally I planned to use the F5 watch (https://uk.gearbest.com/smart-watches/pp_1063842.html) for Bangle.js - and the code is for that. I kept it around because during development some folks had the F5 and wanted to keep testing with it.

There was always the hope I'd try and share common code but basically yes, it's obsolete now - and I'll move it into the unsupported boards branch soon.

gfwilliams:
Thanks!

PR #1841, opened 2020-06-02
Include accented characters directly in fontmap

nebbishhacker:
I made the font map use a 13x19 grid to give the characters a bit more headroom, added the accents directly, and removed all the accent handling code.

The net size increase due to this change is 840 bytes, going by the contents of the zip produced by  `BOARD=BANGLEJS RELEASE=1 DFU_UPDATE_BUILD=1 make`.


gfwilliams:
Wow, thanks! Shame it's that much but even so it'll be a significant improvement over the old vector font, and it contains twice as many characters!

Thanks for all your work on this! It must have taken ages!

PR #1844, opened 2020-06-03
GPS-raw event: check and return UBX protocol checksums

ikkala:


gfwilliams:
Ok, thanks! I was just about to commit my changes for reduced RAM so this'll make the merge interesting anyway :)

gfwilliams:
Sorry - just had to take this out again. It really messes with work I did to get the buffer size down.

While I'd managed to merge it, it turns out that if the message length was bufferSize+1, there'd be bufferSize bytes already queued up, so there wouldn't be the CRC bytes.

IMO something like this should probably be done outside of the interrupt. It's so rare UBX packets will be used I think it's reasonable to do this in JS land - it would also help to know if you'd got a packet but it was corrupt - with the CRC check in the IRQ you wouldn't see that at all.

ikkala:
Ok.. Is your new implementation still giving the checksum bytes in GPS-raw event to the API user, like in my fix? So that when I restore the nav db dump back to GPS receiver I don't "have to" calculate the checksums manually..?

gfwilliams:
We just need to add 2 to ubloxMsgLength?

gfwilliams:
If so I'll dump that in now

gfwilliams:
ok, just done

ikkala:
Thanks!

PR #1848, opened 2020-06-05
Reduce size of vector font

nebbishhacker:
Instead of storing everything in a single array delimited by magic
values, have separate arrays containing
- vertices for each polygon
- \# of vertices for each polygon
- polygon indices for each char
- \# of polygons for each char

This arrangement means that multiple chars can reference the same
polygons. build_vector_font.js takes advantage of this by
normalizing and deduplicated polygons where it can.

Shaves exactly one kilobyte from the final build.


gfwilliams:
Wow! This is extremely cool - thank you!

So actually it's a bit like we were doing with the accents, but it's all done automagically by the build script? That's amazingly clever.

PR #1850, opened 2020-06-05
Eliminate spiFlashReadWrite

FransM:
This is only called once, and can also be done with a Write followed by a Read


gfwilliams:
Neat - thanks!

PR #1855, opened 2020-06-08
Add Graphics.transformVertices

nebbishhacker:
Adds a method to the graphics class for applying a transformation to a
list of vertices. This method can be used alongside drawPoly to translate,
rotate and scale complex polygons in real time.

When I modify my original svg2bangle demo to use this, it stops looking like a slideshow and starts looking like an actual animation :)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -60,6 +60,7 @@
             Bangle.js: Improve SPI flash speed by with specific function for reading and keeping CS asserted (fix #1849)
             Bangle.js: Ensure BTN3 exits debug mode (fix #1842)
             Bangle.js: Now warn if GPS data overflows (fix #1847)
+            Add Graphics.transformVertices()
 
      2v05 : Add Array.includes
             Fix (Number.toFixed) rounding, eg (1234.505).toFixed(2)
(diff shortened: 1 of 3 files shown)


MaBecker:
Nice, have you thought about having scale for x and y as separate values?

nebbishhacker:
I considered it. I would like to keep the `scale` option in order to stay consistent with drawImage, but additional `scaleX` and `scaleY` options could be added.

You can also do pretty much whatever transformation you want by passing a transformation matrix. To do non-uniform scaling:
```
g.transformVertices(verts, [scaleX, 0, 0, scaleY, 0, 0]);
```

gfwilliams:
This looks great - I was wondering how to implement a transform nicely given only draw/fillPoly could use it, and this seems to fit well. I'm all for the current solution with a single scale - specifying a matrix is so easy and if someone wants to do more complex stuff we could always write a module for it.

I'll just change the 'jsvGetFloatAndUnLock' to `jsvReadConfigObject` after merge though... https://github.com/espruino/Espruino/blob/master/libs/graphics/jswrap_graphics.c#L2272-L2279

It should be marginally more efficient but will also error if the user makes a typo - eg `{rotote:50}`

PR #1858, opened 2020-06-09
Remove half-pixel offset from fillPoly

nebbishhacker:
This makes the results of fillPoly align with those of drawPoly when
both are fed whole-pixel coordinates.

For applications where it makes a noticeable difference the offset can
be added manually before calling fillPoly, which is what the vector
font rendering code now does.


gfwilliams:
Ok, thanks! And this fixes the issue where a drawPoly+fillPoly would leave gaps?

PR #1862, opened 2020-06-13
Make Bangle.showLauncher() use less ram

nebbishhacker:
I have a lot of apps installed, and I noticed the loading the launcher was failing with `MEMORY` errors even when memory utilization was only at ~65%. This change solves that problem.


gfwilliams:
Wow, thanks! Yeah, that's a great addition.

Issue #1806, opened 2020-04-24
Any plans to extend the WIZnet550 driver?


MaBecker:
Would be cool to have 

* getHostname()  and setHostname() 

* reset() to release ip given with setIP() only for WS550 because it has a default ip after reset/powerup

* ......

To Be extended ...

gfwilliams:
> getHostname

If the hostname is actually configurable, then yes. I'm not sure if the driver does mDNS, so if it doesn't there's be no point having the hostname.

> reset() 

What's the use for this one? Can you not just call setIP with `0,0,0,0`?

One thing I've started doing is adding a function that allows the user to send command straight to the chip. Might be a way to allow some of the more interesting functionality to be exposed...

MaBecker:
@opichals do you have any whishes or needs to enhance WIZnet lib?

MaBecker:
> I'm not sure if the driver does mDNS

Look's like mDNS and uses a fixed hostname.

This is hows it could be [done](https://github.com/arduino-libraries/Ethernet/pull/50/commits/a45bb068120e1e5bdf7943222373f3ad8076ba68#diff-961b3f6679dc8b161750718d975b9747). 

> What's the use for this one? Can you not just call setIP with 0,0,0,0?

Some boards have a reset pin and if there is a reset they startup with a default like  this 

so just a optional pin {rst: 
WIZ550io

```
={
  ip: "192.168.1.2",
  subnet: "255.255.255.0",
  gateway: "192.168.1.1",
  dns: "0.0.0.0",
  mac: "00:08:dc:57:c1:f0"
 }
```

Ok, for other board a setIP({values}) is ok.

gfwilliams:
Ahh, So I'm not sure that's mDNS - it just gives a hostname when it configures itself via DHCP? That sounds like a great thing to add.

I'm not too sure about the reset - for the cases where you need it, can't you just toggle the reset pin yourself?

MaBecker:
What about adding this too?

* ping(cb) 

* getState()   return value of JsNetworkState as string

gfwilliams:
That'd be cool if you want to add it.

Although with ping - maybe be careful it doesn't add too much extra size to the binary?

MaBecker:
```
>eth.getStatus()
={
  state: "online",
  id: "W5500",
  phylink: "up",
  phypowmode: "normal"
 }
> 
```

MaBecker:
[...]

PR #1870, opened 2020-07-01
wiznet_addons

MaBecker:
* setHostname()
* getHostname()
* getStatus()

### DIFF
--- a/libs/network/wiznet/DHCP/dhcp.c
+++ b/libs/network/wiznet/DHCP/dhcp.c
@@ -44,10 +44,14 @@ uint32_t DHCP_XID;
 
 RIP_MSG  MSG;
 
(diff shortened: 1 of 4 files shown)


MaBecker:
#1806

gfwilliams:
This looks great - thanks!

MaBecker:
Can't find the new function in the new function in http://www.espruino.com/Reference#WIZnet, 
is there something missing in JSON{}?

gfwilliams:
You'd declared them to be part of `Ethernet` which I think is the right class anyway. They seem to be there just fine:  http://www.espruino.com/Reference#Ethernet

MaBecker:
Thanks for checking

PR #1880, opened 2020-07-13
ESP32 improvements

MaBecker:
- add FLASH_BAUD
- fix erase_flash
- switch jsWarn()  to jsDevInfo()
- keep jsWarn()/ jsError() for not implemented functions
- switch jsError() to jsExceptionHere() if helpfull

### DIFF
--- a/libs/filesystem/fat_sd/flash_diskio.c
+++ b/libs/filesystem/fat_sd/flash_diskio.c
@@ -42,7 +42,6 @@ DSTATUS disk_initialize (
 {
   if (drv != 0)
      return STA_NODISK;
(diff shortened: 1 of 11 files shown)


MaBecker:
#1880

gfwilliams:
Thanks - looks great!

However with `bluetooth.c`, don't we actually want to keep the warnings, or some kind of print that appears for normal release users? It feels quite important to tell them that a function isn't implemented - potentially it should even be an exception.

MaBecker:
Thanks, yes keep jsWarn() for BLE and just add jsExceptionHere() where useful.

gfwilliams:
Thanks!

PR #1885, opened 2020-07-15
add 6x8 font

MaBecker:
there is enough space an ESP32 to have have 6x8 font


gfwilliams:
Seems like a good plan!

PR #1886, opened 2020-07-16
add jshSPISendMany

atc1441:
only TX support for now, size max 256 bytes at once.

this improves the sending of SPI data.


gfwilliams:
Other half of this is https://github.com/espruino/Espruino/pull/1887

Issue #1884, opened 2020-07-15
What about extend SPI with DMA to use TFC LCD's in unbuffered mode for all boards


MaBecker:
In times of 1bit or 2bit display standard SPI is expectable fast, but looking at those TFT LCD's like in Bangle.JS you like to use those kind of displays with other boards too.

Yes, there are already some modules available to use them with limitation in color because of using buffered mode. Yes, Bangle.JS is using eight GPIOS for data transfer and most of the other display do not offer this feature so they will never be as fast as Bangle.JS. But using DMA can make them much much faster.

**How could this be done ?**

* eg. add DMA feature to the existing SPI jswrapper and to the targets

     eg: SPI.setup({..... fast: true});

* or eg. come up with  a new lib for this, similar to neopixel and name it FastSPI

* or .......

**In my mind this will allow writing fast spi display modules base on the fast spi class** 
  
* keep the init sequence out of the firmware to handle any tft lcd

* add the init commands to an array in a [special way](https://github.com/espruino/Espruino/blob/master/libs/graphics/lcd_st7789_8bit.c#L28-L55) to easily decode and send  them 

* use require ('TFTLCD_Module').connect({<all the required pins and size as object>, [ array with init sequence]})

Would be great to have a module based implementation using on FastSPI.

Let me know how you think about this.

gfwilliams:
SPI should now already use DMA as far as I'm aware - maybe not on ESPxx but on nRF52 (and I think STM32) as well.

Honestly, the issue here is the JS execution speed. You're never going to have a fast SPI LCD if you have to call into JS for every pixel, so I think the `TFTLCD_Module` approach is needed.

There is already an SPI LCD driver for buffered mode that can be compiled in here: https://github.com/espruino/Espruino/blob/master/libs/graphics/lcd_spilcd.c

But I believe @fanoush has an SPI LCD driver for unbuffered mode, and maybe we could pull that in? I think he even had a version using Inline C?

However I'm not sure about having this built into every firmware - flash space is so tight on many platforms now I think this is something that'd need compiling in on request.

MaBecker:
Thanks - Yep, it is all about must have and nice to have functions and modules:)

I guess you mean this nice piece of code:

https://gist.github.com/fanoush/3dede6a16cef85fbf55f9d925521e4a0

I will soon have a TFT LCD standalone display in my hands to run some tests with all types of Espruino boards (Pico, Pixel, Wifi and MDBQ42T) and come back with some results.

[...]

PR #1887, opened 2020-07-16
add jshSPISendMany

atc1441:
only tx supported for now and no callback, Max 256 byte per request

this improves the sending of SPI data.

### DIFF
--- a/targets/esp32/jshardwareSpi.c
+++ b/targets/esp32/jshardwareSpi.c
@@ -81,6 +81,10 @@ Test with ILI9341 works, but could be faster.
 Espruino supports sendig byte by byte, no mass sending is supported.
 */
   
+  
+volatile spi_transaction_t spi_trans;
+volatile bool spi_Sending = false;
+
 /**
  * Initialize the hardware SPI device.
  * On the ESP32, hardware SPI is implemented via a set of default pins defined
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks, this looks good - do you think you'll be able to get RX and >256 byte done any time soon?

Personally I'm not sure I should merge this right now as it'll break SPI for anyone that's trying to receive data as well. You could always copy the 'weak' jshSPISendMany implementation from `jshardwarecommon.c` and then use that if there's an RX buffer or >256 bytes?

atc1441:
The new changes should now enable RX and up to 4096 bytes on TX.

I also tested the Callback but that did brake the sending of the new byte because of the way it handles the sending. and the end of the sending.

gfwilliams:
Looks great - thanks! Do you think you could just do

```
jshSPIWait(device);
if (callback) callback();
```

at the end?

That way even though sending in the background doesn't work, it doesn't matter. It'll just send in the foreground and call the callback when done? Then it doesn't break anything.

atc1441:
Good idea

gfwilliams:
Great! Are you happy that this is good to go? It looks fine to me. @MaBecker ?

MaBecker:
Yes good to go, because it is  already in use locally with a generic **tft lcd unbuffered spi display driver** started in #1884.

PR #1891, opened 2020-07-20
add lcd spi unbuffered driver

MaBecker:
required make options:
USE_LCD_SPI_UNBUF=1
DEFINES="-DSPISENDMANY_BUFFER_SIZE=128"

### DIFF
--- a/libs/graphics/lcd_spi_unbuf.c
+++ b/libs/graphics/lcd_spi_unbuf.c
@@ -130,22 +130,22 @@ JsVar *jswrap_lcd_spi_unbuf_connect(JsVar *device, JsVar *options) {
 }
 
 void disp_spi_transfer_addrwin(int x1, int y1, int x2, int y2) {
-  uint32_t wd;
+  unsigned char wd[4];
   x1 += _colstart;
   y1 += _rowstart;
   x2 += _colstart;
   y2 += _rowstart;
   spi_cmd(0x2A);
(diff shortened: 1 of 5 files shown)


MaBecker:
#1884

libs/graphics/lcd_spi_unbuf.c
```diff
@@ -0,0 +1,189 @@
...
+  jshPinOutput(_pin_cs, 1);
+  jshPinSetValue(_pin_cs, 1);
+  
+  lcd_spi_unbuf_setCallbacks(&gfx);
+  graphicsSetVar(&gfx);	
+  return parent;
+}
+
+void disp_spi_transfer_addrwin(int x1, int y1, int x2, int y2) {
+  uint32_t wd;
```
gfwilliams, line 133: This looks good, but wouldn't it be easier just to use a `unsigned char wd[4]` here? The whole shifting into a 32 bit number just seems a bit involved and hard to read?

atc1441: It would also work yes, 

it should also be possible to use an uint16_t wd[2] so no shifting is needed at all.

that way its even more readable. 

not tested right now.

sometimes you dont see it because you looked so long onto it^^

MaBecker:
Ok, changed and tested. So from my side we are good to go.

gfwilliams:
Great - thanks!

PR #1893, opened 2020-07-21
add jshSPISendMany()

MaBecker:


gfwilliams:
I'm not sure I understand this one... Didn't you just copy & paste the `jshSPISendMany` from `jshardware_common`?  `jshSPISendMany` is declared WEAK, so if it's not defined the one in `jshardware_common` is used automatically so this should have no effect whatsoever.

When you said you'd got it working I assumed you'd actually implemented a jshSPISendMany that called into the ESP8266 SPI send function :)

MaBecker:
Ups, I missed that declaration stuff .......

MaBecker:
But I added some jsWarn(".") to check if it is been used and got many "." printed when using the hardware SPI1.

MaBecker:
If adding `__attribute__((weak)) bool` is not what you mean, than please guide me to get this corrected as expected.

gfwilliams:
Yeah, that's not what I meant :)

My point is that your code is being used, but it seems identical to the `jshSPISendMany` in `jshardwarecommon.c` that is used by default anyway?

I don't believe there will be *any difference whatsoever* between this PR and what happens without it.

If you actually want a performance improvement you need to implement your own `spi_transaction` function that will take an array of data, and then call that from jshSPISendMany

MaBecker:
ok, will send a new pr :)

PR #1897, opened 2020-07-27
add jshSPISendMany

MaBecker:
calling spi_transaction() directly with 8 or 32 bits

Hope this is a implementation you have in mind :) 

please check and let me know

### DIFF
--- a/targets/esp8266/jshardware.c
+++ b/targets/esp8266/jshardware.c
@@ -937,7 +937,43 @@ void jshSPISet16(
   //os_printf("> jshSPISet16 - device=%d, is16=%d\n", device, is16);
   //os_printf("< jshSPISet16\n");
 }
+// ToDo: implement spi_transaction64 to handle 16 x 4 byte which is the max possible buffer to send in one call
+bool jshSPISendMany(IOEventFlags device, unsigned char *tx, unsigned char *rx, size_t count, void (*callback)()) {
+  size_t txPtr = 0;
+  size_t rxPtr = 0;
 
+  // transmit the data
+  if (!count%4) { 
+    while (txPtr<count && !jspIsInterrupted()) {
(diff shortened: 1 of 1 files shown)


gfwilliams:
Is the new code actually any faster than the default implementation?

I'd imagined that you would extend `spi_transaction` to take a big buffer of data (so it's not doing the peripheral setup for each byte). Often you can push a new byte to send in *before* reading the returned byte out, so you can have gapless SPI send. I bet there's a bunch of ESP8266 example code out there.

But even ignoring that, I think this code is broken. As far as I can tell, if you're sending a multiple of 4 bytes out, you only receive one byte for every 4 sent :(

MaBecker:
> Is the new code actually any faster than the default implementation?

Yes,  but not significantly as i thought.

> I'd imagined that you would extend spi_transaction to take a big buffer

I underestimated the preps in spi_transaction and thougth reducing the call by 3/4 would be a big improvement. 

used this snipped to  compare the difference of versions:

```
    t1=getTime(); 
    g.clear();
    t2 = getTime();
    console.log(t1-t2);
```

> I bet there's a bunch of ESP8266 example code out there.

That's what I though too, but  just found one, called [fast lcd driver](https://github.com/wdim0/esp8266_fast_lcd_driver_hspi) using all possible [16 dword register](https://github.com/wdim0/esp8266_fast_lcd_driver_hspi/blob/master/WLCD_for_NONOS_(tested_with_2_0_0)/driver/wlcd.c#L149)

> you only receive one byte for every 4 sent :(

Hmm,  easy to fix, but does not help much to improve speed. 

Next:
Will look forward to enhance the existing `spi_transaction` to take a big buffer of data.

gfwilliams:
Closing this one for now...

MaBecker:
Thanks, lost touch.....

PR #1907, opened 2020-08-12
Update jswrap_regexp.c

### LINKED ISSUE #1906: Regex doesn't recognize difference between `\b`
```js
>"b".replace(/[\b]/,(function(){return "q";}))
="q"
>"n".replace(/[\n]/,(function(){return "q";}))
="n"             
>"\b".replace(/[\b]/,(function(){return "q";}))
="\b"
>"\b".replace(/[b]/,(function(){return "q";}))
="\b"           
```
See example above. Not a huge deal, but breaks functionality for certain cases (ClojureScript).

kiranshila:
Added backspace escape character
Fixes #1906


gfwilliams:
Awesome, thanks!

PR #1945, opened 2020-10-12
disable DMA hack when SPI0_USE_EASY_DMA==0

fanoush:
ifdef is not enough as the value can be zero = DMA is disabled


gfwilliams:
Thanks! that was a stupid mistake :)

PR #1952, opened 2020-10-28
enable bootloader and DFU package build for SMAQ3

fanoush:
- add BOOTLOADER_SETTINGS_FAMILY variable for 52840 targets
- fix softdevice size in flash_available
- fix missing button 2 and 3s timeout in bootloader main


fanoush:
With this bootloader works for me on Q3 - it allows entering DFU mode when holding button and flashes DFU zip correctly

Without correct bootloader settings family the bootloader will not accept application hex as valid on 52840 device, surprisingly dfu package still can have `--hw-version 52` when generated so the family is needed only when flashing hex over SWD.

gfwilliams:
That's great - thanks!

PR #1958, opened 2020-11-16
move jshInitDevices() after GPIO clocks are enabled

### LINKED ISSUE #1957: STM32F4 pull up not set properly at boot - gpio clock not enabled yet when calling jshInitDevices()
I have [STM32F411CEU6 board](https://stm32-base.org/boards/STM32F411CEU6-WeAct-Black-Pill-V2.0.html) that needs "input_pullup" on button. Even with correct pull setting in board file it is not set at boot time.

`BTN1.getMode()` shows `"input"` instead of `"input_pullup"`. However after calling `reset()` it is correct. When debugging it looks like `jshPinSetState` correctly triggers write to  `GPIOx->PUPDR` but it has no effect. Looks like this line 
https://github.com/espruino/Espruino/blob/157c4c77ab11a4271a2207095e00084ff8a89f2e/targets/stm32/jshardware.c#L1105
is called too early as GPIO clocks are not yet enabled here. It helped me to move this line below, after clocks are enabled, over here https://github.com/espruino/Espruino/blob/157c4c77ab11a4271a2207095e00084ff8a89f2e/targets/stm32/jshardware.c#L1152

Seems to work for me but not sure if calling it later may not break something else. However I hope it is OK because `jshInitDevices()` is not platform specific so perhaps the STM32 platform should be initialized first before calling common code(?).

Found the solution here https://os.mbed.com/questions/79074/Nucleo-144-boards-and-GPIO-HAL-Not-all-p/ someone had similar issue and solution was to enable GPIO Clock first before setting up GPIO registers.

Not sending PR as maybe you  want to reshuffle it in different way, like moving only clock enabling code to the beginning of the method instead?

fanoush:
before GPIO clocks are enabled setting up GPIO registers inside jshInitDevices() has no effect (e.g. setting pull up on pin does nothing)
resolves #1957

### DIFF
--- a/targets/stm32/jshardware.c
+++ b/targets/stm32/jshardware.c
@@ -1102,7 +1102,6 @@ static void jshResetPeripherals() {
 }
 
 void jshInit() {
-  jshInitDevices();
   int i;
   // reset some vars
   for (i=0;i<16;i++)
@@ -1177,7 +1176,7 @@ void jshInit() {
   GPIO_Init(GPIOG, &GPIO_InitStructure);
 #endif
 #endif // ESPRUINOBOARD
-
+  jshInitDevices();
 #ifdef LED1_PININDEX
   // turn led on (status)
   jshPinOutput(LED1_PININDEX, 1);


gfwilliams:
Thanks!

PR #1965, opened 2020-12-05
Update SPI (fix #1963)

### LINKED ISSUE #1963: ESP32 hardware SPI receive broken since 2v06
See: http://forum.espruino.com/conversations/357245/#comment15677900

```
// Short D12 and D13 with a jumper

var s = new SPI();
s.setup({miso:D13, mosi:D12, sck:D14});
print(JSON.stringify(s.send("Hello World")));
// "Hello World"
print(JSON.stringify(s.send("Hello World")));
// "Hello World"

var s = SPI1;
s.setup({miso:D13, mosi:D12, sck:D14});
print(JSON.stringify(s.send("X")));
// "\xFF"
print(JSON.stringify(s.send("X")));
// "\x00"
print(JSON.stringify(s.send("XY")));
// "XY" <- actually works (seems it works if a multiple of 2 bytes?)
print(JSON.stringify(s.send("Hello World")));
// "Hello Worl\xFF"
print(JSON.stringify(s.send("Hello World")));
// "Hello Worl\x00"

// Trying to use SPI2 at this point causes a full reboot
// E (76511) spi_master: spi_bus_initialize(156): dma channel already in use

// But let's assume we don't use SPI1 so this
// actually works...

//var s = SPI2;
//s.setup({miso:D13, mosi:D12, sck:D14});
// print(JSON.stringify(s.send("XY")));
// "He"
// print(JSON.stringify(s.send("This is a test")));
// "Hell\u0000\u0000\u0000\u0000\u0000\u0000Hell"

// So it seems SPI2 is actually using SPI1's receive buffer?
```

This seem related to the SPI DMA modifications that went in after 2v06: https://github.com/espruino/Espruino/pull/1887

@MaBecker @atc1441 could you take a look at this please? This seems totally broken and I'd have thought even the quickest of tests would have shown it wasn't working correctly.

MaBecker:
* restore jshSPISend() version before 2v0.6
* move g_lastSPIRead to it's proper position
* reduced number of comments

### DIFF
--- a/targets/esp32/jshardwareSpi.c
+++ b/targets/esp32/jshardwareSpi.c
@@ -175,6 +175,7 @@ bool jshSPISendMany(IOEventFlags device, unsigned char *tx, unsigned char *rx, s
     if (count==1) {
       int r = jshSPISend(device, tx?*tx:-1);
       if (rx) *rx = r;
+      if(callback)callback();
       return true;
     }
 	jshSPIWait(device);
--- a/targets/esp32/jshardwareSpi.c
+++ b/targets/esp32/jshardwareSpi.c
@@ -163,7 +163,7 @@ int jshSPISend(
   } else {
     SPIChannels[channelPnt].g_lastSPIRead = (uint32_t)-1;
   }
-  retData = (int)SPIChannels[channelPnt].g_lastSPIRead;
+  int retData = (int)SPIChannels[channelPnt].g_lastSPIRead;
   return (int)retData;
 }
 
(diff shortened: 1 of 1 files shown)


MaBecker:
will let you know when it is time to merge, like to spend some more hours on testing.

gfwilliams:
Thanks! It may be you can call `spi_device_queue_trans` *only* when you have `rx==NULL` - so you get a fast hardware transmit for write only, but then use the slow one for reads?

MaBecker:
> It may be you can call spi_device_queue_trans only when you have rx==NULL - so you get a fast hardware transmit for write only, but then use the slow one for reads?

So let's wait until someone complaints .....

PR #1973, opened 2021-01-25
add check for LD_SRAM_SIZE symbol to nrf52_ble_espruino linker script, define for MICROBIT2 as 128KB

fanoush:
I was trying if I can push some parameters from board file to linker script and looks like it works. Could be used also for bootloader starting address in future so there can be e.g. same linker script for banglejs and others.

I can only see that resulting hex and binary looks like it has enlarged RAM (top of stack points to 0x20020000) but cannot test it in real device now. Also blindly enlarged variables by 64k/16=4096.

if not defined it goes to 0x20010000 so other devices should be fine as 0x20002c40+0xd3c0 = 0x20010000.

Can you eventually test it on yours? if not then no need to merge it for now.
I'll be getting my microbit v2 soon or maybe can borrow one from someone tomorrow.

### DIFF
--- a/boards/MICROBIT2.py
+++ b/boards/MICROBIT2.py
@@ -30,7 +30,7 @@ info = {
  'default_console_tx' : "H1",
  'default_console_rx' : "H0",
  'default_console_baudrate' : "9600",
- 'variables' : 2050, # How many variables are allocated for Espruino to use. RAM will be overflowed if this number is too high and code won't compile.
+ 'variables' : 2050+4096, # How many variables are allocated for Espruino to use. RAM will be overflowed if this number is too high and code won't compile.
  'binary_name' : 'espruino_%v_microbit2.hex',
  'build' : {
    'optimizeflags' : '-Os',
@@ -47,7 +47,8 @@ info = {
      'DEFINES += -DMICROBIT', # enable microbit-specific stuff
      'INCLUDE += -I$(ROOT)/libs/microbit',
      'WRAPPERSOURCES += libs/microbit/jswrap_microbit.c',
-     'DEFINES+=-DNEOPIXEL_SCK_PIN=27' # an unused pin
+     'DEFINES+=-DNEOPIXEL_SCK_PIN=27', # an unused pin
+     'LDFLAGS += -Xlinker --defsym=LD_SRAM_SIZE=0x20000' #
    ]
  }
 };
--- a/targetlibs/nrf5x_12/nrf5x_linkers/linker_nrf52_ble_espruino.ld
+++ b/targetlibs/nrf5x_12/nrf5x_linkers/linker_nrf52_ble_espruino.ld
@@ -3,10 +3,13 @@
 SEARCH_DIR(.)
 GROUP(-lgcc -lc -lnosys)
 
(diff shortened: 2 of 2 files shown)


gfwilliams:
Nice - thanks! You mean you could just define the start address, so the bootloader uses the same linker file as everything else?That could tidy things up a lot. 

I'm a bit busy now but I'll try and check it later today

fanoush:
> I'm a bit busy now but I'll try and check it later today

There is no hurry at all.

> You mean you could just define the start address, so the bootloader uses the same linker file as everything else

I meant that https://github.com/espruino/Espruino/blob/master/targetlibs/nrf5x_12/nrf5x_linkers/banglejs_espruino.ld could be merged with 'normal' app linker file `linker_nrf52_ble_espruino_bootloader.ld` and then the two bootloader linker files could be merged too - `banglejs_dfu.ld` vs `secure_dfu_gcc_nrf52.ld` , they both just has the difference resulting from bootloader start being 0x78000 vs 0x7a000

But true that more could be merged e.g. the 'app with bootloader'  `linker_nrf52_ble_espruino_bootloader.ld` vs 'app without bootloader' `linker_nrf52_ble_espruino.ld` is about LENGTH  of FLASH memory part.

I did not think about merging app linker files with bootloader linker files to have just one for everything - app and bootloader, that felt like too much but could be perhaps done too.

It is all tradeoff between complexity of this magic vs maintaining simpler duplicated linker files.

Apart from FLASH size the RAM size could be tuned like that too - like done here for microbit - then 52832 vs 52840 vs 52833 could use same file with just different sizes for ram set in board file. In fact ram size is already there but not used for this.

Again tradeoff between complexity and manual work, e.g. the MEM_SRAM_SD_OFFSET that I currently still hardcoded to 0x2c40 as it was before, could be in future derived  from softdevice version/configuration so e.g. having more board files or build configs for different MTU size or different number of connections would not need changing linker file as everything could be passed from board file or makefiles.

BTW, unrelated to this - for different MTU etc., I still think about better RAM layout for linker. For nrf it would  be better to have static data+bss not dependent on softdevice by being linked just after it, so the ultimate solution for me would be layout like: dynamic heap (=variables) first, then stack growing down and after stack the data+bss till the end of RAM. Then you could even have same binary and compute area for JS variables at boot time depending on what softdevice needs in particular configuration. We could even have some config saved to Storage like MTU size, # of connections and you could change it from JS code  and then call E.reboot() to reconfigure it. If storage is empty then use some defaults like now of course.

fanoush:
Oh you just did a (same) change across all linker files. That shows the benefit of having much less of them :-)  Redid same thing  in possibly better way for just the linker file in this PR to resolve conflict.  If you like this way it should be done in all others too or they could be merged to have this change only in one (per SDK).

gfwilliams:
This is great - works perfectly on MICROBIT2. Just need to pull it in for other devices.

I think I'll still likely need a per-SDK linker file as Nordic have been adding linker stuff to each SDK recently, but modding it wouldn't be too hard at all. I wonder whether we couldn't create the linker defines automatically based on linker specs

PR #1977, opened 2021-02-04
This is the fix for SMAQ3 external flash.

jeffmer:
I have tested that this builds ok, and I have tested the same code on my version of jsHardware.c which does not have touch buttons and it works fine. However, the health warning is that I have not actually tested this formeware build on my SMAQ3 as ity is not compatible with my current apps.


gfwilliams:
Thanks!

conoro:
Just confirming as an end-user that this works great. Thank you so much! Also, for fun, I loaded up the default apps for the Bangle.js and they all fit in flash for the first time. Whilst the screen scaling (and lack of buttons 😁) is problematic, the Welcome App, main menu, Settings and Morphing clock are all working. It's a huge leap forward for the SMAQ3.

gfwilliams:
Great! Actually if you tap the screen on the right hand side it should fake the 3 buttons :)

I think if I add a screen scaling mode we should be a bunch more apps being usable too

conoro:
Ah I think I accidentally did that a few times by tapping and swiping randomly :-) Yeah a bit of scaling would sort out a lot of them for sure. So excited to see this!

PR #1978, opened 2021-02-11
Adjust setLCDPower for SAMQ3 always on display

jeffmer:
Some simple changes so that setLCDPower only controls touch and backlight. g.flip() does not turn on display as it is always on and you do not want to turn on the backlight every time a minute hand moves in an always on display. To avoid burn in, the toggle of extcomm_in is not disabled by setLCDPower(0).

The pull also fixes the minor bug in the SMAQ3 board file 31->38.


jeffmer:
need to fix small bug

PR #1984, opened 2021-02-15
BMP280 pressure sensor bug fixes

jeffmer:
As promised - now seems to work well though although on of my smaq3 watches gives pressure readings that are more than 20hpa out. .


gfwilliams:
Great - thanks! I suppose 20hPa is only ~2% out? I don't know if that's within spec or not...

jeffmer:
When I have used bma280 sensors in the past, they have usually been + or -2hPa. I think they are individually or batch calibrated during manufacture - thus the plethora of coefficients.

PR #1993, opened 2021-03-18
Fix Python 3.9 compatibility

iYaroslav:
When trying to build a project, after calling `python scripts/hexmerge.py ...` an error like this appeared:

```bash
File ".../Espruino/scripts/intelhex/__init__.py", line 593, in write_hex_file
asstr(hexlify(bin.tostring()).translate(table)) +
AttributeError: 'array.array' object has no attribute 'tostring'
```
I just updated `intelhex` to version `2.3.0`, they already fixed that
https://github.com/python-intelhex/intelhex/commit/4125dce5b174401d38cc0fcf9d2e1aad07997f5e


gfwilliams:
That's great - thanks!

PR #2032, opened 2021-07-15
Exit if Ctrl+D on empty input line

FnControlOption:
(REPLs such as `node`, `python`, and `irb` support exiting by <kbd>Ctrl</kbd> + <kbd>D</kbd> on empty input line)

### DIFF
--- a/src/jsdevices.c
+++ b/src/jsdevices.c
@@ -411,10 +411,16 @@ static bool jshPushIOCharEventAppend(IOEventFlags channel, char charData) {
 
 /// Try and handle events in the IRQ itself. true if handled and shouldn't go in queue
 static bool jshPushIOCharEventHandler(IOEventFlags channel, char charData) {
-  // Check for a CTRL+C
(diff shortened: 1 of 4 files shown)


gfwilliams:
Sorry for the delay, I've just come back from holiday.

Does anyone else have thoughts on this? We could definitely add it, but I'd prefer the `#ifdef`s were done in such a way that it doesn't add extra (unused) code on embedded targets.

Ctrl-C does already exit if you just press it twice, so I guess this is more of a personal preference thing

jes:
It doesn't need to handle Ctrl-D explicitly, it just needs to exit when there's EOF on stdin. That's how most programs support "exit when you type Ctrl-D": it's just a consequence of exiting on end-of-file.

gfwilliams:
Ahh - that would actually be a perfect fix for this, and could just be put in `targets/linux/jshardware.c`

gfwilliams:
Fixed with #2131

PR #2042, opened 2021-08-29
step count final config X=6, RAW_THRESHOLD=14, high walking accuracy preferrable to ultra low step count when driving

hughbarney:


gfwilliams:
Thanks!

PR #2045, opened 2021-09-09
Added support for double and long touch gestures on Bangle2

coylen:
To minimise changes in interface this has added an extra "type" parameter to the existing touch event
 - 0 for standard click
 - 1 for double click
 - 2 for long click

Note: cannot currently get double click to fire


gfwilliams:
Great - thanks! odd about the double tap. Just a thought but do you think it means a tap with 2 fingers?

coylen:
I had it uploaded onto my watch and tried double click with single finger as well as any other interpretation I could think of, including tap with 2 fingers with no luck.

Can't find any authoritative datasheet on the chip to provide any guidance.

PR #2050, opened 2021-09-25
increase TX buffer to 256 for boards with RAM > 128KB

fanoush:
This helps on 52840 boards with figuring out MTU over 127 bytes. With current MTU 131 and 127 byte TX buffer the process.env is sent in two chunks, first having 127 bytes due to TX buffer limit so the MTU is guessed to be smaller by 1 byte (see WebIDE console log). With this patch it is guessed as 131-3=128 instead of 127. Not sure whether it is worth it but maybe yes?

### DIFF
--- a/scripts/build_platform_config.py
+++ b/scripts/build_platform_config.py
@@ -344,6 +344,7 @@ else:
   # TX buffer - for print/write/etc
   bufferSizeTX = 32 
   if board.chip["ram"]>=20: bufferSizeTX = 128
+  if board.chip["ram"]>=128: bufferSizeTX = 256
   bufferSizeTimer = 4 if board.chip["ram"]<20 else 16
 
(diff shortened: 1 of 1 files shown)


MaBecker:
What about the RX buffer?

MaBecker:
or bufferSizeIO?

fanoush:
good questions but I don't know,  bufferSizeIO is set already to 256 for nrf52 in that file. I'm even not sure this is worth it. However 52840 with 256KB can perhaps have a bit larger IO buffers if it could speed things up.

gfwilliams:
I forget but the RX buffer is already 256 x 4 bytes I think, which is the max? I have an open issue for it though as the RX buffer can definitely stored more efficiently than it is right now.

Thanks - good idea about the TX buffer!

PR #2051, opened 2021-09-25
fix Uint32Array/Uint24Array when SAVE_ON_FLASH is enabled

fanoush:
fix Uint32Array constructor when SAVE_ON_FLASH is enabled, line was switched with Uint24Array constructor so both worked in wrong way, note the Uint24Array class name :-)
```
>var a32=new Uint32Array(16)
=Uint32Array: {  }
>var a24=new Uint24Array(16)
=new À(16)
>a24
=new À(16)
```
after fix it is as expected, Uint24Array class and constructor is disabled, Uint32Array works
```
>var a24=new Uint24Array(10)
Uncaught ReferenceError: "Uint24Array" is not defined
 at line 1 col 13
var a24=new Uint24Array(10)
            ^
>var a32=new Uint32Array(10)
=new Uint32Array(10)
```


gfwilliams:
Great, thanks! Sorry, not sure how that got put in by mistake. Copy/paste error I guess!

PR #2052, opened 2021-09-26
nrf51 fixes

fanoush:
- compact variables broke cortex M0 build for me, I see hard fault at startup unless the array is aligned, on device with 16K RAM it picks up 12byte variables so otherwise it works, not sure if 32K can still be broken if variables are larger
- fds/fstorage takes quite some flash and even nontrivial amount of ram in nrf51 build and is not needed as peer manager is disabled, gives more variables or bigger stack
- with _disable_irg() it is quite unstable with heavy bt traffic, seen lockup below when doing NRF.setScan and filling array inside callback. According to https://devzone.nordicsemi.com/f/nordic-q-a/627/what-is-the-easiest-way-to-disable-all-interrupts disabling interrupts like this is asking for trouble, maybe same code would be same/better for nrf52 too?

```
(gdb) bt
#0  0xfffffffe in ?? ()
#1  <signal handler called>
#2  0x000000fe in ?? ()
#3  <signal handler called>
#4  0x0001be8e in jsvFreePtrInternal (var=var@entry=0x20002314 <jsVars+516>) at src/jsvar.c:548
#5  0x0001c020 in jsvFreePtr (var=0x20002314 <jsVars+516>) at src/jsvar.c:659
#6  jsvUnLockFreeIfNeeded (var=var@entry=0x20002314 <jsVars+516>) at src/jsvar.c:723
#7  0x0001c10a in jsvUnLock (var=var@entry=0x20002314 <jsVars+516>) at src/jsvar.c:735
#8  0x0001c0a6 in jsvFreePtr (var=0x20002260 <jsVars+336>) at src/jsvar.c:642
#9  jsvUnLockFreeIfNeeded (var=0x20002260 <jsVars+336>) at src/jsvar.c:723
#10 0x0001c10a in jsvUnLock (var=<optimized out>) at src/jsvar.c:735
#11 0x0002ecb8 in jsiExecuteEvents () at src/jsinteractive.c:1682
#12 jsiIdle () at src/jsinteractive.c:2204
#13 jsiLoop () at src/jsinteractive.c:2301
#14 main () at targets/nrf5x/main.c:37

(gdb) list *0x1be8e
0x1be8e is in jsvFreePtrInternal (src/jsvar.c:548).
543       jshInterruptOff(); // to allow this to be used from an IRQ
544       jsvSetNextSibling(var, jsVarFirstEmpty);
545       jsVarFirstEmpty = jsvGetRef(var);
546       touchedFreeList = true;
547       jshInterruptOn();
548     }
549
550     ALWAYS_INLINE void jsvFreePtr(JsVar *var) {
551       /* To be here, we're not supposed to be part of anything else. If
552        * we were, we'd have been freed by jsvGarbageCollect */
(gdb) list jshInterruptOff
66        \details Disables IRQ interrupts by setting the I-bit in the CPSR.
67        Can only be executed in Privileged modes.
68       */
69      __attribute__( ( always_inline ) ) __STATIC_INLINE void __disable_irq(void)
70      {
71        __ASM volatile ("cpsid i" : : : "memory");
72      }
73
```


fanoush:
BTW, I have few more ram/memory tuning fixes that helps me to trim down nrf51 build size and free some bytes of ram but is not nrf51 specific so it is better to create another PR later(?)
- disable USART code when board file has 0 uarts
- disable TWI code when board file has 0 i2c
- move exportPtrs array to flash, on nrf51 every byte of ram counts (I do have native code enabled and InlineC compiler working on nrf51, it still fits)

src/jsvar.c
```diff
@@ -46,7 +46,7 @@ unsigned int jsVarsSize = 0;
 unsigned int jsVarsSize = 0;
 JsVar *jsVars = NULL;
 #else
-JsVar jsVars[JSVAR_CACHE_SIZE];
+JsVar jsVars[JSVAR_CACHE_SIZE] __attribute__((aligned(4)));
```
gfwilliams, line 49: Does this align every array element, or just the start of the array? If it's the first then it's going to make compact_vars on every board useless.

fanoush: just the array (99%sure) you would need attribute the JsVar typedef itself to apply for each element, thin I tested this, not sure now

gfwilliams:
Thanks! I'd been hoping to track down the issue with setScan being unreliable on nRF51. `__set_BASEPRI(4<<5)` was an attempt at just disabling all but the critical softdevice stuff, but `sd_nvic_critical_region_enter` does look quite promising.

Honestly, it seems to work fine on nRF52 (I haven't seen any instabilities at all?) so I'm not sure about changing it if it's not broken!

Those other RAM/etc changes sound great - so yeah, another PR would be really good.

With the nRF51 hard fault, do you think it's that the M0 can't access unaligned memory? I'd posted above, if the line you've got there tries to align every element it's not something I could include, but we can always do `JSVAR_FORCE_16_BYTE` and maybe even have an `#if ... #error` to catch the nRF51 case.

src/jsvar.c
fanoush: checked with nrf52 build

```
200030b6 g     O .bss   00008039 .hidden jsVars
```
vs
```
200030b8 g     O .bss   00008039 .hidden jsVars
```

size still same 2039, not even divisible by 2

gfwilliams:
Ok, perfect - thanks for checking! I'm kind of surprised it was never aligned to be honest, but maybe it's `-Os`.

Thanks for all your work on this - merging now.

fanoush:
sd_nvic_critical_region_enter is there just for nrf51 now. It is actually inline code very similar to setbasepri, it is not svc call into softdevice  see https://github.com/espruino/Espruino/blob/master/targetlibs/nrf5x_12/components/softdevice/s132/headers/nrf_nvic.h#L425

fanoush:
> ! I'm kind of surprised it was never aligned to be honest

It was aligned before you merged the compact var branch, the JsVar typedef had the aligned attribute before so the array did not need it.

PR #2053, opened 2021-09-27
move exportPtrs to flash

fanoush:
can be seen that `process.env.EXPTR.toString(16)` now points to ram, with this patch it points to flash area. currently it saves 0x48 bytes of RAM. In flash it is free to grow in future.

I hope this stuff is supposed to be read only (?) as the entries itself point to flash anyway  - as seen e.g. on  peek32(process.env.EXPTR).toString(16)

### DIFF
--- a/src/jswrap_process.c
+++ b/src/jswrap_process.c
@@ -71,7 +71,7 @@ Returns the version of Espruino as a String
 #ifndef SAVE_ON_FLASH
 /* NOTE: The order of these is very important, as 
 the online compiler has its own copy of this table */
-const void *exportPtrs[] = {
+const void * const exportPtrs[] = {
     jsvLockAgainSafe,
     jsvUnLock,
     jsvSkipName,


fanoush:
I have verified that compiled javascript  with this even works on nrf51 (with EspruinoCompiler patch for cortex-m0)
Tried with
```
function foo(a,b) {
  "compiled";
  return a*53 + b*2;
}
```
So I hope the EXPTR is used in this case.

gfwilliams:
This is great - thanks! I obviously hadn't really thought about how `const` applies in this case. The idea was always for this to go in flash. I wonder if there are any other things like this?

fanoush:
> I wonder if there are any other things like this?

not in Espruino with SAVE_ON_FLASH however most nordic drivers allocate some static data and there is the standard newlib/gcc stuff that runs before the `main()` and that currently allocates impure_data array of 0x60 size

```
  2 .data         0000008c  20002c40  0004a530  00042c40  2**2
20002c40 l    d  .data  00000000 .data
20002cc8 l     O .data  00000000 __do_global_dtors_aux_fini_array_entry
20002cc4 l     O .data  00000000 __frame_dummy_init_array_entry
20002c40 l     O .data  00000060 impure_data
20002cc8 l       .data  00000000 __init_array_end
20002cc4 l       .data  00000000 __preinit_array_end
20002cc4 l       .data  00000000 __init_array_start
20002cc4 l       .data  00000000 __preinit_array_start
```
googled a bit and found this https://devzone.nordicsemi.com/f/nordic-q-a/1520/how-to-get-rid-of-impure_data but did not find a way to trim it down while keeping gcc/newlib startup stuff in.

However I finally gave up with the startup stuff and managed to trim it down by linking with -nostartfiles and renaming main() to _start() 

It crashed at startup first but then I let the targetlibs/nrf5x_12/components/toolchain/gcc/gcc_startup_nrf5x.S to clear BSS area (it is turned off by default) and now it seems it was the only thing missing and it looks working :-)

Since the `__init_array_start|end` appears to be zero there is hopefuly no library code (like printf, softfloats of whatever) that need initialization at startup.

PR #2054, opened 2021-09-27
allow zero usarts in board file

fanoush:
allow `'usart' : 0,` in board file, currently it break in several places when building

jsiHandleIOEventForUSART renamed to jsiHandleIOEventForSerial  and kept in as it looks not USART specific(?). If it is really USART specific then it should be #ifdef-ed too but doesn't look like it.

tested by building nrf51 and nrf52 with usart:0 and 1


gfwilliams:
Thanks!

fanoush:
no HW i2c no HW serial

STORAGE: 253952 -> 262144
CODE: 110592 -> 251832 (141240 bytes)
Code area Fits before Storage Area

with 1 usart added +3KB more

STORAGE: 253952 -> 262144
CODE: 110592 -> 255124 (144532 bytes)
==========================
 CODE AND STORAGE OVERLAP
   by 1172 bytes
==========================

with 1 usart + 1 i2c another 2KB

STORAGE: 253952 -> 262144
CODE: 110592 -> 257084 (146492 bytes)
==========================
 CODE AND STORAGE OVERLAP
   by 3132 bytes
==========================


also saves some bytes in RAM - 0x48 less in BSS with no i2c or serial

<   3 .bss          00001298  20002088  0003ecc8  00032088  2**3
---
>   3 .bss          00001250  20002088  0003d840  00032088  2**3


not actually tested with pure software I2C (or SPI) added back

gfwilliams:
This is great - thanks! I suppose the area reserved for the FD library is probably in the wrong place to be able to use for anything useful though? :(

I guess it's possible that Storage could be tweaked to allow it to work with non-contiguous pages though...

fanoush:
previous nrf51-fixes removed that storage but the board files actually didn't account for it, microbit1 has
```
  'saved_code' : {
    'address' : ((256 - 2) * 1024),
    'page_size' : 1024,
    'pages' : 2,
    'flash_available' : (256 - 108 - 2) # total flash pages - softdevice - saved code
  }
```
which I supposed is 2 blocks for Storage module? without PEER_MANAGER no code probably initialized the storage (?) but it was still linked in and took some ram (like ~0x70 bytes) and flash

gfwilliams:
Ahh, great! So on the microbit it's actually using the 2 pages reserved for FD for storage already. That makes sense thanks - I should have checked :)

PR #2056, opened 2021-09-27
allow no HW i2c for nrf5x

fanoush:
allow build with ` 'i2c' : 0,` for nrf5x

There already was something in common code so maybe other platforms allow it(?) but it still failed with nrf5x build.

Not actually sure about i2c slave case. It builds fine with i2c count 0 or 1 but I don't know how to enable I2C slave to test.
Not even sure I2c slave has sense to have without having i2c master enabled. I suppose yes as there are own headers and espruino has own device names so I tried those ifdefs to allow i2c slave without having master but it probably needs a test.

Also ifdef-ed "nrf_drv_spi.h" here like others. SPI was already possible to be build with zero count.

### DIFF
--- a/targets/nrf5x/jshardware.c
+++ b/targets/nrf5x/jshardware.c
@@ -72,13 +72,17 @@ void app_error_fault_handler(uint32_t id, uint32_t pc, uint32_t info) {
 #if USART_COUNT>0
 #include "nrf_drv_uart.h"
 #endif
+#if TWI_ENABLED
 #include "nrf_drv_twi.h"
+#endif
 #ifdef I2C_SLAVE
 #include "nrf_drv_twis.h"
 #endif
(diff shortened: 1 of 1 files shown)


fanoush:
btw tested nrf51 build for my board with only HW or SW I2c
```
no i2c CODE: 110592 -> 251832 (141240 bytes)

only sw i2c CODE: 110592 -> 253260 (142668 bytes)
BSS size 00001250

only hw i2c CODE: 110592 -> 253888 (143296 bytes)
BSS size 00001278
```
so if software i2c would work for microbit v1 (it is there for V2) it would save few bytes of flash and 0x28 bytes of ram = 3 variables :-)

gfwilliams:
Great - thanks! And yes, I imagine sw I2C should work on the micro:bit - I tend to use it often on other devices for accessing the hardware

PR #2058, opened 2021-09-28
use -nostartfiles for microbit1 to save some more RAM

fanoush:
saves 0x60 bytes of ram by removing newlib/gcc impure_data array
btw maybe this could work fine for whole nrf5x?

also bump storage up to 4 pages as recent nrf51 changes made code smaller
there is space for 5 pages now but maybe some extra code could be enabled later (arrow functions, watchdog,..)

### DIFF
--- a/boards/MICROBIT1.py
+++ b/boards/MICROBIT1.py
@@ -36,6 +36,8 @@ info = {
      'DEFINES+=-DSAVE_ON_FLASH_EXTREME',
      'DEFINES+=-DJSVAR_FORCE_NO_INLINE=1',
      'CFLAGS += -ffreestanding', # needed for SAVE_ON_FLASH_EXTREME (jswrap_math, __aeabi_dsub)
+     'CFLAGS += -D__STARTUP_CLEAR_BSS -DLD_NOSTARTFILES',
+     'LDFLAGS += -nostartfiles',
      'BLACKLIST=boards/MICROBIT1.blocklist', # force some stuff to be removed to save space
      'DEFINES+=-DCONFIG_GPIO_AS_PINRESET', # Allow the reset pin to work
      'DEFINES += -DMICROBIT', # enable microbit-specific stuff
@@ -48,6 +50,9 @@ info = {
    ]
  }
 };
+
+saved_code_pages = 4;
+
 chip = {
   'part' : "NRF51822",
   'family' : "NRF51",
@@ -63,10 +68,10 @@ chip = {
    # If using DFU bootloader, it sits at 0x3C000 - 0x40000 (0x40000 is end of flash)
    # Might want to change 256 -> 240 in the code below
   'saved_code' : {
-    'address' : ((256 - 2) * 1024),
+    'address' : ((256 - saved_code_pages) * 1024),
     'page_size' : 1024,
-    'pages' : 2,
-    'flash_available' : (256 - 108 - 2) # total flash pages - softdevice - saved code
+    'pages' : saved_code_pages,
+    'flash_available' : (256 - 108 - saved_code_pages) # total flash pages - softdevice - saved code
   }
 };
 
--- a/targets/nrf5x/main.c
+++ b/targets/nrf5x/main.c
@@ -40,3 +40,9 @@ int main() {
   //jsvKill();
   //jshKill();
 }
+
+#ifdef LD_NOSTARTFILES
+void _start(){
+  main();
+}
+#endif


gfwilliams:
Thanks! Do we know what that impure_data is? I wonder whether it's needed for C++ (tensorflow) or possibly maths, and it might cause build problems for other devices. But then if everything works ok I'm happy to switch over to it?

fanoush:
btw recent savings in RAM when compared to 2.10 - 0x180+0x68 = 384+104= 488 bytes = 40 variables :-)
currernt value of 350 in microbit built is quite optimistic but at least microbit1 gets more stack

My nrf51 build is back to 300 varaibles and listing storage via WebIDE works (when arrow functions are enabled)
```
RELESE_2V10 
$ grep "\.data.*00" espruino_2v10_microbit1.lst
  2 .data         00000080  20002008  0003f560  00032008  2**2
20002008 l    d  .data  00000000 .data
20002084 l     O .data  00000000 __do_global_dtors_aux_fini_array_entry
20002080 l     O .data  00000000 __frame_dummy_init_array_entry
20002074 l     O .data  00000004 rand_m_w
20002078 l     O .data  00000004 rand_m_z
20002008 l     O .data  00000060 impure_data
20002084 l       .data  00000000 __init_array_end
20002080 l       .data  00000000 __preinit_array_end
20002080 l       .data  00000000 __init_array_start
20002080 l       .data  00000000 __preinit_array_start
2000206c g     O .data  00000004 .hidden inputLineLength
20002008 g       .data  00000000 __data_start__
2000206a g     O .data  00000001 .hidden consoleDevice
2000207c g     O .data  00000004 SystemCoreClock
20002008 g       .data  00000000 _sdata
20002088 g       .data  00000000 __data_end__
20002070 g     O .data  00000002 .hidden m_peripheral_conn_handle
20002068 g     O .data  00000002 .hidden bleAdvertisingInterval
20002088 g       .data  00000000 _edata
$ grep "\.bss.*00" espruino_2v10_microbit1.lst
  5 .bss          000016fc  200020a8  0003f600  000320a8  2**3


LATEST

$ grep "\.data.*00" espruino_2v10.70_microbit1.lst
  2 .data         00000018  20002008  0003eaa8  00032008  2**2
20002008 l    d  .data  00000000 .data
20002014 l     O .data  00000004 rand_m_w
20002018 l     O .data  00000004 rand_m_z
2000200c g     O .data  00000004 .hidden inputLineLength
20002008 g       .data  00000000 __data_start__
2000200a g     O .data  00000001 .hidden consoleDevice
2000201c g     O .data  00000004 SystemCoreClock
20002008 g       .data  00000000 _sdata
20002020 g       .data  00000000 __data_end__
20002010 g     O .data  00000002 .hidden m_peripheral_conn_handle
20002008 g     O .data  00000002 .hidden bleAdvertisingInterval
20002020 g       .data  00000000 _edata
$ grep "\.bss.*00" espruino_2v10.70_microbit1.lst
  4 .bss          0000157c  20002030  0003ead0  00032030  2**3





```

fanoush:
> Do we know what that impure_data is?

it it related no non-reentrant functions in standard library, see e.g.

https://github.com/eblot/newlib/blob/master/newlib/libc/reent/impure.c#L23

it gets pulled in by gcc startup code, did not find way to avoid it
actually looks more like shutdown code - atexit - https://sourceware.org/legacy-ml/crossgcc/1999-q2/msg00018.html
https://e2e.ti.com/support/tools/code-composer-studio-group/ccs/f/code-composer-studio-forum/881297/ccs-msp430fr6989-gcc-produces-data-impure_data-in-ram-which-i-want-to-get-rid-off

gfwilliams:
Ok, nice, thanks for looking into it - sounds like it'd be pretty sensible to remove that for everything then?

fanoush:
> it'd be pretty sensible to remove that for everything then?

it is sensible to remove the impure-data if no reentrant newlib method is used, however the only way  I found is to remove gcc startup files via -nostartfiles which breaks any initialization https://gcc.gnu.org/onlinedocs/gccint/Initialization.html unless we reimplement part of it back. So if no library (like tensorflow) uses it (.init,.fini,.ctors,.dtors linker sections) it is OK. Not sure if you would get link errors or the library will silently not initialize when this is missing. However I guess checking size of those linker sections would tell if anything is there.

fanoush:
Tried this `-nostartfiles` with full nrf52 build. It seems to start and run fine and startup is skipped however it still does pull in the impure_data array becasue of global [errno](https://github.com/eblot/newlib/blob/master/newlib/libc/errno/errno.c) variable that gets set from `exp()` and `log()` math functions.
Looks like this is documented in both functions as  "If a domain error occurs, the global variable errno is set to EDOM". 

errno variable usage in math library is mentioned also here https://www.gnu.org/software/libc/manual/html_node/Math-Error-Reporting.html 

However  I have finally found pretty good workaround for this - providing replacement errno and exit() method , it didn't work before  and I always got newlib version linked in but after I removed extra `-lc -lnosys` from Makefile like this
```
 LDFLAGS += $(OPTIMIZEFLAGS) $(ARCHFLAGS) --specs=nano.specs
# -lc -lnosys
```
it now links fine and works :-) I'm not sure why you had this extra `-lc -lnosys` there, the `--specs=nano.specs` should already provide correct libraries according to the specs file and for me it links fine without it. Anyway with this removed and having simple stubs like below I can even link normal startup code in and the impure_data 0x60 byte structure in RAM is not pulled into build :-)

``` C
// __errno pointer getter is referenced from math library functions (exp,log) to set errno variable
int errno;
int *__errno()
{
  // non-reentrant version
  // original returns _impure_ptr->errno
  return &errno;
}

// exit is called from mainCRTStartup() when main() returns
void exit(int code){
  // non-reentrant do nothing version - we have nothing to clean up
  // original calls _impure_ptr->__cleanup
  while(1);
}
```
can make PR with this. Is it OK to put these stubs to the bottom of `targets/nrf5x/main.c`?
In fact the same should work also for other targets build by same toolchain - STM32 so maybe there is better place?
I'll try some STM32 build.

gfwilliams:
This looks great - thanks for digging in to this!

I think for now it might be better to just restrict this to nRF5x though to avoid any other risks? We can always look at pulling it in in the future.

Part of me wonders whether we shouldn't have a `src/fixup.c` kind of file with stuff like https://github.com/espruino/Espruino/blob/master/libs/math/jswrap_math.c#L24-L36 in it as well. I feel like there are a few cases where we can tweak/replace existing library functions to save a bunch of space.

fanoush:
> Part of me wonders whether we shouldn't have a `src/fixup.c` kind of file with stuff like

Yes, good idea, that's same type of stuff indeed.
BTW the `-ffreestanding` option you used there is interesting and a bit temperamental, it is used in microbit1 build and there it reduces size however I had it enabled for full nrf52 build together with -nostartfiles and the resulting code was actually larger than standard build. after removing it (while keeping -nostartfiles) it went down below the standard build size. Was checking generated code and it was memcpy() - `-ffreestanding` does not dare to optimize it away. without it it knows what memcpy is and can optimize several cases in nordic BLE libraries where memcpy copies just few bytes, there it is unrolled/inlined todirect mov instructions producing smaller code. in microbit1 case those are probably not included but I guess it allows to call this __aeabi_dsub instead of inlining longer code(?) so size goes down.

gfwilliams:
Ahh, that's very interesting!

Yes, `-ffreestanding` was literally just for `__aeabi_dsub`. It was such a big improvement it seemed worth it, but I wasn't aware of the `memcpy` optimisations. Makes sense though!

I wish we could replace `__aeabi_dsub` without  `-ffreestanding`, but I wasn't able to find a way :(

As an aside, so many of the double maths functions are absolutely massive when there are slower but massively smaller versions available. I had assumed that the `--specs=nano.specs` would enable something a bit more sensible, but it seems not

Issue #2057, opened 2021-09-27
possible deadlock in NRF.setScan callback


fanoush:
Hello, probably this is my bad code and already found way to avoid it, but here is something that appears to hang and never end
```
Remote debugging using :3333
0x000240d8 in jshTransmit (data=32 ' ', device=EV_SERIAL_DEVICE_STATE_START) at src/jsdevices.c:228
228         while (txHeadNext==txTail) {
(gdb) bt
#0  0x000240d8 in jshTransmit (data=32 ' ', device=EV_SERIAL_DEVICE_STATE_START) at src/jsdevices.c:228
#1  jshTransmit (device=EV_SERIAL_DEVICE_STATE_START, data=32 ' ') at src/jsdevices.c:175
#2  0x0002413c in jsiConsolePrintChar (data=<optimized out>) at src/jsinteractive.c:219
#3  0x00024160 in jsiConsolePrintString (str=0x200020a5 <buf+1> "") at src/jsinteractive.c:228
#4  0x00020a9c in vcbprintf (user_callback=0x24145 <jsiConsolePrintString>, user_data=0x0, fmt=0x3db46 "%s", fmt@entry=0x24145 <jsiConsolePrintString> "\265\004", argp=..., argp@entry=...)
    at src/jsutils.c:815
#5  0x00020c46 in cbprintf (user_callback=<optimized out>, user_data=<optimized out>, fmt=0x3db44 "%q%s") at src/jsutils.c:850
#6  0x00035336 in jsfGetJSONForObjectItWithCallback (it=it@entry=0x20003710, flags=flags@entry=(JSON_SOME_NEWLINES | JSON_PRETTY | JSON_SHOW_OBJECT_NAMES),
    whitespace=whitespace@entry=0x3dae2 "  ", nflags=nflags@entry=(JSON_SOME_NEWLINES | JSON_PRETTY | JSON_SHOW_OBJECT_NAMES | JSON_INDENT),
    user_callback=user_callback@entry=0x24145 <jsiConsolePrintString>, user_data=user_data@entry=0x0, first=<optimized out>, first@entry=true) at src/jswrap_json.c:280
#7  0x00034e38 in jsfGetJSONWithCallback (var=0x2000247c <jsVars+948>, varName=0x0, flags=<optimized out>, whitespace=0x3dae2 "  ", user_callback=0x24145 <jsiConsolePrintString>, user_data=0x0)
    at src/jswrap_json.c:450
[...]

PR #2059, opened 2021-09-28
nrf51 - do not care for proper irq on/off call nesting

fanoush:
do not care for balancing of calls and always try to enable interrupts in  jshInterruptOn
with previous code callling twice jshInterruptOff and then twice jshInterruptOn would left interrupts disabled as the region variable would stay at 1, see also #2057


gfwilliams:
Thanks!

PR #2083, opened 2021-10-29
Bangle.js: add `wrap` option for numerical E.showMenu items

rigrig:
Same as EspruinoDocs `graphical_menu.js`: https://github.com/espruino/EspruinoDocs/pull/614

I just realized maybe instead of 
```js
if (item.min!==undefined && item.value<item.min)
  item.value = (item.wrap && item.max!==undefined) ? item.max : item.min;
```
it would be better to just do
```js
if (item.min!==undefined && item.value<item.min)
  item.value = item.wrap ? item.max : item.min;
```
Either way your menu won't work properly if you specify `wrap` without also both `min` and `max`, but this way saves a few bytes.


gfwilliams:
Great - thanks! Good point about the rollover check - very happy to have that change if you want it in

PR #2085, opened 2021-11-01
E_showMenu improvements

rigrig:
Same as https://github.com/espruino/EspruinoDocs/pull/616

- remove unused variable
- remove unused parameter from `select()`
- shave a few bytes off `move()`

If a numerical item has `item.wrap` we no longer check if `item.min/max` is set,
because your menu will be broken either way.


rigrig:
Is `E_showMenu_F5.js` actually used? I noticed that the minified version isn't in the repo.

gfwilliams:
> Is E_showMenu_F5.js actually used?

Yes and no... There's a build file for the DT No.1 F5 watch, but I haven't actually built in some time.

Builds can (and sometimes do) use the unminified version if I'm not too bothered about space/speed.

I've just got 10 or 20 of the F5 sitting here from when I was considering it for Bangle.js 1, and I feel like at some point I should use them for something. But it's definitely super low priority so I wouldn't waste much time making improvements to the F5 showmenu :)

PR #2088, opened 2021-11-03
make jsfFindFile consistent with jsfCreateFile = always strip drive letter

fanoush:
also fix jsfCreateFile - shifting name duplicated last two letters for longest (28 chars) name, clear that space


fanoush:
As discussed in https://github.com/espruino/Espruino/pull/2079 it is preferred way to strip drive letter always so also any "x:file" becomes "file".

BTW I realized that JsfFileName union is passed by value (all 28 bytes) so actually I don't need to make copy as original caller value is not modified. This came as a surprise to me, passing such arrays by value does extra copies. Even the jsfNameFromString returns JsfFileName by value so the memory is copied twice and generated code is larger but I guess simplicity and readability is worth it.

gfwilliams:
Hi - thanks! Yes, the copy by value was intentional. It's not like it gets passed around much so it's not a performance impact, and as you say it's a little more readable. If it hits the flash memory usage hard it's something we could look at changing, but my gut feel was that it shouldn't.

PR #2091, opened 2021-11-06
jsfWriteFile - prevent duplicate file creation in internal bank when already present in external

fanoush:
-  when drive is specified (C:xxx) prefer existing file in different bank instead of creating duplicate in C:. Without this second file is created in internal flash when same name already exists in external flash. and list() gives two same files.
- also added methods for stripping drive letter from name and determining bank based on drive to simplify and remove duplicated drive letter code

with two banks jsfFindFile for "C:xxx" will search only C. File is not found when in external flash. I think it good to have it like that so that you can determine where the file really is by `require("Storage").read("C:xxx")` giving `undefined` when it is not in internal flash. However in jsfWriteFile it is better to relax this and use existing file in other bank to prevent file duplication.

feel free to shuffle/rename stuff if it not named correctly.

also alternative in jsfWriteFile could be to erase file in wrong location and let it create new one instead of using existing one but that felt more complicated to me. But maybe it would make more sense?

### DIFF
--- a/src/jsflash.c
+++ b/src/jsflash.c
@@ -484,8 +484,7 @@ char jsfStripDriveFromName(JsfFileName *name){
 void jsfGetDriveBankAddress(char drive, uint32_t *bankStartAddr, uint32_t *bankEndAddr){
 #ifdef JSF_BANK2_START_ADDRESS
   if (drive){
-    if (drive&(~0x20) == 'C'){ // make drive case insensitive
-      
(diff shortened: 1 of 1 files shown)


fanoush:
second idea - erasing file in wrong bank seem to be much better, this way file gets created where it belongs

gfwilliams:
Thanks! Yes, that sounds good.

PR #2099, opened 2021-11-19
stm32 allow no HW uart,spi,i2c

fanoush:
allow STM32 uart,spi,i2c count to be configurable including 0
tested on STM32F103 ($2 STLink V2 clone dongle with F103 inside)

BTW what is targets/stm32_ll ? looks like more or less a copy of targets/stm32 , is it still used?


gfwilliams:
Nice - thanks!

`targets/stm32_ll` is ST's attempt at porting to their new HAL layer for STM32L4. Rather than just modify the existing one they copied the whole thing, which obviously makes tweaking stuff a total nightmare now :( It'd be great to be able to merge back in but I'm unsure how hard that'd be

PR #2131, opened 2022-01-25
Exit on Ctrl-D (#2032)

jes:
This makes the REPL exit as soon as you type Ctrl-D, as per #2032.

Making it look for end-of-file is more annoying than expected since it puts the terminal in raw mode, and `getch()` currently does the wrong thing if `read()` returns 0.

I thought the simplest fix was just to explicitly check for Ctrl-D, but maybe something else is better?


gfwilliams:
Awesome, thanks! This is a *far* nicer solution to #2032

Issue #2104, opened 2021-11-29
Scroll indicator in menu shows up inconsistently on Bangle.js 2


mabuchner:
I've noticed, that the scroll indicator at the bottom shows up inconsistently on my Bangle.js 2 (2v10.232).

For example, when opening the settings, then the bottom scroll indicator is initially visible

![settings_0](https://user-images.githubusercontent.com/6006874/143897361-d50ab39e-1a5f-49f9-aa8b-059ef5b1a385.png)

The scroll indicator stays visible until the "Theme" option

![settings_1](https://user-images.githubusercontent.com/6006874/143897510-81f0f229-937d-4153-90a7-99beb4c38a67.png)

This kind of makes sense, since the list already arrived at the bottom.

However, when I scroll down to the next option, then the scroll indicator appears again.

![settings-2](https://user-images.githubusercontent.com/6006874/143897796-2d2129fa-12d4-4001-920d-72e01b5d28b4.png)

When I scroll up from this position, then the scroll indicator appears also for the "Theme" setting

![settings-3](https://user-images.githubusercontent.com/6006874/143898047-029d6581-8d03-4f6d-9a09-b56c05e5cf49.png)

mabuchner:
(Random side note: The battery percentage in the battery widget increased although the watch isn't charging)

jes:
https://github.com/espruino/Espruino/blob/master/libs/js/banglejs/E_showMenu_Q3.js#L74

This is because it decides whether the arrow needs to be drawn based on whether or not the index 1 more than the number of the last drawn row is less than the number of menu items, but the `draw()` function takes arguments to specify the min and max row to update, to save it from having to redraw every row.

So when you only ask it to redraw 1 row, it thinks the last row that was drawn was not the last menu item, which means there must be more menu items, so it draws the arrow. I'll try and work on a sensible fix.

jes:
I've fixed it here: https://github.com/jes/Espruino/commit/5794dc3d270c5f0f49234b1484291076c72e338f

But it probably also wants to be fixed in the other 2 implementations of E.showMenu(), and I am struggling with working out how to minify it.

gfwilliams:
Great! There's an update_js.sh file in the parent directory there. If you have EspruinoDocs installed next to Espruino it should minify for you - but I'm very happy to do that. I'm just merging that change in now and will do changelog + minified JS

gfwilliams:
[...]

PR #2134, opened 2022-01-28
Fix menu arrow behaviour on Bangle.js 1

jes:
This is the same as #2133 but for Bangle.js 1.

First reported in #2104.


gfwilliams:
Thanks!

PR #2135, opened 2022-01-29
Ad DC Filter to step counter

jeffmer:
As discussed extensively in the forum at [Better Pedometer - HELP NEEDED!](http://forum.espruino.com/conversations/359542/) and at @HughB 's suggestion this is to update `stepcount.c` with the DCFilter. No worries if you think this is premature and needs further testing.


gfwilliams:
Great - thanks!

PR #2159, opened 2022-02-23
Update E_showScroller.js

### LINKED ISSUE #2153: Bangle 1: E.showScroller() triangle drawn over top item
Using the sample [E.showscroller code from the documentation](http://www.espruino.com/Reference#l_E_showScroller), when run on the Bangle.js 1 emulator the top triangle is drawn over the list items.

When I originally posted about this, Gordon suggested it was intended to work with widgets. It does look better with widgets enabled, but that would mean the triangle overdraws the widgets instead?

andrewgoz:
Respect Bangle.appRect
Only draw arrows when needed
Prefer to draw selection in middle of screen
Implement scroll function
Pass selected flag to draw function
Make selection box drawing optional
Fixes #2153

### DIFF
--- a/libs/js/banglejs/E_showScroller.js
+++ b/libs/js/banglejs/E_showScroller.js
@@ -36,7 +36,7 @@ var s = {
             if (i==s.scroll) {
               g.setColor(g.theme.fgH)
                .drawRect(X,Y+y,w-1,Y+y+options.h-1)
-               .drawRect(1,Y+y+1,w-2,Y+y+options.h-2);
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks - although it looks like `scroll` is now a function? If you assigned that through then I don't think it'd work

andrewgoz:
For some reason my head said scroll needed to be a function. It's now back to being a property.

gfwilliams:
Thanks! You think this is ready for merging?

gfwilliams:
Great - thanks!

PR #2169, opened 2022-03-13
Bangle.js2: don't make setUI go "back" on button press

rigrig:
Not sure if this was the intended behavior, but I think it works better without it: I like having the "back" widget, but do want to use the button for something else.

### DIFF
--- a/libs/js/banglejs/Bangle_setUI_Q3.js
+++ b/libs/js/banglejs/Bangle_setUI_Q3.js
@@ -102,15 +102,11 @@
     var touchHandler = (_,e) => {
       if (e.y<24 && e.x<48) options.back();
     };
-    Bangle.on("touch", touchHandler);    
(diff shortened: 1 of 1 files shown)


gfwilliams:
Yes, this was the intended behaviour, as I've had several people (Bip users?) expecting that they can always press the button to go back.

I think longer term we should probably have a setting for this though as I think some people will also want to be able to do it via a swipe

rigrig:
>Yes, this was the intended behaviour, as I've had several people (Bip users?) expecting that they can always press the button to go back.

Well, I guess it is good to be consistent. And it does simplify things a bit to just never use the button as input if there is a back option.   
In that case, maybe it should not set a button watcher if `options.back` is present?

> I think some people will also want to be able to do it via a swipe

Yeah, I was actually already adding swipe left-ro-right for back myself, so that would be nice.

PR #2170, opened 2022-03-13
Bangle.js2: make setUI ignore touches in widgetbar

rigrig:
Mostly because the "back" button lives there.   
Maybe we should do it right and add checks for the entire `appRect` right away?

### DIFF
--- a/libs/js/banglejs/Bangle_setUI_Q3.js
+++ b/libs/js/banglejs/Bangle_setUI_Q3.js
@@ -38,7 +38,7 @@
       }
     };
     Bangle.on('drag',Bangle.dragHandler);
-    Bangle.touchHandler = d => {b();cb();};
+    Bangle.touchHandler = (_,e) => {if (e.y >= Bangle.appRect.y){b();cb();}};
     Bangle.on("touch", Bangle.touchHandler);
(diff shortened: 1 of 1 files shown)


gfwilliams:
Do you think this might make it harder to select some things? I think if someone's selecting something - especially in the launcher - near the top of the screen, the reported touch coordinates might actually go into the widget area?

Have you hit issues with pressing the back button actually firing an event off? Because maybe we should actually just try and get the back touch event handler executed first, and then ignore the touch event if it was pressed?

rigrig:
> Have you hit issues with pressing the back button actually firing an event off?

Yes, using code like this:
```js
Bangle.setUI({
    mode: "updown",
    back: goBack(),
  }, ud => {
    if (!ud) doSomething();
  });
```
Touching the back button fires both `goBack()` and `doSomething()`

We also have e.g. the messages widget listening for touches, so I figured excluding the whole bar would be best.

gfwilliams:
Thanks - I've just pushed in a slightly different fix for this - so if the 'back' handler is defined it now doesn't pass on touch events in other areas. I also made the active area for the back button a bit bigger so it's a bit easier to press (and harder to accidentally press a menu item when you want to go back)

PR #2173, opened 2022-03-18
Provide visual feedback on prompt button press

GrandVizierOlaf:
After the user presses a button, highlight the button briefly before closing the prompt and returning.


gfwilliams:
Nice idea - thanks!

PR #2177, opened 2022-04-03
Add support for __arm64__ (e.g. Apple Silicon) build

opichals:
Related to http://forum.espruino.com/conversations/373451/ discussion.

### DIFF
--- a/src/jsnative.c
+++ b/src/jsnative.c
@@ -24,7 +24,7 @@
   #error USE_CALLFUNCTION_HACK is required to make i386 builds work correctly
 #endif
 
-#if defined(__x86_64__)
+#if defined(__x86_64__) || defined(__arm64__)
     #define USE_SEPARATE_DOUBLES
 #endif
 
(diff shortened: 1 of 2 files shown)


gfwilliams:
Great - thanks!

For the [libs/tensorflow/third_party/gemmlowp/fixedpoint/fixedpoint.h](https://github.com/espruino/Espruino/pull/2177/files#diff-29e008ebfdca082bebf42d80819e89d125a67ef10902852801c95e511eab41e4) change, do you think there's another way around that (like trying to undef GEMMLOWP_NEON in the Makefile or some header that's not part of Tensorflow?).

I can just imagine that unless this is a change that's also in upstream Tensorflow, next time we pull the latest Tensorflow version it'll overwrite this.

opichals:
Added an empty `./libs/tensorflow/third_party/gemmlowp/fixedpoint/fixedpoint_neon.h` which should not give you any trouble when upgrading later on.

gfwilliams:
Excellent - thank you for sorting this so quickly!

PR #2179, opened 2022-04-08
Extend setWatch to allow buffering

mikabytes:
Hi. When receiving many interruptions on a pin, the ESP8266 can't handle it. More often than not, it will fire an exception FIFO_FULL. The callback can't do anything useful in the short time available, not even adding the value to an array or print it to screen.

 I have tried increasing the `bufferSizeIO` in `scripts/build_platform_config.py` to 512. That works for a while, but then random errors start happening. I'm guessing there's some overlap issue.

So I resolved to attack this problem from another angle. Changing the signature of setWatch to avoid having to go into javascript:

```
setWatch(Array, Pin, options)
```

Instead of executing a JS callback, just add the event to the provided array. This should work, I think. But I'm having some issues with the code. I get `Execution Interrupted` on the first interruption, and then the chip is broken until I re-flash it.

First guess is that there needs to be some locking, or unlocking, whatever that is? I see it all over the code... Any help on this code would be appreciated. Thanks!

### DIFF
--- a/src/jsinteractive.c
+++ b/src/jsinteractive.c
@@ -2031,7 +2031,10 @@ void jsiIdle() {
                 if (jshIsPinValid(dataPin))
                   jsvObjectSetChildAndUnLock(data, "data", jsvNewFromBool((event.flags&EV_EXTI_DATA_PIN_HIGH)!=0));
               }
-              if (!jsiExecuteEventCallback(0, watchCallback, 1, &data) && watchRecurring) {
+              if (jsvIsArray(watchCallback)) {
+                // add this event to supplied array instead of using callback
+                jsvArrayAddToEnd(watchCallback, data, 1);
+              } else if (!jsiExecuteEventCallback(0, watchCallback, 1, &data) && watchRecurring) {
                 jsError("Ctrl-C while processing watch - removing it.");
                 jsErrorFlags |= JSERR_CALLBACK;
                 watchRecurring = false;
--- a/src/jswrap_io.c
+++ b/src/jswrap_io.c
@@ -729,8 +729,8 @@ JsVar *jswrap_interface_setWatch(
     repeat = jsvGetBool(repeatOrObject);
 
   JsVarInt itemIndex = -1;
-  if (!jsvIsFunction(func) && !jsvIsString(func)) {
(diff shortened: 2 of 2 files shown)


gfwilliams:
Hi - what are you trying to receive with this? Is it continual data, or just a burst?

When this works, I guess it could be handy but my concern would be that this is only really a stopgap...

If data is coming in so fast that you get 256 events in a short time period, you may not be able to process them fast enough anyway. ESP8266 only has 1600 vars available so one event object may = maybe 7 vars? That's only another 200 events you're able to store assuming you had all your RAM free to start with.

> I have tried increasing the bufferSizeIO in scripts/build_platform_config.py to 512

Was this a recent version? I only fixed it maybe a month ago so you could have sizes >256. However you may need to decrease the amount of JsVars available so you still have enough execution stack.

mikabytes:
Hi @gfwilliams 

Thanks for getting back to me.

For this, I'm following the Infrared Remote Control tutorial. But I've had this issue in several other projects. https://www.espruino.com/Infrared+Remote+Control It is likely more than 200 events.

So, yes, the fix would only be viable where bursts of events are concerned. Continual data at that rate would not be possible.

It doesn't error if the watch does very light work, such as assigning a single variable some value. It cannot `console.log` things or place the events in an array. So, I conclude that the watch implementation is fast enough, but the JS stuff is too slow to keep up.

Hmm, so if I understand, if I were to be able to fix the speed issue, I would run into a memory issue instead? Hmm.. not too good. 

> Was this a recent version?

I built it from master branch, last commit is showing Tue Apr 5. I see some commits from Apr 21 related to io_buffer_size, so perhaps it may be worth to give that another try? I'll see if I can find some time this weekend.

gfwilliams:
>  I see some commits from Apr 21 related to io_buffer_size, so perhaps it may be worth to give that another try?

Yes, please can you give that a go? Check out the commits - you just add a line to the `BOARD.py` file. 

Basically I thought it was in `master` branch, but it wasn't. I pulled it across later but then I didn't get around to updating you in this post.

Before, it 100% would have failed above 256. After those commits you should be fine as long as it's a power of 2.

gfwilliams:
Closing this for now as I think with that recent change this should now be fixed

PR #2182, opened 2022-04-19
[E_showMenu_Q3] Recheck string wrapping after font change

storm64:


gfwilliams:
Thanks! That's great!

PR #2183, opened 2022-04-27
Bangle.js1: show back button for "< Back" item in E.showMenu

rigrig:
Simply by pass it on to setUI.
This also works for options.back, because that adds a "< Back" item.

This should be a simple 1-line (151) change, but looking at the diff:
1. Apparently my editor stripped out end-of-line whitespace
2. The minifier picked a different variable name (`l`->`k`)

If that's a problem I'd be happy to clean it up a bit.


rigrig:
PR #2184 also includes this

gfwilliams:
Thanks!

PR #2184, opened 2022-04-27
Bangle.js1: add submenu to E.subMenu

rigrig:
This also merges in #2183, making the back button work correctly in submenus.

Minor flaw: If the main menu has no back button, the submenu doesn't have one either.

Note: I only tested this with a `E.showMenu=...` `boot.js` file, not by updating my firmware.


gfwilliams:
What do you mean by 'submenu'? Do you have any pics/info on what this change is?

If it's what I think it it, my concern is I'm not sure it it really helps people much?

The submenu on the Bangle.js 2 was done so that people could just 'tap' the correct menu item, but since we're using buttons on Bangle.js 1 I'm not sure if that's helpful?

rigrig:
> Do you have any pics/info on what this change is?

Whoops, sorry. I'll add some images when I get home tonight.

For items with a numeric value (and `min`,`max` constraining it to less than 20 choices), it opens a submenu with available options.
Not that big a deal for "real" numeric values, but if they represent choices (like quiet mode or theme colours), it means you get to see the list instead of having to discover choices one item at a time.
Actually, it might be better if it only uses this if the item has a`format`, and for values just keeps the in-place editor.

rigrig:
From the settings app:
System->Theme->Customize->Foreground
screenshot (1)
Alerts->Beep
screenshot (2)
Alerts->Quiet Mode
screenshot (3)

I also updated the PR to only do this for items that have `format()`, and `min=0,step=1`, on the assumption that those are likely to be the index into an option list.

PR #2185, opened 2022-04-27
setUI: don't redraw widgets when replacing 'back'

rigrig:
On the Bangle.js 1 at least, this makes navigating settings much smoother, because otherwise `Bangle.drawWidgets()` happens twice every time it shows a new menu. (Well, provided your [E.showMenu shows back buttons](https://github.com/espruino/Espruino/pull/2183))

Only tested for Bangle.js 1 with `Bangle.setUI=...` in a `boot.js` file, not actual firmware **or on a Bangle.js 2**.


gfwilliams:
Thanks!

PR #2186, opened 2022-04-27
[E_showMenu_Q3.js] Add option to prevent scroller

storm64:
~~Prevent using E.showScroller() on wraping items.
This adds the option to control if the scroller or the simple box should be displayed.
In my opinion there is no need for a separate property, because the item.wrap property is not used inside E.showScroller().~~

This adds a new ".noList" property to control if the simple box instead of the scroller/list should be displayed inside E.showMenu().


gfwilliams:
Did you have a specific reason for this change? Which app isn't working as you'd expect?

Because I'm not sure it's going to be great for most people. I just did a quick search in the repo and one of the first hits is `qmsched` with `min:0, max:2, wrap: true,`. You're saying that now, instead of popping up a menu with 3 menu items, this is only going to allow folks to scroll between items.

Same for antonclk where we're choosing between `"Long", "Short", "ISO8601"`. It doesn't seem like an improvement... In fact I'm struggling to see *any* cases where wrap=true and there are less than 20 menu items where the scrolling menu wasn't fine.

gfwilliams:
Which specific apps did you have a problem with?

Maybe you could change the code so that if there's no `format` - eg it's pure numeric value - then it doesn't pop up the selection window?

storm64:
> Which specific apps did you have a problem with?

Its not specially one app, but I think that it would be nice to have the possibility to force the boxed design.

I tried my code on some menus and understand your concerns. I changed the code to use a separate property to control this.

gfwilliams:
Thanks! I'll add the same for the recently added F18 code

PR #2187, opened 2022-04-29
[locale] Fix default dow and month functions

alessandrococco:
I'm testing the new `date_utils` module and I noticed some wrong outputs.

With the default hard-coded locale:

```
>require("locale").dow({ getDay: () => 1 }, 1)
="[ob" <--- NOT OK! :-(
>
```

After installing the `EN_GB `locale:

```
>require("locale").dow({ getDay: () => 1 }, 1)
="Mon" <--- OK :-D
>
```

After some digging I found the issue in the `locale.js`: the hard-coded functions call the `toString()` but if `d` is not a `Date` then the output is not what we expected.

The same bug exists with `month()`.

### DIFF
--- a/libs/js/banglejs/locale.js
+++ b/libs/js/banglejs/locale.js
@@ -22,8 +22,8 @@ exports = { name : "en_GB", currencySym:"£",
       return (" "+h).substr(-2)+":"+("0"+m).substr(-2)+"."+("0"+d.getSeconds()).substr(-2)+" "+r;
     }
   },
-  dow : (d,short) => short?d.toString().substr(0,3):"Sunday,Monday,Tuesday,Wednesday,Thursday,Friday,Saturday".split(",")[d.getDay()], // Date to "Monday" or "Mon"(short)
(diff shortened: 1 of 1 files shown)


gfwilliams:
Great - thanks for this! Not sure how you feel but do you think the library needs to be able to work on 2v13 and previous firmwares?

gfwilliams:
Unfortunately I don't think that workaround will work properly. It's fine for English, but other translations use different words/lengths. For example german: `abday: "So,Mo,Di,Mi,Do,Fr,Sa",`

You could check `require("Storage").read("locale")!==undefined` before applying the workaround so it only works for the built-in function? Or I guess you could pass in `new Date((day+3.5)*86400000);` instead of overriding `getDay`?

gfwilliams:
`new Date((month-0.5)*2628000000);` for month too?

alessandrococco:
I sent the PR! https://github.com/espruino/BangleApps/pull/1773 :-)

PR #2188, opened 2022-04-29
[E_showMenu_Q3] Add "Back" on submenus

storm64:
1. Add "Back"-Button on submenus inside E_showMenu
2. Regenerated E_showMenu_Q3.min.js via IDE with "Closure  Simple Optimizations"
3. Add ChangeLog entry
4. Readd ChangeLog entry for https://github.com/espruino/Espruino/pull/2186


gfwilliams:
Thanks! Please can you do the minification using the `minify.js` in EspruinoDocs? There's a script `update_js.sh` in `libs/js`. It just uses an offline version of closure I know works well so there's no risk of stuff breaking because the closure version changed :) Also I generally add the newest ChangeLog entries at the bottom of the list :)

PR #2199, opened 2022-05-21
[Bangle.setUI] Add "clockcustom" mode

storm64:
Think this might be useful for more complex clocks.
For now it would be necessary to set `Bangle.CLOCK = 1` manually after a custom setUI to show other apps that a clock is used.

Is it possible for me to edit the http://www.espruino.com/Reference#l_Bangle_setUI to help via a PR, too?

Edit to 531c08e: The .min.js need to be regenerated...


gfwilliams:
Hi - yes, that could be a good idea... What are your thoughts on:

```
Bangle.setUI({type:"clockcustom", ...
Bangle.setUI({type:"custom", clock:1, ...
```

I don't know if we care but `Bangle.setUI({type:"clockcustom",` would throw and exception if we ran a 'new' clock on a firmware without the updated setUI, but doing the other way would at least silently fail.

To edit the docs, look in `libs/bangle/jswrap_bangle.c` - the docs are all there in the `JSON` style comments

storm64:
I like the approach with an additional argument specifying the clock status.

Will have a look into this in the next days and update the .min.js and docs according to your suggestion.

gfwilliams:
Great, thanks! Don't worry too much as I can do the minified files here - I use the minifer from EspruinoDocs so I know the code will turn out more or less the same each time :)

PR #2201, opened 2022-05-24
Basic support for numeric separator

mariusGundersen:
This is an initial working version, but it has some known bugs:

* parseInt("1_0") returns 10, not 1. That's a breaking change
* multiple separators in a row is supported, but that's not supported according to the spec

I'm open for suggestions for the first issue. It seems like either a new parse method is needed (which increases the codesize) or a new argument needs to be passed in to the existing one (which might be a breaking change).

For the second one I'm not sure it's worth it to fix this


gfwilliams:
Hi! Just looking at this, and I see we've now got:

```
while (isNumeric(lex->currCh) || (!canBeFloating && isHexadecimal(lex->currCh)) || lex->currCh=='_') {
    jslTokenAppendChar(lex->currCh);
    jslGetNextCh();
  }
```

What about doing:

```
while (isNumeric(lex->currCh) || (!canBeFloating && isHexadecimal(lex->currCh)) || lex->currCh=='_') {
    if (lex->currCh!='_') jslTokenAppendChar(lex->currCh);
    jslGetNextCh();
  }
```

Would that not sort this out without requiring a change to stringToIntWithRadix (which I guess might break other things)?

gfwilliams:
Great - thanks for this!

Issue #1302, opened 2018-01-09
ES5/ES6 wishlist


gfwilliams:
ES5/ES6 features that aren't in Espruino but that would be useful... I'll do one post per feature, and please use the `add reaction` button to vote on each one.

gfwilliams:
- [ ] `String.startsWith/endsWith/includes` - would be trivial to add

gfwilliams:
- [ ] Proper `let` implementation - it's just treated as `var` at the moment, but nobody seems to actually care

gfwilliams:
- [ ] Proper `const` implementation - it's just treated as `var` at the moment

gfwilliams:
- [ ] `Array.find/findIndex` - find by executing a function for comparison

gfwilliams:
- [ ] JS classes - probably not too painful, and seems like it'd tidy up a lot of code

gfwilliams:
- [ ] Default argument values

gfwilliams:
- [ ] `Array.includes` - easy, just `indexOf()>=0`

gfwilliams:
- [ ] `Object.entries` and `Object.values`

gfwilliams:
- [ ]  `String.padStart/padEnd`

gfwilliams:
- [ ] `async/await` - although this it likely too difficult to implement in Espruino as it is

gfwilliams:
- [ ] getters and setters

FlowerOfLife:
Arrow functions to bind "this" PLEASE

gfwilliams:
- [ ] [Destructuring](https://github.com/lukehoban/es6features#destructuring)

gfwilliams:
- [ ] [Rest and Spread](https://github.com/lukehoban/es6features#default--rest--spread)

jjok:
`import` and `export`?

vshymanskyy:
Ok, I was actually able to get `async/await` working on Espruino (using Babel transpiler and some hacking).
**Here is a full example:  https://github.com/vshymanskyy/espruino-await**
Looks pretty usable, but a native implementation would be so much better ;)
What do you think?

louisvangeldrop:
Webassembly support?

andrewwakeling:
I did a similar exercise to @vshymanskyy and got various modern JS features working using Webpack/Babel. See: https://github.com/andrewwakeling/espruino-webpack-babel-sample

[...]

PR #2202, opened 2022-05-24
Added support for method declaration in objects

mariusGundersen:
This makes it possible to write 

```js
var obj = {
  method(a){
    return a*5;
  }
}
```

### DIFF
--- a/src/jsparse.c
+++ b/src/jsparse.c
@@ -1295,7 +1295,13 @@ NO_INLINE JsVar *jspeFactorObject() {
         }
       } else
 #endif
-      {
+      if (lex->tk == '(') {
+        JsVar *contentsName = jsvFindChildFromVar(contents, varName, true);
+        if (contentsName) {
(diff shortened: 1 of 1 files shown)


mariusGundersen:
As wished for here: https://github.com/espruino/Espruino/issues/1302#issuecomment-389628601

gfwilliams:
Wow, that's excellent - thanks! Nice clean addition!

gfwilliams:
Good point, thanks! Yes, I think that was left over from before `jslIsIDOrReservedWord`. Doesn't seem to make it fail any of the built-in tests, and `{true:2}` works anyway.

I'll just put that in. I've just surrounded your code by a `#ifndef ESPR_NO_OBJECT_METHODS` as for microbit 1 where we're really low on flash, that (albeit small) addition along with the const/let stuff I added pushes it over the edge so it no longer fits. Otherwise it's enabled on pretty much every other build.

PR #2203, opened 2022-05-25
Implemented property shorthand

mariusGundersen:
As wished for in https://github.com/espruino/Espruino/issues/1302#issuecomment-389628601


gfwilliams:
Nice - thanks!

gfwilliams:
Thanks for adding the ESPR_NO_ check too! It's a pain we have to do it, but some devices are just really tight for memory.

If you manage to find more redundant checks then that'll help though ;)

Just one thing - please can you put a note in the ChangeLog too? I did it this time, but if it's in the PR it makes things tidier :)

PR #2204, opened 2022-05-25
Nullish coalescing operator

mariusGundersen:
This adds support for the nullish coalescing operator:

```js
var a = (null ?? true);
```

It is similar to doing `a || b` but it only uses `b` if `a` is null or undefined.

### DIFF
--- a/src/jslex.c
+++ b/src/jslex.c
@@ -112,6 +112,7 @@ typedef enum {
   JSLJT_NUMBER,
   JSLJT_STRING,
 
+  JSJLT_QUESTION,
   JSLJT_EXCLAMATION,
   JSLJT_PLUS,
   JSLJT_MINUS,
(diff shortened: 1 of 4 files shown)


gfwilliams:
Thanks! This looks great - BUT...

In the Web IDE and the Bangle.js App Loader we do 'pretokenisation' - basically detecting reserved words/tokens and converting them to the token itself - which reduces memory usage and increases execution speed.

The gotcha is that each of those has a list of tokens and their values - so when you add LEX_NULLISH it changes all the enum values, and the pretokenised code no longer works. For instance updating any Bangle to the new firmware would likely stop all the apps from working.

In retrospect I should have left some padding for new operators/symbols so it didn't matter, but I didn't :( So I guess maybe to make this backwards compatible we'd need to create a new list (`_LEX_OPERATOR_START2`) after the symbols - this time with some padding - and put `LEX_NULLISH` in there

gfwilliams:
I'll see about tweaking this now though...

gfwilliams:
Just sorted I think - thanks!

mariusGundersen:
I'm planning to add the null propagation operator `?.` and the rest/spread operator `...` as well, so there might be more issues like this. Where is the pretokenization done? How can I work around it?

gfwilliams:
It should be fine after https://github.com/espruino/Espruino/commit/aa8167b2477d0e67347f6c9f17971a671b38a1c9 - just stick them after LEX_NULLISH and you should be good now

PR #2210, opened 2022-05-29
Issue-2209 Make BangleJS en_GB local distance purely imperial

rjaduthie:
Stop strange behaviour where metres are used < 1km and then switching to miles.


rjaduthie:
**TODO**: Minify the modified JS

rjaduthie:
> **TODO**: Minify the modified JS

DONE.  

I used the command in the `update_js.sh` script - i.e. clone the EspruinoDocs repo and used: 
```bash
node ../../../EspruinoDocs/bin/minify.js banglejs/locale.js banglejs/locale.min.js
```

rjaduthie:
TODO: Any tests for this?

rjaduthie:
TODO: Update CHANGELOG

gfwilliams:
Just checking, but what is the issue you are wanting to fix with this?

* That you don't like the mix of meters and miles
* That you don't like the switchover to miles at 999m?

Because I'm pretty sure I'll get complaints when the default distance changes to yards. While for navigation it makes sense, things like the height above sea level probably still makes sense in meters, not yards even in England :)

The 999m changeover was actually an attempt at changing over when the value got above 3 characters long - but maybe changing over at one mile does make more sense.

Or maybe it's actually more useful to a greater proportion of users if we just move to kilometers by default?

rjaduthie:
> Just checking, but what is the issue you are wanting to fix with this?
> 
>     * That you don't like the mix of meters and miles
> 
>     * That you don't like the switchover to miles at 999m?
> 
> 

The mixture is the main issue, for me: I'm not the only one - cf., linked issue in BangleApps: https://github.com/espruino/BangleApps/issues/1824

The change I've proposed is to make the distance consistent with its unit; I figured that if we're going with miles, then yards is the logical choice.

> Because I'm pretty sure I'll get complaints when the default distance changes to yards. While for navigation it makes sense, things like the height above sea level probably still makes sense in meters, not yards even in England :)

Should `height` be a different metric? ...as you point out, it _is_ distinct from `distance`!  You're absolutely right that yards is a bad unit for altitude, I would say the natural choice for imperial units would be **feet**. 

> The 999m changeover was actually an attempt at changing over when the value got above 3 characters long - but maybe changing over at one mile does make more sense.
> 
> Or maybe it's actually more useful to a greater proportion of users if we just move to kilometers by default?

Kilometres is what I would personally choose to show on my watch for running, but I'm pretty sure there's a case for sometimes having miles.  Anecdotally, I have a friend whose an ultra-marathon runner and a cyclist, they use km for the former and miles for the latter.

How about...

- An additional property handles height - e.g. `altitude`, as this may be in a different unit to distance - e.g. in feet.
- Large numbers of feet can be expressed as (say) 10k5 ft.
- The default GB locale is switched to feet, yards & miles.
- To cater for the choice within an app for imperial or metric units, the `exstats` module offers non-locale representations of distance - e.g.  `distance imp.` & `distance met.` - in addition to the existing `distance` which gives the locale representation.

...?

PR #2221, opened 2022-06-10
Optional chaining

mariusGundersen:
This adds support for the three types of optional chaining:

```js
a?.b // returns b unless a is nullish
a?.() // calls the function a unless a is nullish
a?.[0] // returns the first item in the array a unless a is nullish
```

### DIFF
--- a/src/jsparse.c
+++ b/src/jsparse.c
@@ -1094,6 +1094,7 @@ NO_INLINE JsVar *jspeFactorMember(JsVar *a, JsVar **parentResult, bool *isOption
         }
         // skip over current token (we checked above that it was an ID or reserved word)
         jslGetNextToken();
+#ifndef ESPR_NO_OPTIONAL_CHAINING
       } else if ((lex->tk == '(' || lex->tk == '[') && optionalTk) {
         // handle a?.() and a?.[0]
 
@@ -1106,6 +1107,7 @@ NO_INLINE JsVar *jspeFactorMember(JsVar *a, JsVar **parentResult, bool *isOption
           a = aVar;
         }
         continue;
+#endif
       } else {
         // incorrect token - force a match fail by asking for an ID
         JSP_MATCH_WITH_RETURN(LEX_ID, a);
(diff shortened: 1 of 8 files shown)


mariusGundersen:
This is wished for here: https://github.com/espruino/Espruino/issues/1302#issuecomment-655780267

src/jsparse.c
```diff
@@ -783,10 +783,10 @@ NO_INLINE JsVar *jspeFunctionCall(JsVar *function, JsVar *functionName, JsVar *t
...
 
             if (execInfo.execute&EXEC_DEBUGGER_STEP_INTO) {
-              if (functionName)
-                jsiConsolePrintf("Stepping into %v\n", functionName);
-              else
-                jsiConsolePrintf("Stepping into function\n", functionName);
+	      if (functionName)
+		jsiConsolePrintf("Stepping into %v\n", functionName);
+	      else
+		jsiConsolePrintf("Stepping into function\n", functionName);
```
gfwilliams, line 789: I just noticed these spaces got replaced with tabs. Please can you try and tweak these (and change your editor so tabs don't get added?). It really screws up the layout if someone views the code and doesn't have 8 space tabs set

src/jsparse.c
```diff
@@ -1123,6 +1150,10 @@ NO_INLINE JsVar *jspeFactorMember(JsVar *a, JsVar **parentResult) {
         jsvUnLock(aVar);
       }
       jsvUnLock(index);
+
+#ifndef ESPR_NO_OPTIONAL_CHAINING
+      JSP_RESTORE_EXECUTE();
```
gfwilliams, line 1155: I'm afraid this breaks stuff.  `undefined[1]` should cause an exception (like 1143) but JSP_RESTORE_EXECUTE removes the error flag, so it just quietly completes. I can imagine there might be other code you could come up with (not just errors, but return/break/etc) too

gfwilliams:
Sorry it's taken so long to look into this! But I am a bit concerned about the use of `JSP_SAVE_EXECUTE/RESTORE` - they need to be used with real care so you don't accidentally remove execution state. I think it might be possible to make it work without them though, as all you care about is executing or not (rather than any of the other state)

PR #2230, opened 2022-07-04
Daylight Savings Time support

deirdreobyrne:
Adds `E.setDST()`, which gives Espruino rules for determining when daylight savings time starts and ends. Note that `E.setTimeZone()` will have no effect when daylight savings time rules are in effect. (I note that Gadgetbridge sets the time zone, so by disabling the effectiveness of `E.setTimeZone()` we prevent clashes between Gadgetbridge and this DST code when DST changes).

Note that one of the tests in `tests/test_date.js` was actually wrong. It was assuming GMT, when the timezone was explicitly set with `E.setTimeZone(2)`.

DST checks have been added to `tests/test_date.js`, and all of them are now passing.


src/jswrap_date.c
```diff
@@ -128,7 +128,7 @@ int jsdGetEffectiveTimeZone(JsVarFloat ms, bool is_local_time, bool *is_dst) {
         } else { // dstEnd <= sec < dstStart
           // Southern hemisphere - DST has ended for the winter
           if (is_dst) *is_dst = false;
-          return dstSetting[0];
+          return dstSetting[1];
```
deirdreobyrne, line 131: Bug fix

src/jswrap_espruino.c
```diff
@@ -1698,13 +1698,80 @@ Set the time zone to be used with `Date` objects.
...
+Note that when DST parameters are set (i.e. when `dstOffset` is not zero), `E.setTimeZone()` has no effect.
+*/
+void jswrap_espruino_setDST(JsVar *params) {
+  JsVar *dst;
+  JsvIterator it;
+  unsigned int i = 0;
+
+  if (!jsvIsArray(params)) return;
+  if (jsvGetLength(params) != 12) return;
+  jsvIteratorNew(&it,params,JSIF_DEFINED_ARRAY_ElEMENTS);
```
gfwilliams, line 1755: Would just replacing the following with `jswrap_typedarray_constructor(ARRAYBUFFERVIEW_INT16, params, 0, 0)` work here? It'd save a bit of code space

src/jswrap_date.c
```diff
@@ -19,29 +19,157 @@
...
+    y--;
+    m+=12;
+  }
+  ans = (y/100);
+  ans = 365*y + (y>>2) - ans + (ans>>2) + 30*m + ((3*m+6)/5) + d - 719531;
+  return ans;
+}
+
+// Convert a number of days since 1970 into y,m,d. 0<=m<=11
+void getDateFromDayNumber(int day, int *y, int *m, int *date) {
```
gfwilliams, line 40: Is there any kind of reference for this that could be linked? This looks a lot like witchcraft to me and it might just be handy to know where it came from.

gfwilliams:
Hi! This looks really good, thanks! I've just enabled the CI check, and there do seem to be some issues though:

* ESP32/Pixl/Pico/MDBT42/etc builds fail because of `td.zone = jsdGetTimeZone()` in `libs/filesystem/jswrap_fs.c` (FAT32 filesystem) - it should be a pretty easy fix as I think it's what you're already doing with a call to setCorrectTimeZone(&td);? https://github.com/espruino/Espruino/blob/042dbffdd585a1b1654b18688be0be9c9b6a7f96/libs/filesystem/jswrap_fs.c#L349
* micro:bit 1 build fails because this seems to use about 1400 bytes more than the previous code (so adding about 1% to code size). The stuff I suggested may help a little but I'm actually pretty surprised this adds so much - not sure if you have any thoughts here? I guess we could just do `#ifndef ESPR_NO_DAYLIGHT_SAVING` and add that define on micro:bit 1, which is already a slightly limited build anyway?
* Are you using tabs? If you look at jswrap_date.c at 209/318 and other places, the indenting seems a bit off?

But thanks - this looks really neat. It'll be a great addition.

src/jswrap_date.c
deirdreobyrne: It is adopted from an algorithm in a book called "Astronomical Algorithms" by Jean Meeus. The book - from 1991 - is well out of print by now.

I'm in the process of doing [my own write-up on these algorithms](https://github.com/deirdreobyrne/CalendarAndDST). I've managed to convince myself of the correctness of everything but the constant 2877911. I've also ran getDayNumberFromDate against getDateFromDayNumber, and they are consistent.

src/jswrap_date.c
```diff
@@ -19,29 +19,157 @@
...
+  JsVar *dst = jsvObjectGetChild(execInfo.hiddenRoot, JS_DST_SETTINGS_VAR, 0);
+  if ((dst) && (jsvIsArrayBuffer(dst)) && (jsvGetLength(dst) == 12) && (dst->varData.arraybuffer.type == ARRAYBUFFERVIEW_INT16)) {
+    int y;
+    JsVarInt dstSetting[12];
+    JsvArrayBufferIterator it;
+  
+    jsvArrayBufferIteratorNew(&it, dst, 0);
+    y = 0;
+    while (y < 12) {
+      JsVar *setting = jsvArrayBufferIteratorGetValue(&it);
```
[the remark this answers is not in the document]
deirdreobyrne: Done

src/jswrap_date.c
```diff
@@ -647,6 +750,30 @@ JsVar *jswrap_date_toISOString(JsVar *parent) {
 
   return jsvVarPrintf("%d-%02d-%02dT%02d:%02d:%02d.%03dZ", date.year, date.month+1, date.day, time.hour, time.min, time.sec, time.ms);
 }
+/*JSON{
+  "type" : "method",
+  "class" : "Date",
+  "name" : "toLocalISOString",
+  "generate" : "jswrap_date_toLocalISOString",
```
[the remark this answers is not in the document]
deirdreobyrne: Done

deirdreobyrne:
> * ESP32/Pixl/Pico/MDBT42/etc builds fail because of `td.zone = jsdGetTimeZone()` in `libs/filesystem/jswrap_fs.c` (FAT32 filesystem) - it should be a pretty easy fix as I think it's what you're already doing with a call to setCorrectTimeZone(&td);?

Yes - that should fix the problem

> * micro:bit 1 build fails because this seems to use about 1400 bytes more than the previous code (so adding about 1% to code size). The stuff I suggested may help a little but I'm actually pretty surprised this adds so much - not sure if you have any thoughts here? I guess we could just do `#ifndef ESPR_NO_DAYLIGHT_SAVING` and add that define on micro:bit 1, which is already a slightly limited build anyway?

I'm also surprised at the 1400 bytes this takes. If it fails again then I guess micro:bit will have to do without DST.

> * Are you using tabs? If you look at jswrap_date.c at 209/318 and other places, the indenting seems a bit off?

I've had *horrible* problems with tabs.

src/jswrap_date.c
deirdreobyrne: I had to do quite a bit of work to the algorithm in "Astronomical Algorithms", as it had its year zero in the year 4713BC. I wanted my year zero to be 1970 to correspond with the zero of Unix time, and the conversion was quite the mind bend.

I just *love* algorithms like that - it was a joy to work on!

deirdreobyrne:
I've done a bit of refactoring so that the code should take a bit less space. Expect another commit probably later today.

gfwilliams:
Excellent - thanks!

gfwilliams:
Excellent - thanks!

deirdreobyrne:
I think that's about as good as I can make the code. I've implemented the ESPR_NO_DAYLIGHT_SAVING for microbit1, but I can't get it to compile for that platform! Could you review and see if I did ESPR_NO_DAYLIGHT_SAVING correctly? Or am I doing something wrong with my attempts to compile the microbit1 version?

gfwilliams:
Great - thanks! I'll take a look tomorrow and will see about merging

gfwilliams:
This looks great to me - the microbit build seems to work fine as well. Merging now - Thanks for this, and for making those tweaks!

deirdreobyrne:
OK - figured out why I wasn't able to compile for microbit1 - I didn't have RELEASE=1 in my makefile.

The microbit1 cannot run the DST code for want of a measly 332 bytes. So unless I can think of some way of squeezing out another 332 bytes we can call this project done.

gfwilliams:
Hi! I know this is a while ago, but a segfault issue was just reported at #2456

Some weirdness can be reproduced just by doing `d = new Date(-12,8)` which seems valid, and this looks like it's down to your replacement getDateFromDayNumber which comes from https://github.com/deirdreobyrne/CalendarAndDST

It's giving negative values for days/months in some cases if the date is before 1970.

Before your changes went in, this worked fine.

```
// before 
>d = new Date(-12,8)
=Date: Tue Aug 31 -12 00:00:00 GMT+0000

// after
>d = new Date(-12,8).ms
=-62524915200000
>d = new Date(-12,8)
=Date: Wed h -29 -11 00:00:00 GMT+0000
```

It seemed you had a really thorough write-up in  https://github.com/deirdreobyrne/CalendarAndDST so I thought this was worth asking you about first

deirdreobyrne:
Eeeeep - my code is only supposed to be used with the Gregorian calendar
(i.e. post-1582, or post-1752 in the UK, US and elsewhere).

And yes I can see how it might have a problem pre-1970, as I coded it with
the idea that Date() starts in 1970.

Which is the most important thing to fix? That it doesn't segfault? That it
is consistent going between millis and Dates? Or that it gives what could
be considered the "right" answer (which would be for example to use
Gregorian after 1582 and Julian before)?

D.

gfwilliams:
Thanks! Well, from my point of view:

* It shouldn't ever segfault (eg month always 0..11, day always 1..31)
* pre-1970 at minimum (ideally 1582) I don't really care, if you do `new Date(-1000,7,1)` you should expect it'll say 
`=Date: ... Jul 1 -1000 00:00:00 GMT+0000` in return. Right now it says `Date: Fri ength -27 -999 00:00:00 GMT+0000`

I don't think the Date<->millis thing matters pre-1582, but it'd be nice if within reason you could add one day's worth of milliseconds and get to the next day.

deirdreobyrne:
I believe I have a bugfix in my github account.

The good news is that I believe it will be good for over 71,000 years.
Though in the very distant past and future you might find yourself unable
to add or subtract a day, due to the precision of floating point, and you
may also have an "incomplete" calendar.

HOWEVER in astronomy (which is one of my hobbies) the convention is to use
the Gregorian calendar as far back as the Gregorian reform in 1582, and the
Julian calendar before that. (And it is also standard practice to have no
year zero). My code breaks these two conventions.

The solutions I see are -

1) Break the conventions
2) Develop code for the Julian calendar, and code to switch between Julian
and Gregorian on the appropriate date in 1582. Whereas different countries
implemented the switch on different dates, the convention is to use the
1582 date.
3) Throw some sort of exception for dates before that date in 1582 (if I'm
reading your code correctly it's fairly easy to throw exceptions?)

Solution 3 is the most technically accurate easy solution, and solution 2
is a reasonably technically accurate difficult solution. I don't like the
current state of affairs.

What do you think?

D.

gfwilliams:
Sorry for the delayed response - what you've managed with #2459 looks great though!

Having as wider range as possible using Gregorian and throwing an exception outside of that seems great

PR #2234, opened 2022-07-14
Add action "Test Factory Apps"

BartS23:
This PR adds an action to test the default apps.

What it does:
- check out Espruino
- check out EspruinoWebIDE
- check out EspruinoAppLoaderCore
- install emsdk
  - at the moment fixed to version 3.1.12, because I get an error with higher versions, which I don't understand
  `error: undefined symbol: emscripten_memcpy_big (referenced by top-level compiled C/C++ code)`
- install node.js
- compile banglejs emulator (emscripten + emscripten2)
- run the tests for BangleJS1 and BangleJS2
- if one of the two tests fails, the whole action fails

The test Script
- adopted from [BangleApps/bin/thumbnailer.js](https://github.com/espruino/BangleApps/blob/master/bin/thumbnailer.js)
- load emulator
- made a factory reset
- load the test script line by line and send them to the emulator
- checks, if the response contains "Uncaught"

The Tests (very simple functional tests)
- about
  - replace the load method to set a variable when called
  - start about app
  - press button 1
  - check if load was called
- alarm
  - Remove sched.json
  - start the alarm app
  - create an alarm
  - create a timer
  - check if sched.json exists
- health (BangleJS2 only)
  - remove health.json
  - load health app
  - change HRM interval
  - check if health.json exists
- sched
  - remove sched.json
  - create a timer for 1 second
  - start sched app
  - "stop" timer
  - check if there is an deactivated alarm/timer
- setting
  - remove .boot0
  - start setting app
  - goto Utils -> rewrite settings
  - check if .boot0 exists
- welcome
  - replace load method to run welcome.boot.js (writes welcome.json file) on call
  - remove welcome.json
  - run welcome app
  - step by step through the scenes
  - finiexit sh welcome (calls load)
  - echo if welcome.json exists

### DIFF
--- a/.github/workflows/testFactoryApps.yml
+++ b/.github/workflows/testFactoryApps.yml
@@ -1,0 +1,79 @@
+name: Test Factory Apps
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+    inputs:
+      git-ref:
+        description: Git Ref (Optional)    
+        required: false
+  
+permissions:
+  contents: read
+
+jobs:
+  # This workflow contains a single job called "build"
+  tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Espruino
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+          path: 'Espruino'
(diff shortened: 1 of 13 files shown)


gfwilliams:
Wow, this is really impressive - thanks!

Since it's adapted from thumbnailer, did you consider taking/comparing screenshots too? I guess that would be a really good way to check things were definitely working as expected?

PR #2235, opened 2022-07-20
Identify libraries and optional arguments in JSON

qucchia:
For the TypeScript typing I have to know which classes are libraries (which must be retrieved using `require()`) and which are global variables. Hopefully this shouldn't break Tern?

### DIFF
--- a/scripts/build_tern_json.js
+++ b/scripts/build_tern_json.js
@@ -72,16 +72,22 @@ require("./common.js").readAllWrapperFiles(function(json) {
       } else if (["constructor"].indexOf(j.type)>=0) {
         var o = tern[j.class]; // overwrite existing class with constructor
         if (o===undefined) o=tern[j.class]={};
-        o["!type"] = j.getTernType();
+        var [type, rest] = j.getTernType();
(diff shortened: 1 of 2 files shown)


qucchia:
This last commit also adds a question mark to arguments that are explicitly described as optional (their documentation beginning with `[optional]`).

gfwilliams:
Thanks - have you been able to test the generated file with tern?

The other option is just to copy `scripts/build_tern_json.js` and instead use it to output typescript decls directly? Then you have access to a lot more info (for instance if needed we could add extra fields to the JSON which you could then access)

qucchia:
Oh true! That would be neater. Perhaps it would be cool to put the types in their own dedicated NPM package?

gfwilliams:
> Perhaps it would be cool to put the types in their own dedicated NPM package?

Not sure I understand - you mean the generated types, rather than the type info that's currently sitting in JSON?

qucchia:
Yeah - something like this: https://www.npmjs.com/package/@types/espruino. If not it doesn't matter though.

PR #2237, opened 2022-07-21
Auto-generate types for TypeScript

qucchia:
New script that generates a `main.d.ts` file for BangleApps.
- Currently it assumes the `Espruino` and `BangleApps` are in the same folder and that the script is run from the Espruino folder.
- I've temporarily added a third item to the return option in JSON comments to specify TypeScript types (see src/jswrap_arraybuffer.c).

### DIFF
--- a/scripts/build_types.js
+++ b/scripts/build_types.js
@@ -1,0 +1,201 @@
+#!/usr/bin/node
+
+const fs = require("fs");
+
+function indent(string) {
+  return string
+    .split("\n")
+    .map((line) => (line ? "  " + line : line))
+    .join("\n");
+}
+
+// TODO: Document parameters
+function getDocumentation(object) {
+  if (!object) return "";
+  return (
+    "/**\n" +
+    object
+      .getDescription()
+      .split("\n")
+      .filter((line) => line)
+      .map((line) => line)
+      .concat([`@url ${object.getURL()}`])
+      .map((line) => " * " + line)
+      .join("\n") +
+    "\n */"
+  );
+}
+
+function getBasicType(type) {
+  if (!type) return "any";
+  if (["int", "float", "int32"].includes(type)) return "number";
+  if (type == "pin") return "Pin";
+  if (type == "bool") return "boolean";
+  if (type == "File") return "EspFile";
+  if (type == "JsVarArray") return "any";
+  if (type == "JsVar") return "any";
+  if (type == "Array") return "any[]";
+  if (type == "Promise") return "Promise<any>";
+  return type;
+}
+
+function getDeclaration(object, c) {
+  if (
+    ["function", "method", "staticmethod", "constructor"].includes(object.type)
+  ) {
+    // function
+    let args = [];
+    if ("params" in object)
+      args = object.params.map((param) => {
+        // hack because digitalRead/Write can also take arrays/objects (but most use cases are Pins)
+        if (param[0] == "pin" && param[1] == "JsLet") param[1] = "Pin";
+        if (param[0] === "function") param[0] = "func";
+        if (param[0] === "var") param[0] = "variable";
+        let doc = typeof param[2] === "string" ? param[2] : param[2].join("\n");
+        let optional = doc && doc.startsWith("[optional]");
+        let rest = param[1] === "JsVarArray";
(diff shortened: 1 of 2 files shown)


qucchia:
Hadn't event thought of using `return_object`, will do!

gfwilliams:
Thanks for getting that change in so quick!

Is `ArrayBufferLike` something built into Typescript? Because Espruino should hopefully return `ArrayBuffer` for `.buffer`?

Just a thought, but longer-term I wonder if Typescript types couldn't be included in the declaration - either in the description, like `[optional, number|String]` or in a 4th argument (or after the 2nd argument, like `"JsVar,number|String"`

qucchia:
So the reason I set the return type to `ArrayBufferLike` is because ArrayBufferView is already defined in TypeScript (see https://github.com/microsoft/TypeScript/blob/main/lib/lib.es5.d.ts#L1605) and if we don't set the type of `buffer` to ArrayBufferLike, TypeScript throws an error.

Also, that suggestion is a good idea. I guess putting it after the 2nd argument makes most sense?

gfwilliams:
Ahh, that's an interesting one... Is there a way to disable including the default TypeScript decls? They may do more harm that good (for example making users think they can use functionality that Espruino doesn't provide).

From the point of view of the IDE, documentation, etc - saying it returns `ArrayBufferLike` isn't great, because it's not related to Espruino at all - it's purely so it can work with TypeScript. I guess we could add a `typescript_returns` field to make it clearer that actually it's just related to typescript rather than what it actually returns?

mariusGundersen:
I too would love some better TypeScript types for espruino, and it would be great if it was part of either the `espruino` package or the `@types/espruino` package. 

I have a [branch that I've started working on](https://github.com/mariusgundersen/espruino/tree/typedefsbut) it's not ready for pr yet. There I've added a new property to the json specifically for ts type definitions. That makes it possible to define complex parameters and return types, overloaded methods and union types. Most of these types can be auto generated based on the existing types, but there are many that can be improved upon to have much better TypeScript types. That is unfortunately a manual job though.

Other things I want to add is generating multiple ts files, one for each supported board. For example, I don't want to see types for the bangle while working on the pico. BTW, that is a feature I want to add to the reference documentation too. It would be really nice to be able to see which boards a function works on and be able to filter the functions based on a board on the documentation page.

I have summer holiday now and therefore can't work on this, maybe I'll be able to do it in August.

mariusGundersen:
@gfwilliams there is a way to not include the default declarations, by using the tsconfig [noLib](https://www.typescriptlang.org/tsconfig#noLib) option,but I haven't tried it yet.

qucchia:
Thanks for all that info and your suggestions make a lot of sense! I'll try out that `noLib` option.

qucchia:
OK, now `ArrayBufferLike` is not needed anymore :+1:

gfwilliams:
> OK, now ArrayBufferLike is not needed anymore

Excellent - thanks! This looks perfect - merging now!

> BTW, that is a feature I want to add to the reference documentation too

`BOARD=boardname make docs` should do you :) I'm not sure if `common.js` supports that, but it should be pretty easy to add it if not - and once that's done we could easily rig this up so that you could do ts types for each individual board

> I have a [branch that I've started working on](https://github.com/mariusgundersen/espruino/tree/typedefs)

Personally, I think your approach there might be overkill... Stuff like:

```
/*JSON{
  "type" : "class",
  "class" : "TypeError",
  "typedef": "class TypeError"
}
```

Feels redundant to me - it's trivial for the code in this PR to work out what the typescript decl should be, and the more redundant code we have, the more likely it is it'll end up being unmaintained and broken in the future.

But we should absolutely have some kind of field where we can override the default handling of the declaration if it's nontrivial, for example in the case of `poke` you mention, or where an object with known fields is returned. Personally I'd call the field `typescript` so it's obvious, but I think if we added something like that, we could then add the functionality slowly rather than requiring PRs that change 100 or more files.

mariusGundersen:
I'll open some issues where this can be discussed as this one is merged.

gfwilliams:
Ok - maybe just **one** issue? Otherwise it's just going to get complicated

PR #2238, opened 2022-07-22
TypeScript: Document using JSDoc format

qucchia:
Automatically adds [JSDoc](https://jsdoc.app/) tags (`@param`, `@returns`, and `@constructor`) to give info about each function. See the result: https://github.com/qucchia/BangleApps/blob/master/typescript/types/main.d.ts.


gfwilliams:
Great!

PR #2241, opened 2022-07-22
Add custom typescript declarations in JSON comments

qucchia:
You can now define custom types like so:
```c
/*TYPESCRIPT
type NRFFilters = {
  services?: string[];
  name?: string;
  namePrefix?: string;
  id?: string;
  serviceData?: object;
  manufacturerData?: object;
};
*/
```
And custom function types: (the first item is the arguments, and the second the return type):
```json
{
  ...
  "typescript" : ["callback: (devices: BluetoothDevice[]) => void, options?: number | { filters?: NRFFilters, timeout?: number, active?: boolean }", "void"]
}
```

Now we can start slowly but surely covering all the API with their correct types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ Espruino JavaScript for Microcontrollers
 |   __|___ ___ ___ _ _|_|___ ___
 |   __|_ -| . |  _| | | |   | . |
 |_____|___|  _|_| |___|_|_|_|___|
-          |_|      
+          |_|
 </pre>
 https://www.espruino.com &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; [![Join the chat at https://gitter.im/espruino/Espruino](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/espruino/Espruino?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 
@@ -16,7 +16,7 @@ About
 
 Espruino is a JavaScript interpreter for microcontrollers. It is designed for devices with as little as 128kB Flash and 8kB RAM.
 
-Please support Espruino by [ordering one of our official boards](https://www.espruino.com/Order) or [donating](https://www.espruino.com/Donate)
+Please support Espruino by [ordering one of our official boards](https://www.espruino.com/Order) or [donating](https://www.espruino.com/Donate).
 
 
 Documentation
@@ -28,7 +28,7 @@ Browse the [Espruino Website](https://www.espruino.com) (try using search in the
 
 There's also a [Reference](https://www.espruino.com/Reference) for JavaScript commands as well as [Tutorials](https://www.espruino.com/Tutorials). However the documentation on the Espruino website will match the version [available for download](https://www.espruino.com/Download) but **not** the latest version on GitHub.
 
-Builds for the [Espruino Board](https://www.espruino.com/EspruinoBoard) and [Pico Board](https://www.espruino.com/Pico) (built automatically for each Git commit) are [available from here](https://www.espruino.com/binaries/git)
+Builds for the [Espruino Board](https://www.espruino.com/EspruinoBoard) and [Pico Board](https://www.espruino.com/Pico) (built automatically for each Git commit) are [available from here](https://www.espruino.com/binaries/git).
 
 Other documentation of use is:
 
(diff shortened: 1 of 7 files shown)


qucchia:
I've quickly done the types for E.showMenu

mariusGundersen:
I think the TypeScript property should be the full type declaration, not the just the parameters and return type. I know that it's a bit of duplication of information, since there are properties already for knowing if it's a class, method, event etc. But with the proposed definition there is no way to define generic methods (needed for promises for example) or methods with overloads (for example https://github.com/mariusGundersen/Espruino/blob/de8648ccd49ad4982b1dd49a80777ecf849a4fc6/src/jswrap_io.c#L36). But by having it be the full type string or an array of strings it covers all scenarios.

qucchia:
I understand that, but let me explain my reasons. The thing is, functions are defined in _two ways_ depending on the context:
- `function a(...args: any[]): any` if it's global, or:
- `a: (...args: any[]) => any` if it's a property of a class.

We would have to declare global methods and class methods in _different formats_ and I thought that would get too confusing.

By the way, generic methods _can_ be defined this way (just add the `<T>` bit before the arguments list; see [here](https://github.com/qucchia/Espruino/blob/39314d59336f390a4e4f78ef53db0c4f409585fe/src/jswrap_promise.c#L208)). As for overloads, they could be handled by having an array, like: `[[arguments0, return0], [arguments1, return1]]`.

Hope this makes sense!

mariusGundersen:
Yeah, global methods and class methods and static methods and properties and static properties and variables would all need to be defined differently. But is that a big problem? I think it would be easier to define them correctly when you can type out the entire definition, rather than having to try to assemble parts correctly and then checking with the generated output if it was right?

For variables and (static) properties, would there only be one item in the array, the return type? 

Yes, generics like that could work, but not for events (although I doubt there are any generic events).

mariusGundersen:
I would also type some of the classes as `interface`, for example the http request and response types that shouldn't be constructable by the programmer. And I would want to mark some of the class properties as `readonly`.

qucchia:
Allow me to suggest this:
- **Functions**: always defined as `"typescript": "a(...args: any[]): any"`.
  - I was wrong when I said class methods had to be declared with arrow functions. They can be declared this way too!
  - If it belongs to a class, you can add `readonly` to the start and `?` after the function name.
  - The keywords `declare function` will be automatically added if it's a global method. (There really is no need to add this in the comments as it's superfluous.)
  - If there are multiple overloads, specify an array of strings like the above, e.g.
  `"typescript": ["a(...args: any[]): any", "a(): void"]`.
- **Variables and properties**: always defined as `"typescript": "b: any"`.
  - Again, you can add the `readonly` and `?` keywords.
  - The keywords `declare const` will be automatically added if it's a global variable.
- **Classes**: usually no definition is necessary, but it can be defined as an interface (`"typescript": "interface httpSRq"`), or generic properties can be specified: `"typescript": "interface Array<T>"`. This argument can then be used in the class' methods and properties.

Hopefully this satisfies both of our needs!

qucchia:
As another thought, perhaps it may be a good idea to make _all_ member functions and properties readonly. In that case we wouldn't have to explicitly define it.

mariusGundersen:
> Hopefully this satisfies both of our needs!

Haha, I think we have the same need; better TypeScript support! 😜

In general I'm not a fan of increased complexity and cleverness just to save on bytes, and I don't think DRY is more important than clear and understandable documentation. And it makes it harder to add other things in the future, for scenarios that we haven't thought of here. So why not have a bit of superfluous documentation and be able to support all kinds of TypeScript features? Is it really that bad? 

Readonly and static can be added where needed. Is every variable a const? Or are some variables writable? For certain every property isn't readonly, so that can't be added everywhere. And I don't think functions need to be marked readonly. For arrays the array index type needs to be defined, somehow.

mariusGundersen:
I'd say it also makes it easier to document features for us and for others, as we can write TypeScript declarations and copy paste them into the documentation, rather than having to pick it apart and remember what will be automatically added. For new contributers it is impossible to know how it works. Much easier to have the full TypeScript declaration.

qucchia:
Ok sure, let's do it like that. I'll revert the commits so as to not cause clutter with too many changes.

qucchia:
Ok, sure, let's do it like that. I'll close this and open a new PR with that implemented.

PR #2242, opened 2022-07-23
Add custom TypeScript declarations in comments

qucchia:
You can now define custom types like this:
```c
/*TYPESCRIPT
type NRFFilters = {
  services?: string[];
  name?: string;
  namePrefix?: string;
  id?: string;
  serviceData?: object;
  manufacturerData?: object;
};
*/
```
And custom function types:
```js
{
  ...
  "typescript": [
    "declare function require<T extends keyof Libraries>(moduleName: T): Libraries[T];",
    "declare function require<T extends Exclude<string, keyof Libraries>>(moduleName: T): any;"
  ]
}
```
Finally you can also declare generic classes:
```js
{
  ...
  "typescript": "interface Array<T>"
}
```

### DIFF
--- a/libs/banglejs/jswrap_bangle.c
+++ b/libs/banglejs/jswrap_bangle.c
@@ -4828,6 +4828,7 @@ to select an application to launch.
       ["menu","JsVar","An object containing name->function mappings to to be used in a menu"]
     ],
     "return" : ["JsVar", "A menu object with `draw`, `move` and `select` functions" ],
+    "typescript": "showMenu(menu: Menu): MenuInstance;",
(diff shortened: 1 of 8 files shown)


qucchia:
See the results here: https://github.com/qucchia/BangleApps/blob/typescript/typescript/types/main.d.ts

qucchia:
Documented the build script so it hopefully isn't a completely incomprehensible mess.

gfwilliams:
This looks great to me - fits in nicely without complicating things too much.

You think this is ready to merge?

qucchia:
Yep!

PR #2248, opened 2022-07-27
More Typescript

qucchia:


gfwilliams:
This looks great - thanks! It's going to make a huge difference to autocomplete in VS Code even for those not using TypeScript directly.

PR #2250, opened 2022-07-27
Typos

qucchia:
If you're not happy with me replacing eg with e.g. I can undo that.


gfwilliams:
Great, thanks! Do you have some spellcheck script?

Personally I wouldn't have changed `eg` but as it's done now lets leave it as I don't actually care that much :)

PR #2254, opened 2022-07-27
Typescript

qucchia:
Fixed some minor things and created a Readme explaining how to add TypeScript declarations.

### DIFF
--- a/TYPESCRIPT.md
+++ b/TYPESCRIPT.md
@@ -97,6 +97,8 @@ Here are some guidelines for making manual declarations:
   with `"[optional]"`. If adding this keyword saves having to declare the
   function manually, add it!
 
+- Include the `declare` keyword in global functions and variables.
+
 - Try to keep argument names the same as the ones specified in the `"param"`
   field. In cases that this isn't possible (with reserved keywords such as
   `function` or `var`) you'll have to change the name a little (for example,
--- a/TYPESCRIPT.md
+++ b/TYPESCRIPT.md
@@ -1,0 +1,107 @@
+# Espruino and TypeScript
+
+## Build script
+
+The build script at `scripts/build_types.js` will automatically generate types
+from documentation comments. To use it run:
+
+```sh
+node scripts/build_types.js
+```
+
+This will update the file located at `typescript/types/main.d.ts` in the
+BangleApps repository!
+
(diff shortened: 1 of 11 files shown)


gfwilliams:
Thanks!

mariusGundersen:
A question about the output of the build. You want this for bangle apps, right? Which is why it's being output in the bangleApps directory? There is a `@types/espruino` package for general espruino development, maybe that should be updated?

qucchia:
Yeah, that would be great! I'll try PR-ing [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped).

qucchia:
Right - I can't right now because the tests at [espruino-tests.ts](https://github.com/qucchia/DefinitelyTyped/blob/master/types/espruino/espruino-tests.ts) use the `import` syntax and the pins D0-D17 are not defined. I can't figure out what board this is supposed to be for?

gfwilliams:
> pins D0-D17 are not defined. I can't figure out what board this is supposed to be for?

Going by what's in that test I'd say maybe it's a Pixl.js with Wifi attached - maybe http://www.espruino.com/Pixl.js+Multicolour - but that's so niche only a few people would ever have that setup.

Or maybe it's just trying to test a bunch of different areas from different boards. The pin thing is tricky - do you want to add every possible pin that would be on every board? Or a typescript file per board? Or maybe just don't do pins at all...

mariusGundersen:
I want to try to set up one TypeScript file per board, as mentioned before, but unfortunately I have absolutely no time to work now (9 month old twins), which is why my only contribution right now is comments. Alternatively I was thinking of something advanced using conditional types and setting `const board = 'pico'` but I'm not sure that is any better, or would even work.

qucchia:
Yeah, that would be great - having a TypeScript file per board, I mean. I'm away during August, so when I'm back I'll try to do it.

gfwilliams:
> (9 month old twins)

Honestly I'm just amazed you're able to post anything online at all :)

> I was thinking of something advanced using conditional types

I guess the end result might be neater, but actually implementing that nicely would probably be too tricky. Hopefully the mods for separate board files wouldn't be too bad - it should work exactly like is done for docs (with `BOARD=XYZ make docs`) - the Makefile is responsible for working out some build flags that then effect what gets included in the firmware

PR #2255, opened 2022-07-28
Fixed lexer bug for non UTF-8 characters

OxNinja:
With the following file Espruino crashes with a SEGFAULT:
```js
깵੬
```
Here are the hexdump of the file + its base64:

```hex
00000000: 75ae 6c0a                                u.l.
```

`da5sCg==`

To crash the program:

```sh
echo 'da5sCg==' | base64 -d > crash.js
espruino crash.js
```

The root cause of the crash is at https://github.com/espruino/Espruino/blob/master/src/jsparse.c#L2297
Because the lexer did not sanitize/stopped on non parsed character, thus `*lastDefined` points to address `0` which crashes the program.

I don't know if my way of stopping the execution is the best but it worked on my computer.

### DIFF
--- a/src/jslex.c
+++ b/src/jslex.c
@@ -437,6 +437,13 @@ void jslGetNextToken() {
   // record beginning of this token
   lex->tokenLastStart = lex->tokenStart;
   unsigned char jumpCh = (unsigned char)lex->currCh;
+
+  // lexer must stop parsing for a variable name if the current char is not UTF-8
+  if ((jumpCh < 0x20 || jumpCh > 0x7e) && jumpCh != 0x00) {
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks! I'll take a look at this - I don't believe we can reject non-ASCII chars because we use them for pretokenised code (which is maybe why this code got into jspeStatementVar).

But `jspeStatementVar` returning 0 shouldn't really cause a problem

gfwilliams:
Thanks! I just put a fix in.

It just happened that the characters you used caused a problem, but actually you could have reproduced this with `if (false) const x = 5;` :)

OxNinja:
okay perfect!

PR #2256, opened 2022-08-12
Upload GitHub Build Artifacts

DanTheMan827:
Since you already had an action, I modified it to also upload the build artifacts as well.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,6 +34,16 @@ jobs:
         source ./scripts/provision.sh  ${{ matrix.board }}
         make
         ./scripts/ci_upload.sh
+    - name: Upload ${{ matrix.board }} build artifact
+      uses: actions/upload-artifact@v3.1.0
+      with:
+        name: ${{ matrix.board }}
+        path: |
+          *.bin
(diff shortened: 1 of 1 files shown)


gfwilliams:
Where does it upload them to/how do you get them?

Because they are already all uploaded to https://www.espruino.com/binaries/travis/ /  https://www.espruino.com/binaries/travis/master

DanTheMan827:
@gfwilliams it attaches them to the github action run.

Example from my fork: https://github.com/DanTheMan827/Espruino/actions/runs/2849321887

You don't have to if you don't want, but I figured it's just another source for them

gfwilliams:
Great - thanks! Yes, that looks neat.

Maybe I'll have to re-evaluate if GitHub starts charging for storage somehow, but until then this seems like a neat thing to have. Also for anyone wanting to clone, make a change and have it built for them

DanTheMan827:
@gfwilliams you may want to change the action to run on all branches along with a check for `ci_upload.sh` to only run when on master.

Ideally, someone forking to make changes would also make a branch for those changes rather than work on their master, and the action wouldn't run for them.

Issue #2258, opened 2022-08-24
Math.round(1.5) isn't 2!


gfwilliams:
See http://forum.espruino.com/conversations/337195/#comment16657931

MaBecker:
Just inspection the code and comparing with  other implementation, difference is the use 0.5 and not 0.4999999

gfwilliams:
Fixed with https://github.com/espruino/Espruino/pull/2260

PR #2260, opened 2022-08-27
Fixed round 1.5 bug

nitzanfarhi:
Solving the following issue https://github.com/espruino/Espruino/issues/2258
by changing the value added to x by 0.5 (or -0.5 if negative) instead of 0.4999999999 previously.
In this case, the number 1.5 will be updated to 2 (and after cast will remain 2),
 whereas 1.49 will be updated to 1.99 (and will be floored to 1).


gfwilliams:
Thanks! Merging now!

I'm slightly anxious that using 0.4999999 was done for some reason (it was in 24558e6f7e) as using 0.5 is the most obvious solution but that commit was 8 years ago and I can't remember why now!

PR #2264, opened 2022-09-07
Adds an isWorn method to the Bangle class

splch:
discussion: http://forum.espruino.com/conversations/379538/

I haven’t tested the function, but does the object returned by getAccel allow C-style structure field calls? I call .mag but don’t know if that’s okay.

### DIFF
--- a/libs/banglejs/jswrap_bangle.c
+++ b/libs/banglejs/jswrap_bangle.c
@@ -2413,6 +2413,10 @@ int jswrap_banglejs_isLocked() {
 */
 // emscripten bug means we can't use 'bool' as return value here!
 int jswrap_banglejs_isWorn() {
+  if (jswrap_banglejs_isCharging())
+    return 0;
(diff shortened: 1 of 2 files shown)


gfwilliams:
> does the object returned by getAccel allow C-style structure field calls?

I'm afraid not... you have to use a jsv... method to access JsVar objects. To test this in your branch I believe as long as you enable GitHub actions for your repo, Bangle.js firmware will automatically be built (or attempted - in this case it would fail) and the files are available for you to download.

Actually when I said library I was thinking that rather than building it into the firmware you'd just add a JavaScript library in https://github.com/espruino/BangleApps/tree/master/modules - sorry for the confusion. 

Problem with doing it this way is everyone would have to update their firmware before it was available - and if you ever needed to change the values it'd require another firmware update.

I've just added a library for you at https://github.com/espruino/BangleApps/blob/master/modules/wear_detect.js but initial tests show it could do with a bit of work - closing this PR for now.

PR #2270, opened 2022-09-26
strip file names and line numbers from tensorflow error messages for release build

fanoush:
- add gcc option to strip build folder absolute path from __FILE__ macro
- make tensorflow NDEBUG dependent on RELEASE - may enable more debug stuff for non-release builds
- disable logging lines and source file names based on NDEBUG

### DIFF
--- a/libs/tensorflow/patches/02-ndebug-omit-file-line-logging.diff
+++ b/libs/tensorflow/patches/02-ndebug-omit-file-line-logging.diff
@@ -1,0 +1,61 @@
+diff --git a/libs/tensorflow/tensorflow/lite/c/common.h b/libs/tensorflow/tensorflow/lite/c/common.h
+index 18997298d..4a7ba8bb6 100644
+--- a/libs/tensorflow/tensorflow/lite/c/common.h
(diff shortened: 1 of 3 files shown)


fanoush:
can you check if you like this? the tensorflow patch file for patching original source is not there yet, will generate it if changes are OK

gfwilliams:
Thanks! Looks good to me... Did you spot if not having `-DNDEBUG` included anything specific? It seems odd I'd included it all the time before (I might have done it just so that I could build *something* :) )

fanoush:
NDEBUG should disable asserts and some profiler code, see https://github.com/espruino/Espruino/search?q=NDEBUG&type=code
I actually did not try to **run** non-release build in the device to see what those asserts do now but it builds and links fine in both cases. If you do run such builds maybe you'll decide to put it back later?

I have added patch produced by `git diff`. It looks normal, same unified diff style as the 01 one, just the first line says `diff --git` so  I hope it is fine.

gfwilliams:
Ok, thanks! This looks great! I think the first line of the diff usually gets ignored anyway.

PR #2271, opened 2022-09-30
Basic Fetch for HTTP

JThobbyist:
A simple Fetch-like interface built on top of the native http module. Hopefully in the future, more and more of the Fetch spec will be added. Supports only res.text() and res.json() right now.

Example Usage:  

```
fetch('http://example.com').then(function (res){
    return res.text();
}).then(console.log);
```

Here is the "pre-minified" code:

```
const http = require('http');

class FetchResponse {
    constructor(res) {
        this.res = res;
    }
    text() {
        const _this = this;
        return new Promise(function (r, j) {
            let timeout;
            let data = "";
            function handleClose() {
                clearTimeout(timeout);
                r(data);
            };
            _this.res.on('data', function (chunk) {
                data += chunk;
            });

            _this.res.on('close', handleClose);

            timeout = setTimeout(function () {
                _this.res.removeListener('close', handleClose);
                j('FetchRequest Timed-Out!');
            }, 5e3);
        });
    }
    json() {
        return this.text().then(JSON.parse);
    }
}

function fetch(address, params) {
    if (!params) params = {};
    const options = Object.assign(url.parse(address), params);
    return new Promise(function (r, j) {
        const req = http.request(options, function (res) {
            try {
                r(new FetchResponse(res))
            } catch (e) {
                j("Could Not Create FetchResponse Object: " + e)
            }
        });
        req.on('error', function (e) {
            //console.log('HTTP Error:', e);
            j('HTTP Error: ' + e);
        });
        req.end(params.body);
    });
}

module.exports = fetch;
```


gfwilliams:
Hi! I think you might have actually meant to add this to https://github.com/espruino/EspruinoDocs/tree/master/modules instead?

Adding the code here doesn't automatically do anything (it'd need to be manually referenced from the relevant C source files). Having it in EspruinoDocs would mean that it could be used with `require` from any existing firmware (and could have documentation too) so I think that could be a real benefit

JThobbyist:
Hey Gordon,

Yes you’re right, that is what I was trying to do! I’ll create a new PR for that directory. 

Best,
- Joe Teglasi

gfwilliams:
Ok, great! I'll close this for now then...

PR #2272, opened 2022-10-06
Remove Bangle.CLOCK variable on setUI()

halemmerich:


gfwilliams:
Thanks - do you have any background to this?

Is this because of the launcher-without-reload changes? Because it still won't help with the clock widget because it chooses not to load itself when the widgets are first loaded...

halemmerich:
Yes, it won't help now, but without this change it won't work correctly when the widget loading/reloading/updating stuff is solved.

PR #2274, opened 2022-10-20
Make the launcher configurable

halemmerich:
Allows choosing between multiple installed launchers and saves a bit of time by not iterating through all apps.


gfwilliams:
Thanks! As we're changing it, please could you save the result of `require("Storage").read(launcherApp)` in a variable, so you're not potentially calling `require("Storage").read(launcherApp)` three times with the same value? That should improve speed even further

libs/js/banglejs/Bangle_showLauncher.js
```diff
@@ -1,12 +1,24 @@
...
+    launcherApp = require("Storage").list(/\.info$/)
+      .map(file => {
+        const app = require("Storage").readJSON(file,1);
+        if (app && app.type == "launch") {
+          return app;
+        }
+      })
+      .filter(x=>x)
+      .sort((a, b) => a.sortorder - b.sortorder)[0].src;
+  }
```
gfwilliams, line 13: Actually, how about we update the settings here with the launcher we found?

Because as-is, this is going to slow down everyone's device *until* they manually set the launcher in settings. And then if they change launcher via the app loader and forget to change the settings, it'll silently end up slow again

halemmerich:
Storing the result is a good idea. I did not do that initially to keep it similar to how the clock loading works, but maybe that would be useful there as well.

gfwilliams:
> but maybe that would be useful there as well.

Actually yes! I never thought to check how long that takes.

Just checked - on my relatively unloaded device, 170ms! Setting `.clock` brings that down to 10ms!

If you could do a PR for that, it'd be brilliant.

libs/js/banglejs/Bangle_showLauncher.js
```diff
@@ -1,12 +1,32 @@
...
+    launcherApp = require("Storage").list(/\.info$/)
+      .map(file => {
+        const app = require("Storage").readJSON(file,1);
+        if (app && app.type == "launch") {
+          return app;
+        }
+      })
+      .filter(x=>x)
+      .sort((a, b) => a.sortorder - b.sortorder)[0].src;
+    print(launcherApp);
```
gfwilliams, line 15: Is this left over from debugging?

halemmerich: Yes it was. Fixed another oversight in the new commit as well.

gfwilliams:
Excellent - thanks! With this recent set of changes there are some pretty huge speed improvements

PR #2275, opened 2022-10-21
Fix using not defined variable if uiRemove is available

halemmerich:
Last commit introduced using an incorrect variable for the launcher app code.


gfwilliams:
Thanks! I just spotted that here too ;)

PR #2279, opened 2022-10-26
Software SPI - fix LSB mode when receiving, was always MSB

### LINKED ISSUE #2278: Software SPI MISO always MSB?
I was trying to use software SPI to read SWD data bits however no matter what mode is set - "msb" or "lsb" it comes back same and wrong for me = msb , this line below looks like result is always shifted in same direction
https://github.com/espruino/Espruino/blob/5345b5f71ff4fa2c53ba0a772ea356270b5ca947/src/jsspi.c#L74
Is this a bug or actually a feature of SPI?
I already solved the issue by using combined routine that reverses bit order and switches endianness in one go as I need to do it anyway since SPI transfers are byte oriented, however if this is bug it can be fixed anyway.

BTW it would be nice to send/receive arbitrary number of bits with each SPI transfer. OTOH for sending bits there is already the [shiftOut](https://github.com/espruino/Espruino/blob/master/src/jswrap_io.c#L562) , not sure how it came to be but complementary `shiftIn` would be nice too :-)

fanoush:
fixes #2278

### DIFF
--- a/src/jsspi.c
+++ b/src/jsspi.c
@@ -71,7 +71,7 @@ void jsspiSoftwareFunc(
         if (inf->pinSCK != PIN_UNDEFINED)
           jshPinSetValue(inf->pinSCK, !CPOL );
         if (inf->pinMISO != PIN_UNDEFINED)
-          result = (result<<1) | (jshPinGetValue(inf->pinMISO )?1:0);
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks!

PR #2285, opened 2022-11-07
Only clear if widgets with width>0 exist

halemmerich:


gfwilliams:
My only concern here is what happens if you have a widget that hides and reappears, and it's the only one on a row? In that case the widget won't get erased.

... but maybe in that case it's up to the widget to clear its area when it sets its width to 0? Because I guess there are cases that could happen before as well (eg the back widget that just gets removed)

halemmerich:
Widgets hiding themselves being responsible for clearing their area is probably a valid stopgap solution. I suspect a really clean solution is only possible if drawing and relayouting are two different calls, so apps can react properly to a widget bar no longer being used.
Currently that is probably only practically relevant for self hiding bottom widgets such as widclkbttm. That one showed the problem in clocks using widgets, the bottom bar is cleared because of the widget there, but after the first `drawWidgets` the appRect includes that space and is used by the clock. I think the top bar has at least one additional static widget for close to every user.

PR #2286, opened 2022-11-07
Implement passing through remove methods for E.show*

halemmerich:
This implements additional options for using a remove method in `E.show*`. There are a few things missing as I was not sure if this approach is the correct one for cleaning up after those methods are done.

- [x] Documentation (where is it needed? I have found jswrag_bangle.c with documentation)
- [x] I have no Bangle.JS 1, so I can only test on Bangle.JS 2


gfwilliams:
This looks good, thanks! My only worry is doing `{..., remove:options.remove}` will create `{remove:undefined}` if remove isn't defined. Hopefully other code just checks `if (options.remove)` but if anything were to do if ("remove" in options)` it might break - so it's worth checking that the functions still work *without* remove supplied as an argument :)

> Documentation (where is it needed? I have found jswrag_bangle.c with documentation)

Yes, it's there - if you go to http://www.espruino.com/Reference and click ⇒ next to the title it'll bring you to the right place :)

Some methods can be used on Espruino boards like Pixl.js as well, so it could be worth noting in those cases that `remove` only works for Bangle.js

> I have no Bangle.JS 1, so I can only test on Bangle.JS 2

If you can come up with some code to run and expected outcomes I can quickly run through it to test?

halemmerich:
> If you can come up with some code to run and expected outcomes I can quickly run through it to test?

That would be great. I can test the code by just running it on the Bangle 2, but I don't know how valid the results would be :)

>Hopefully other code just checks if (options.remove)

Do you think that just checking the code in the repositories for this would be enough, or should I just move the object into a variable and conditionally add the remove property before giving it to setUI?

halemmerich:
I have added some documentation, but not quite sure on the syntax for the typescript definitions. Regarding the use of ['remove'] or similar a quick search had no results in Espruino and BangleApps repositories. It seems the back function is handled the same way and can be undefined:
```
  var back = options.back||menu["< Back"];

//later
  var scr = { 
    //...
    back : back,
    //...
  }
```

halemmerich:
For testing I have used this code:
```
Bangle.loadWidgets();
Bangle.drawWidgets();

var mainmenu = {
  "": {
    title: "-- Main Menu --",
    remove: () => { print("removed main"); }
  },
  "Submenu": function() { E.showMenu(submenu); },
  "Exit": function() { E.showMenu(); },
};
var submenu = {
  "": {
    title: "-- SubMenu --",
    back: function() { E.showMenu(mainmenu); },
    remove: () => { print("removed sub"); }
  },
  "Exit": function() { E.showMenu(mainmenu); },
};
var alert = {
  remove: () => { print("removed alert"); }
};
var prompt = {
  remove: () => { print("removed prompt"); }
};
```
Just uploaded that to RAM and then used various combinations of the following commands and interactions with the UI to check if the removal message is printed.
```
E.showMenu(mainmenu);
E.showPrompt("Prompt", prompt);
E.showAlert("Alert", alert);
E.showPrompt();
E.showAlert();
E.showMenu();
Bangle.setUI();
```

gfwilliams:
This looks great - thanks! I'll give it a quick test on Bangle.js 1 now

gfwilliams:
Yes, seems to work for me! Merging...

PR #2288, opened 2022-11-09
Clarify reference re back in setUI

thyttan:
Just  a suggestion for formulation


gfwilliams:
Sounds good to me - thanks!

PR #2289, opened 2022-11-10
Adding Bangle.load() and Bangle.showClock() methods

halemmerich:
Currently launchers mostly call either `load` or something like `setTimeout(eval,0,require("Storage").read(".bootcde"))` to load the clock. This replaces that and some logic for checking if UI-removal is possible by simple methods that can be used instead of duplicating that code in a lot of apps.
The `Bangle.load` method additionally allows to configure a default app to be loaded. This could be used to remove duplications like widclose/widcloselaunch which do exactly the same but load either clock or launcher.
For older firmware these methods could be defined as polyfills in boot code.


gfwilliams:
Thanks! I think this is a good plan... I'm a bit iffy about the 'default app' idea though - I feel like adding a delay loading settings every time apps are reloaded is a bit of a waste, especially as it's to work around https://github.com/espruino/BangleApps/blob/master/apps/widcloselaunch/widget.js - which is only 17 lines and is only installed by 5 people anyway.

Maybe we could take that out for now and then I can merge this? I'd like to do a 2v16 release soon so it would be good to get it in for that (and yes, we'd totally have to add polyfills, but those should be easy)

gfwilliams:
... also worth adding that for the few people that *do* want to return to something other than the clock, it'd be easy to add some boot code that replaced `Bangle.load` for them.

halemmerich:
Yes you are right about the delay, I had not measured that and `readJSON`seems expensive (~30ms just for the checking code). I will remove that default app stuff for now.
Maybe writing a marker file that could be checked for existence way faster than a setting would be a possible way to do this? A `.default` just containing the file name to launch for example?

gfwilliams:
> Maybe writing a marker file that could be checked for existence way faster than a setting would be a possible way to do this? 

I guess so - but I think at some point we should say that if we have a feature that we don't think 75% of the people would use, and we can implement it as an app with minimal impact, we should probably just do that.

halemmerich:
My initial thought was to replace `load()` in all apps, but that probably is solved cleaner by modifying them to use `setUI` anyway.

gfwilliams:
Thanks! Yes... I think in a way having the separate function is safer as the majority of apps are going to need to be modified to handle this correctly or we'll end up with lots of memory leaks and instability

thyttan:
'Should' I use Bangle.showClock() instead of Bangle.load() for readability's sake?

halemmerich:
Yes, the idea is to have one method to remove duplications for all changes to clock. Additionally `Bangle.showClock` could be easily overloaded in case something should be implemented for all app->clock changes.

PR #2291, opened 2022-11-16
E.showScroller_Q3 - Give coordinates in item to select callback

halemmerich:
Main motivation for this is removing the code duplication from https://github.com/espruino/BangleApps/blob/master/apps/iconlaunch/app.js.


gfwilliams:
Nice idea - thanks! I guess making the backwards compatible could be a pain though

halemmerich:
I think attempting that with Bangle.js 1 Buttons/Touch would remove any speed improvements versus the default launcher anyway :)

halemmerich:
Conditionally wrapping the touch handler and select handler, storing the touch event and `select` parameters and then calling the `select` handler with those as parameters could be enough. Iconlaunch does only need the x value to decide which of the three elements has been selected.

PR #2293, opened 2022-11-22
update esp-idf to v3.3.6

MaBecker:
EspruinoBuildTools have been updated, so this is good to go.


gfwilliams:
Just building this and for anyone else looking, you'll need:

```
rm -rf app
rm -rf esp-idf
rm -rf xtensa-esp32-elf
source scripts/provision.sh ESP32
```

I wonder how many folks that'll catch out :)

Part of me thinks maybe we should have a file like `version.txt` in `app` and `esp-idf` that we can easily check in the provision script (or even Makefile)

PR #2299, opened 2022-12-01
Event 'indicateconfirmed' added

yerpj:
event 'indicateconfirmed' created. Raised upon reception of an 'BLE_GATTS_EVT_HVC' softdevice event.
See https://infocenter.nordicsemi.com/topic/com.nordic.infocenter.s132.api.v7.2.0/msc_inline_mscgraph_90.png for more info.

This event can be used in the application to provide app-to-app data transfer reliability.
Primarily added to prevent any buffer overflow or data loss between the server and client.

### DIFF
--- a/targets/nrf5x/bluetooth.c
+++ b/targets/nrf5x/bluetooth.c
@@ -386,6 +386,11 @@ int jsble_exec_pending(IOEvent *event) {
      }
      break;
    }
+   case BLEP_TASK_CHARACTERISTIC_WRITE_HVC: { 
+     bleQueueEventAndUnLock(JS_EVENT_PREFIX"indicateconfirmed", 0);
+     jshHadEvent();
+     break;
+   }
 #if CENTRAL_LINK_COUNT>0
    case BLEP_RSSI_CENTRAL: { //  rssi as data low byte, index in m_central_conn_handles as high byte
      int centralIdx = data>>8; // index in m_central_conn_handles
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks - so just to confirm:

You have to use `NRF.setServices` to create a characteristic with `indicate` on it, and then when you update it with `updateServices` it sends an indication? And when that is handled, you get this global 'indicateconfirmed' event?

yerpj:
> Thanks - so just to confirm:
> 
> You have to use `NRF.setServices` to create a characteristic with `indicate` on it, and then when you update it with `updateServices` it sends an indication? And when that is handled, you get this global 'indicateconfirmed' event?

That's correct. It should allow for the user app (server side) to transmit reliably packets of data without worry of getting buffer overflow from the softdevice.

gfwilliams:
Thanks - I guess from my point of view it'd be nice if this could call an event on the service itself, like we do for `onWrite` so you can see which one it was for if you have more than one characteristic:

```
NRF.setServices({
  0xBCDE : {
    0xABCD : {
      value : "Hello", // optional
      indicate : true,   // optional, default is false
      onIndicateResponse : function(evt) { // optional

      },
```

I'd have thought you could pretty much copy the code for that...

yerpj:
I totally agree with you, this would be fantastic if linked to a specific service. However, I already struggled at understanding how to expose this event globally, it will take me some time to understand how events could be linked to services.

gfwilliams:
Thanks - ok, I'll take a look when I get a minute. I can't merge this at the moment anyway as it looks like it breaks the Microbit 1 build

PR #2300, opened 2022-12-01
Fixed array sorting with undefined values

### LINKED ISSUE #2294: Array sort of undefined elements
At the moment it looks like undefined elements aren't correctly sorted in an array (they should always go on the bottom https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort), tried this in the ide:
```
>[1,undefined,5,undefined,2].sort((a,b)=>b-a)
=[ 5, 2, 1, undefined, undefined ]
>[1,undefined,5,undefined,2].sort((a,b)=>a-b)
=[ 1, undefined, 2, 5, undefined ] //this doesn't work!
>[1,undefined,5,undefined,2].sort()
=[ 1, 2, 5, undefined, undefined ]
```
Is that a desired behaviour/ something which would stay for some time? Should I make any assumption while doing this kind of sorting?

glemco:
_Hopefully_ fixes #2294


gfwilliams:
Thanks! So this was tested with tests/test_array_sort.js?

Just to confirm though, this is what you mentioned in https://github.com/espruino/Espruino/issues/2294#issuecomment-1327498738 - so it's not actually compliant with the spec, but it does work in the case of a simple comparison?

glemco:
I added a test for sorting both with and without a function (not both direction but it should be fine).
Yes, pretty much, it handles the case in which the compare function returns NaN. If the value is handled internally, users could move the `undefined` around.
It is compliant if the function doesn't handle `undefined` values itself (which is what I'd say most users would do).

If that's undesired I can modified to check for undefined values _before_ even running the compare function.

glemco:
What can possibly go wrong is that if the compare function is dereferencing an object and that's undefined, it can break (note: this is the current behaviour in espruino already plus also in the GNOME js interpreter, node and firefox don't break). E.g.:
```javascript
[undefined, {a:1}, {a:0}].sort((a,b)=>a.a - b.a)
```
In the current implementation (and also after this PR), users may need to check for undefined while comparing anyway.
At this stage it's probably better to strictly follow the specs and don't offer the possibility to sort `undefined` away from the bottom (which looks like a niche feature to me anyway).

gfwilliams:
> At this stage it's probably better to strictly follow the specs and don't offer the possibility to sort undefined away from the bottom (which looks like a niche feature to me anyway).

Yes, thanks - I think it's the right idea to just go with the spec too - so just checking `jsvIsUndefined` on each arg before and returning either -1/1/0 to sort them in the right place would be great

glemco:
> > At this stage it's probably better to strictly follow the specs and don't offer the possibility to sort undefined away from the bottom (which looks like a niche feature to me anyway).
> 
> Yes, thanks - I think it's the right idea to just go with the spec too - so just checking `jsvIsUndefined` on each arg before and returning either -1/1/0 to sort them in the right place would be great

Now it should be fine, I'm not checking for the 0 case as I don't really think it'd matter _which_ `undefined` is going before once two are compared.

gfwilliams:
Looks great - thanks!

PR #2305, opened 2022-12-06
Clear only the corners containing widgets.

arthurgleckler:
Before, the whole top and bottom twenty-four vertical pixels of the screen were cleared whenever any widget appeared.  Now, only the area containing the widgets that appear is cleared.

### DIFF
--- a/libs/js/banglejs/Bangle_drawWidgets_Q3.js
+++ b/libs/js/banglejs/Bangle_drawWidgets_Q3.js
@@ -15,7 +15,9 @@
   	p.c++;
   }
   g.reset();
-  if (pos.tl.c || pos.tr.c) g.clearRect(0,0,w-1,23);
-  if (pos.bl.c || pos.br.c) g.clearRect(0,h-24,w-1,h-1);
+  if (pos.tl.c) g.clearRect(0,0,pos.tl.x,23);
+  if (pos.tr.c) g.clearRect(pos.tr.x,0,w-1,23);
+  if (pos.bl.c) g.clearRect(0,h-24,pos.bl.x,h-1);
+  if (pos.br.c) g.clearRect(pos.br.x,h-24,w-1,h-1);
   try { for (wd of WIDGETS) wd.draw(wd); } catch(e) {print(e);}
 })


gfwilliams:
Thanks - although I am a little worried that this may end up causing screen corruption in apps. For example:

```
Bangle.drawWidgets = (function() {
  if (!global.WIDGETS) return;
  var w=g.getWidth(), h=g.getHeight(), pos={
	  tl:{x:0, y:0, r:0, c:0}, // if r==1, we're right->left
	  tr:{x:w-1, y:0, r:1, c:0},
	  bl:{x:0, y:h-24, r:0, c:0},
	  br:{x:w-1, y:h-24, r:1, c:0}
  }, p;
  for (var wd of WIDGETS) {
    p = pos[wd.area];
  	if (!p || wd.width == 0) continue;
  	wd.x = p.x - p.r*wd.width;
  	wd.y = p.y;
  	p.x += wd.width*(1-2*p.r);
  	p.c++;
  }
  g.reset();
  if (pos.tl.c) g.clearRect(0,0,pos.tl.x,23);
  if (pos.tr.c) g.clearRect(pos.tr.x,0,w-1,23);
  if (pos.bl.c) g.clearRect(0,h-24,pos.bl.x,h-1);
  if (pos.br.c) g.clearRect(pos.br.x,h-24,w-1,h-1);
  try { for (wd of WIDGETS) wd.draw(wd); } catch(e) {print(e);}
});


E.showMessage("Hello", "Title");
Bangle.loadWidgets();
Bangle.drawWidgets();
E.showMessage("Hello", "Title");
```

image

I think quite a lot more work would need to be done to test and modify apps and other functions before we can really do this.

Out of interest, why do you want it? Do you have some kind of clock that you wanted to draw partially in the widget bar?

gfwilliams:
Just to add, a slightly less contrived example is:

```
// new drawWidgets

g.fillRect(0,0,176,176); // or anything in the background
E.showMenu({
  "< Back": function() {},
  one : function() {},
  two : function() {},
  three : function() {},
  four : function() {},
  five : function() {},
  six : function() {},
});
```

Issue #2308, opened 2022-12-29
Bangle.js: get poll interval


notEvil:
Hi,

the powerSave option triggers a change in poll interval but there is no way to detect this change except for timing the events triggered by the poll handler. Use case: use "low poll rate" ~= "inactivity" ~= "not worn" to disable certain actions (buzz, HRM, ...) in apps/widgets.

I'm happy to prepare a PR if requested.

gfwilliams:
Yes, that could be a good idea.

Does it make sense to just add this to `Bangle.getOptions()` ?

notEvil:
I tried this already but don't think its a good idea. `Bangle.getOptions` internally calls the same function as `Bangle.setOptions`, and options never change except by `Bangle.setOptions` (I think). Adding poll rate to the options would break those implicit rules.

What about just `Bangle.getPollInterval()`?

gfwilliams:
Good point. We could just add getPollInterval, I guess it's ok.

I'm just a bit anxious about adding a whole bunch of extra functions as it makes it harder to find what you're after. It's also basically impossible for me to remove a method after it's been added :)

There is http://www.espruino.com/Reference#l_Bangle_dbg which returns debug info. It would make a lot of sense to add it in there I guess

notEvil:
I agree. I'll prepare a PR with `Bangle.dbg().pollInterval`

gfwilliams:
brilliant - thanks!

PR #2311, opened 2023-01-05
Add `pollInterval` to `Bangle.dbg`

notEvil:
see https://github.com/espruino/Espruino/issues/2308

Didn't test but don't expect any issues.


gfwilliams:
Great, thanks! And thanks for adding to the docs too - they were definitely lacking!

PR #2317, opened 2023-01-18
fix LINUX crash on reset() with fixed variables size

fanoush:
LINUX defines JSVAR_MALLOC even for fixed variables and at reset() time jsvInit is called with size 0 so we need jsVarsSize filled with default value to malloc right size

alternatively we could avoid it by not defining JSVAR_MALLOC when variable size is fixed for LINUX here https://github.com/espruino/Espruino/blob/master/scripts/build_platform_config.py#L285
that seem to not trigger this bug


gfwilliams:
Thanks! I think this is a good fix anyway

PR #2322, opened 2023-02-08
Documentation/generated typescript updates - Object, Bangle and optionals

bobrippling:
This updates a few methods on `Bangle`, `Object` and `Graphics`, along with changing documented optional properties/arguments to match the `"[optional]"` format used in [build_types.js](https://github.com/espruino/Espruino/blob/149a7f6bace3a56c1892de47bd8baacfb3f496a5/scripts/build_types.js#L132-L132)


bobrippling:
I'm happy to also update the definitions in [`BangleApps`](https://github.com/espruino/BangleApps/) if this is merged

PR #2325, opened 2023-02-13
Free in reverse order

notEvil:
Currently, StringExt blocks as well as children are freed in order. This leads to backward links, e.g.

1: String
2: StringExt
3: StringExt

creates 1 -> 3 -> 2 and

1: Object
2: first name or index
3: first value
4: second name or index
5: second value

creates 1 -> 4 -> 5 -> 2 -> 3

This PR reverses the order. It shouldn't have a significant performance impact but lead to less fragmentation and less frequent GC

### DIFF
--- a/src/jsvar.c
+++ b/src/jsvar.c
@@ -631,21 +631,22 @@ static void jsvFreePtrInternal(JsVar *var) {
   jshInterruptOn();
 }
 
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks for that recent change - this looks great!

notEvil:
You are most welcome :)

PR #2327, opened 2023-02-13
Implement `Object.fromEntries`

Dabolus:
This PR adds support for [`Object.fromEntries`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries) from ES2019.

Opening as a draft because I still have to figure out how to properly run tests.

### DIFF
--- a/src/jswrap_object.c
+++ b/src/jswrap_object.c
@@ -356,6 +356,38 @@ JsVar *jswrap_object_values_or_entries(JsVar *object, bool returnEntries) {
   return cbData[0];
 }
 
+/*JSON{
+  "type" : "staticmethod",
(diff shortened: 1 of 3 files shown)


gfwilliams:
Thanks - I think running tests is as easy as building for Linux and running `./espruino --test ...`

But this code may be broken - I think jsvObjectSetChildAndUnLock expects a string as a second argument and also key would need to be unlocked anyway in the jsvObjectSetChildAndUnLock case.

gfwilliams:
Had to do a few fixes for this, but it's added now

Dabolus:
Thank you! 🥳

PR #2330, opened 2023-02-20
Update String::slice() type - make end argument optional

bobrippling:
A small adjustment to the existing typescript type

(left as draft for now, in case I come across any more tweaks)


gfwilliams:
You think we're ok to merge or do you want to leave this for a while?

bobrippling:
Yes I think we're ok to merge this, and thanks!

PR #2333, opened 2023-02-25
Update Widget type

bobrippling:
This includes the `x` & `y` present when a widget's been drawn, along with the optional `sortorder`.

The `draw` function takes `this` as the widget, but also an argument of `Widget` too, as we setup the call to [pass the widget in both cases](https://github.com/bobrippling/Espruino/blob/1876343c8bf61aeac907fdff2e1c0d1d91323a92/libs/js/banglejs/Bangle_drawWidgets.js#L20-L20)


gfwilliams:
Thanks!

PR #2334, opened 2023-02-25
Update this reference in call/apply/bind types

bobrippling:
`this` as the first argument means the function is callable as in `obj.func()`, and that we've specified the type of `obj`, rather than the type of the first argument.

We can rename `this` to `thisArg` to avoid this behaviour, meaning we now have a "real" first argument, that will be the `this` passed to the function we've specified.

This change also restricts `apply`'s second argument to be an array-like object.

### DIFF
--- a/src/jswrap_object.c
+++ b/src/jswrap_object.c
@@ -1141,7 +1141,7 @@ void jswrap_function_replaceWith(JsVar *oldFunc, JsVar *newFunc) {
   "name" : "call",
   "generate" : "jswrap_function_apply_or_call",
   "params" : [
-    ["this","JsVar","The value to use as the 'this' argument when executing the function"],
+    ["thisArg","JsVar","The value to use as the 'this' argument when executing the function"],
     ["params","JsVarArray","Optional Parameters"]
   ],
   "return" : ["JsVar","The return value of executing this function"]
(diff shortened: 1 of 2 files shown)


src/jswrap_object.c
```diff
@@ -1157,10 +1157,13 @@ This executes the function with the supplied 'this' argument and parameters
   "name" : "apply",
   "generate" : "jswrap_function_apply_or_call",
   "params" : [
-    ["this","JsVar","The value to use as the 'this' argument when executing the function"],
-    ["args","JsVar","Optional Array of Arguments"]
+    ["thisArg","JsVar","The value to use as the 'this' argument when executing the function"],
+    ["args","JsVarArray","Optional Array of Arguments"]
```
gfwilliams, line 1161: Have you tested this? I'm pretty sure this breaks the build.

These headers are not just for Typescript and documentation - they tell the interpreter what to expect. And JsVarArray isn't a JsVar that's an array, it's more like a spread operator

gfwilliams:
We can do this, but can you not just change the script that makes the typescript decls to just convert `this`->`thisArg` if it finds it?

Because I'd imagine we'll find that maybe more functions get added using `this` as a name later on - and there is no real reason not to use it other than the typescript decls?

src/jswrap_object.c
bobrippling: Ah yes - I had totally missed that, thank you

PR #2337, opened 2023-03-01
Add `getSecurityStatus().advertising` property

bobrippling:
As mentioned in espruino/BangleApps#2608, this adds a `advertising` property to the object returned by `NRF.getSecurityStatus()`, allowing us to determine whether the chip is idle, advertising or connected.

### DIFF
--- a/targets/nrf5x/bluetooth.c
+++ b/targets/nrf5x/bluetooth.c
@@ -3100,6 +3100,8 @@ JsVar *jsble_get_security_status(uint16_t conn_handle) {
     jsvObjectSetChildAndUnLock(result, "encrypted", jsvNewFromBool(status.encrypted));
     jsvObjectSetChildAndUnLock(result, "mitm_protected", jsvNewFromBool(status.mitm_protected));
     jsvObjectSetChildAndUnLock(result, "bonded", jsvNewFromBool(status.bonded));
+    bool isAdvertising = bleStatus & BLE_IS_ADVERTISING;
+    jsvObjectSetChildAndUnLock(result, "advertising", jsvNewFromBool(isAdvertising));
 #ifndef SAVE_ON_FLASH
     if (status.connected && conn_handle==m_peripheral_conn_handle)
       jsvObjectSetChildAndUnLock(result, "connected_addr", bleAddrToStr(m_peripheral_addr));
--- a/libs/bluetooth/jswrap_bluetooth.c
+++ b/libs/bluetooth/jswrap_bluetooth.c
@@ -3454,13 +3454,27 @@ void jswrap_ble_setSecurity(JsVar *options) {
   }
 }
 
+/*TYPESCRIPT
+type NRFSecurityStatus = {
+  connected: false
+} | {
+  connected: boolean,
+  encrypted: boolean,
+  mitm_protected: boolean,
+  bonded: boolean,
+  advertising: boolean,
+  connected_addr?: string,
+};
+*/
+
 /*JSON{
     "type" : "staticmethod",
     "class" : "NRF",
     "name" : "getSecurityStatus",
     "ifdef" : "NRF52_SERIES",
     "generate" : "jswrap_ble_getSecurityStatus",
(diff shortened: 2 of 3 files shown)


bobrippling:
I might've gone down totally the wrong path here with obtaining the BLE advertising status - I figure it's not related to a connection and so picked it up from the `ble_advertising` module, rather than modifying `ble_conn_state`, but let me know if there's a better way.

Also waiting on the github firmware builds, I suspect I might need a few `ifdef`s yet.

targets/nrf5x/bluetooth.c
```diff
@@ -3100,6 +3100,9 @@ JsVar *jsble_get_security_status(uint16_t conn_handle) {
     jsvObjectSetChildAndUnLock(result, "encrypted", jsvNewFromBool(status.encrypted));
     jsvObjectSetChildAndUnLock(result, "mitm_protected", jsvNewFromBool(status.mitm_protected));
     jsvObjectSetChildAndUnLock(result, "bonded", jsvNewFromBool(status.bonded));
+#if NRF_MODULE_ENABLED(BLE_ADVERTISING)
```
bobrippling, line 3103: I suspect this isn't a macro issue, but that I need to ditch the `#if` and rebuild the local `nrf5x_12` component, then update the headers that this target includes (assuming they're done as part of the build process). Will look into this further

gfwilliams: Thanks, but I wouldn't worry about ifdef-ing this as all BLE devices will be able to advertise

gfwilliams:
Thanks - but I wouldn't change the NRF52 APIs themselves - Bangle.js uses the v15 API anyway.

What I'd intended was you use:

* https://github.com/espruino/Espruino/blob/master/libs/bluetooth/bluetooth.h#L110
* https://github.com/espruino/Espruino/blob/master/libs/bluetooth/bluetooth.h#L192

so just `bleStatus&BLE_IS_ADVERTISING` - no need for a special function.

targets/nrf5x/bluetooth.c
bobrippling: Cheers - not an issue with the new approach now too

bobrippling:
Thanks, that's a much easier route - let me know if there's any other changes you'd prefer, I've popped in some typescript for the security status too

gfwilliams:
Thanks! But would it be possible to put that code in `jsble_get_security_status` instead? That's where we set all the other fields, and it would avoid the duplication (although it would still have to be copied for NRF52/ESP32 I guess)

bobrippling:
Ah yeah of course - I was thinking `bleStatus` belonged to `libs/bluetooth/` and I wouldn't be able to include it from "down the chain" if you get me, but I hadn't looked to confirm where it was! Thanks again for your guidance

targets/nrf5x/bluetooth.c
```diff
@@ -3100,6 +3100,8 @@ JsVar *jsble_get_security_status(uint16_t conn_handle) {
     jsvObjectSetChildAndUnLock(result, "encrypted", jsvNewFromBool(status.encrypted));
     jsvObjectSetChildAndUnLock(result, "mitm_protected", jsvNewFromBool(status.mitm_protected));
     jsvObjectSetChildAndUnLock(result, "bonded", jsvNewFromBool(status.bonded));
+    bool isAdvertising = bleStatus & BLE_IS_ADVERTISING;
```
bobrippling, line 3103: Or could this be this, to be more accurate?

```suggestion
    bool isAdvertising = bleStatus & (BLE_IS_ADVERTISING | BLE_IS_ADVERTISING_MULTIPLE);
```

(also, happy to remove the intermediate variable if preferred)

gfwilliams:
Thanks for those changes - this looks great now!

targets/nrf5x/bluetooth.c
bobrippling: Great, thanks again!

PR #2340, opened 2023-03-06
Update `NRFSecurityStatus` type

bobrippling:
This updates the `NRFSecurityStatus` type from a prior commit (#2337), always exposing `advertising` and clarifying other properties depending on `connected`'s value - sorry about the double PR!


gfwilliams:
thanks!

PR #2343, opened 2023-03-10
Fix `drawString` and `setClipRect` with 90/270-degree rotation

dzhu:
Previously, the calculation of `minX`, `minY`, `maxX`, and `maxY` inside `jswrap_graphics_drawString` was what `graphicsToDeviceCoordinates` does, just written out explicitly. With 90- or 270-degree rotations, that meant that the final values ended up rotated 180 degrees relative to what they should have been. As a result, characters could end up effectively clipped twice and hence incorrectly not being drawn.  What we actually want instead is to do the _reverse_ transformation of `graphicsToDeviceCoordinates`, which fortunately already exists as `deviceToGraphicsCoordinates`.

The PR includes a test that would fail before and is fixed by this change.


gfwilliams:
Thanks - this looks great! I was sure there was a reason it was done that way, but I can't remember now - maybe it's just that `deviceToGraphicsCoordinates` didn't exist at the time or I'd forgotten about it...

PR #2344, opened 2023-03-13
[jswrap_bangle.c] add info re touch event xy.type

thyttan:
I've referenced this forum post: https://forum.espruino.com/conversations/384200/

... and my testing of the touch event on my Bangle.js 2 via the Web IDE.


gfwilliams:
Thanks!

PR #2347, opened 2023-04-04
[E.showMenu_Q3] (wip) add setClipRect to avoid draw over widgets

thyttan:
WIP

No extensive testing done yet, but seems to work in the case I've tested.

One solution to https://github.com/espruino/BangleApps/issues/2670#issuecomment-1496426671


gfwilliams:
Sorry, but I'm afraid this won't work:

* I believe ClipRect  is set by the scroller specifically to avoid redrawing lots of extra stuff when scrolling to make it faster - setting it manually to the apprect undoes all that work, and it also allows menu items to then overwrite other menu items
* You don't ever reset cliprect so it'll stay set after drawing even once, widgets will then not be able to update themselves

I think what I'd suggested on https://github.com/espruino/BangleApps/issues/2670 - either redrawing the widgets if it overlaps them, or scrolling onto the screen would be more reliable

thyttan:
Thank you for the explanation! 🙂

PR #2348, opened 2023-04-07
Split `setUI` types into overloaded functions

bobrippling:
This prevents erroneous usage, e.g. a callback with "clock" or expecting `undefined` to be passed to a "clockupdown" callback.

It also "blesses" `{mode: "clock" | "updown" | ...}` with a subset of options which are handled by `setUI` (`remove` and `back`).

Also make `removeAllListeners()`'s first argument optional.


gfwilliams:
Great - thanks!

PR #2350, opened 2023-04-18
add pipe method to StorageFile

mariusGundersen:
This adds the pipe method to storage files, so they can be piped to for example an http request or response. 

I have tested and this seems to work:
```js
const logFile = Storage.open("log", "r");

// without this change
E.pipe(logFile, req);

// with this change:
logFile.pipe(req)
```

### DIFF
--- a/src/jswrap_storage.c
+++ b/src/jswrap_storage.c
@@ -927,3 +927,18 @@ void jswrap_storagefile_erase(JsVar *f) {
   jsvObjectSetChildAndUnLock(f,"addr",jsvNewFromInteger(0));
   jsvObjectSetChildAndUnLock(f,"mode",jsvNewFromInteger(0));
 }
+
+
+/*JSON{
+  "type" : "method",
+  "class" : "StorageFile",
+  "name" : "pipe",
+  "ifndef" : "SAVE_ON_FLASH",
+  "generate" : "jswrap_pipe",
+  "params" : [
+    ["destination","JsVar","The destination file/stream that will receive content from the source."],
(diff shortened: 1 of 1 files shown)


src/jswrap_storage.c
```diff
@@ -927,3 +927,18 @@ void jswrap_storagefile_erase(JsVar *f) {
...
+/*JSON{
+  "type" : "method",
+  "class" : "StorageFile",
+  "name" : "pipe",
+  "ifndef" : "SAVE_ON_FLASH",
+  "generate" : "jswrap_pipe",
+  "params" : [
+    ["destination","JsVar","The destination file/stream that will receive content from the source."],
+    ["options","JsVar",["[optional] An object `{ chunkSize : int=32, end : bool=true, complete : function }`","chunkSize : The amount of data to pipe from source to destination at a time","complete : a function to call when the pipe activity is complete","end : call the 'end' function on the destination when the source is finished"]]
+  ]
```
bobrippling, line 941: We could type that argument if you like:
```suggestion
    "typescript": "pipe(dest: any, options?: { chunkSize?: number, end?: boolean, complete?: () => void }): void"
  ]
```

mariusGundersen: We should probably do that in all the places it's defined: https://github.com/espruino/Espruino/search?q=jswrap_pipe

gfwilliams: Maybe it makes sense for you to do another PR for that @bobrippling ?

gfwilliams:
This looks great, thanks! I guess just having the `read(num)` method is good enough to make it work.

I'm wondering whether it makes sense to allow `E.pipe` to take a String as a source? It would then allow piping of large (non-StorageFile) files in storage to a socket - which I guess you can't easily do

src/jswrap_storage.c
bobrippling: Sure, it would be my pleasure

PR #2358, opened 2023-05-09
Add function deepSleepExt0

cwilling:
This PR adds ability for ESP32 boards to wake from deep sleep by trigger from external interrupt (setting a GPIO pin). It may be confirmed with this example code:
```
const buttonPin = D4;
pinMode(buttonPin, 'input');

// Flash LED1 (while awake)
var on = false;
setInterval(function() {
  on = !on;
  LED1.write(on);
}, 500);

// Every 10secs, sleep until buttonPin is pressed
setInterval(function() {
  console.log("Time to sleep");
  ESP32.deepSleepExt0(buttonPin, 1);
}, 10000);
```

### DIFF
--- a/targets/esp32/jswrap_esp32.c
+++ b/targets/esp32/jswrap_esp32.c
@@ -88,6 +88,24 @@ void jswrap_ESP32_deepSleep(int us) {
   esp_deep_sleep_start(); // This function does not return.
 } // End of jswrap_ESP32_deepSleep
 
+/*JSON{
+  "type"     : "staticmethod",
+  "class"    : "ESP32",
+  "ifdef" : "ESP32",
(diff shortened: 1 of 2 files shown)


gfwilliams:
Looks great - thanks! Very happy to have stuff like this if it makes the ESP32 port more usable.

I'm merging now, but quick question: Can `esp_sleep_enable_ext0_wakeup` return an error code (eg if the wrong pin is specified like `esp_sleep_enable_ext0_wakeup(700, ...)`?) 

If so it might be worth checking it and doing `jsExceptionHere(JSET_ERROR, "Invalid Arguments")` rather than trying to sleep the ESP32 with no way to wake it?

PR #2359, opened 2023-05-10
Add function getWakeupCause for ESP32

cwilling:
This function returns a number representing the reason that a device has woken from deep sleep. Example code:
```
var wakeupCause = ESP32.getWakeupCause();
if (wakeupCause == 0) {
  // Assume a reboot i.e.completely new start
  OnInit(); 
} else {
  // Wakeup from deep sleep.
  // Could check values 2-6 for explicit wakeup source
  OnWakeup();
}
```


gfwilliams:
Looks great, thanks!

I really appreciate the docs in the JSON header too - those will make big difference when folks go to use it

PR #2367, opened 2023-05-10
for SD>5 use static buffers for advertising and scan response data

### LINKED ISSUE #2000: advertising broken with S140 >= 6.1.0
with 6.0.0 it works but with 6.1.0 and up no device is visible. Found the issue and solution described here

https://devzone.nordicsemi.com/f/nordic-q-a/37473/advertising-does-not-start-with-s140-6-1-0-same-hex-run-without-issue-with-s140-6-0-0/145157#145157

without understanding details of espruino advertising data lifecycle, quick fix like this seems to work (= device is visible so I can at least connect to console over bluetooth repeatedly)
```
diff --git a/targets/nrf5x/bluetooth.c b/targets/nrf5x/bluetooth.c
index b667c0ff..1d9815da 100644
--- a/targets/nrf5x/bluetooth.c
+++ b/targets/nrf5x/bluetooth.c
@@ -2513,7 +2513,7 @@ uint32_t jsble_advertising_start() {
   adv_params.interval = bleAdvertisingInterval;

   uint32_t err_code = 0;
-  uint8_t m_enc_scan_response_data[31]; // BLE_GAP_ADV_SET_DATA_SIZE_MAX
+static  uint8_t m_enc_scan_response_data[31]; // BLE_GAP_ADV_SET_DATA_SIZE_MAX
   uint16_t m_enc_scan_response_data_len = sizeof(m_enc_scan_response_data);
 #if NRF_SD_BLE_API_VERSION<5
   err_code = adv_data_encode(&scanrsp, m_enc_scan_response_data, &m_enc_scan_response_data_len);
@@ -2528,8 +2528,10 @@ uint32_t jsble_advertising_start() {

   //jsiConsolePrintf("adv_data_set %d %d\n", advPtr, advLen);
 #if NRF_SD_BLE_API_VERSION>5
-  ble_gap_adv_data_t d;
-  d.adv_data.p_data = (uint8_t*)advPtr;
+static  ble_gap_adv_data_t d;
+static  uint8_t m_adv_data[BLE_GAP_ADV_SET_DATA_SIZE_MAX];
+memcpy(m_adv_data,advPtr,MIN(advLen,BLE_GAP_ADV_SET_DATA_SIZE_MAX));
+  d.adv_data.p_data = m_adv_data;
   d.adv_data.len = advLen;
   d.scan_rsp_data.p_data = m_enc_scan_response_data;
   d.scan_rsp_data.len = m_enc_scan_response_data_len;
```
however I guess there is maybe better way?  the
`JSV_GET_AS_CHAR_ARRAY(advPtr, advLen, advDataVar);` could read it into static array directly?

fanoush:
S140 6.1.x and up no longer makes its own copy of advertising data, provided buffers need to be valid all the time, also any change needs new set of buffers (otherwise it fails with NRF_ERROR_INVALID_STATE)

fixes #2000


targets/nrf5x/bluetooth.c
```diff
@@ -2524,16 +2531,18 @@ uint32_t jsble_advertising_start() {
...
   if (!non_scannable) {
-    d.scan_rsp_data.p_data = m_enc_scan_response_data;
-    d.scan_rsp_data.len = m_enc_scan_response_data_len;
+    memcpy(m_scan_rsp_data, m_enc_scan_response_data, m_enc_scan_response_data_len);
+    m_ble_gap_adv_data.scan_rsp_data.p_data = m_scan_rsp_data;
+    m_ble_gap_adv_data.scan_rsp_data.len = m_enc_scan_response_data_len;
   }
 
-  err_code = sd_ble_gap_adv_set_configure(&m_adv_handle, &d, &adv_params);
+  err_code = sd_ble_gap_adv_set_configure(&m_adv_handle, &m_ble_gap_adv_data, &adv_params);
```
gfwilliams, line 2545: I know this is the first use so `jsble_advertising_update` wouldn't work as-is, but it'd be nice if we could find a way to remove the duplication here (and in the ifdefs below where we do sd_ble_gap_adv_data_set as well).

I guess if we `m_ble_gap_adv_data` was all zero'd out it'd be good enough.

libs/bluetooth/jswrap_bluetooth.c
```diff
@@ -1164,25 +1164,7 @@ void jswrap_ble_setScanResponse(JsVar *data) {
...
-    d.adv_data.p_data = (uint8_t *)advPtr;
-    d.adv_data.len = advLen;
-    d.scan_rsp_data.p_data = (uint8_t *)respPtr;
-    d.scan_rsp_data.len = respLen;
-    if (bleStatus & BLE_IS_ADVERTISING) sd_ble_gap_adv_stop(m_adv_handle);
-    err_code = sd_ble_gap_adv_set_configure(&m_adv_handle, &d, NULL);
-    if (bleStatus & BLE_IS_ADVERTISING) sd_ble_gap_adv_start(m_adv_handle, APP_BLE_CONN_CFG_TAG);
-    jsvUnLock(advData);
-#endif
+  err_code=jsble_advertising_update_scanresponse((uint8_t *)respPtr, respLen);
```
gfwilliams, line 1167: Any chance we could just remove the NRF5X define as well, and just dump a jsble_advertising_update_scanresponse placeholder in for ESP32?

gfwilliams:
Thanks! Looks really good to me - happy to merge as-is (assuming this doesn't break stuff?) but it could also be a good opportunity to really tidy this up

targets/nrf5x/bluetooth.c
fanoush: m_ble_gap_adv_data  is in bss so is zeroed out on startup but then it is not after stopping advertising or any time later, this one does not need to be static for SD>6.1 to work but it keeps enough data for making the temporary copy and when you want to set just the other half

libs/bluetooth/jswrap_bluetooth.c
fanoush: yes, could do that, I can just put that DEAD FIXME stuff into esp32 implementation of that method?

fanoush: done

targets/nrf5x/bluetooth.c
gfwilliams: Leaving the structure is fine by me - also the JS vars do get tricky when we're trying to configure stuff as we tear down and restart the JS engine :)

What I meant was just whether we could use `jsble_advertising_update` here again to reduce some duplication since they are almost identical.

libs/bluetooth/jswrap_bluetooth.c
fanoush: oh, but it is reported as conflict as  I branched from older version and next unrelated lines below are different, not sure how to fix that except moving it elsewhere, I don't want to update this branch now, I'll try to move it

targets/nrf5x/bluetooth.c
fanoush: oh, got it, but don't know how, it does set it twice with temporary copy because the guess is that it is already advertising at that point but that may not be true. when properly knowing it is not advertising it could skip that and then it could be used

fanoush: could probably use ` bleStatus & BLE_IS_ADVERTISING`, will test

libs/bluetooth/jswrap_bluetooth.c
gfwilliams: no problem - i merged it here (I hope) - github actions seem broken at the moment

targets/nrf5x/bluetooth.c
gfwilliams: Or I was thinking if both the lengths of data are zero (eg m_ble_gap_adv_data was memset(0) or cleared by reboot) then it could assume there was no need for that first call.

gfwilliams: Sounds good - thanks!

fanoush:
So I reworked it to reuse the jsble_advertising_update also when starting the advertising.
I tested with 52840 S140 6.1.1 and 52832 S132 3.1.0. I also tried to set array of advertising data so that it changes.  I also tried t osen the connectable and scannable to false. For some reason when setting scannable true, connectable false it does look in nrfconnect as connectable. Maybe it is BLE thing and invalid combination? Anyway it seems to do the same before the change. Only connectable:false,scannable:false looks as not connectable but then I could not set non empty scan response, empty one worked `NRF.setScanResponse([])`. So not sure if you can have non connectable device with scan response to keep more data there.

AFAIK the change in behaviour in this PR is 
- `non_conectable` now does not set scan response for all SD versions, previously only >5
- NRF.setScanResponse now does not start/stop advertising (and seem to work without it), it also does not take advertising data from jswrap_ble_getCurrentAdvertisingData for SD>5 but keeps previous static buffer

fanoush:
> Is there any way to change that easily?

oh, ok, I asked about that change here https://github.com/espruino/Espruino/pull/2367#discussion_r1190792913 was not sure why there is such code, BTW it does not affect NRF.setResponse so maybe it still covers your case? Only when starting advertising it does that, can you actually provide scan response to that? And as mentioned in previous comment this combination does not work properly, at least nrfconnect still shows this as connectable in flags and shows connect button even with current firmware so not sure if this is even valid combination.

Anyway I can ifdef that bit too if needed.
It is true that NRF.setScanResponse call with nonempty array does indeed throw ERR 07 on SD6.x  in non scannable case and it did not show that arror on 52832 but still I think the scanresponse data was not there, will check again.

gfwilliams:
Ahh sorry - I thought that was about non_scannable, but you were talking about non_connectable?

I just tried `NRF.setAdvertising({},{connectable:false, scannable:true}); ` on an nRF52DK with SDK12 and it appears to work. There is scan data, and while there is a 'Connect' button still, if you click it you can't actually connect.

I'm not 100% sure what could be done to make it advertise as not connectable since as far as I can tell from the advertising data, it's not advertising anything saying it's connectable or not (there's nothing in the BLE flags for it).

gfwilliams:
Ahh - thanks! Yes, that makes more sense - yes, if it's nonscannable we shouldn't have to worry about scan response.

Thanks for this! I'll merge and hopefully we'll have some time to see if any issues surface

PR #2368, opened 2023-05-10
Test validity of wakeup pin for ESP32 deep sleep

cwilling:
PR #2358 added wakeup from deep sleep by setting a GPIO pin. However not all GPIOs are valid wakeup pins and could result in a deep sleep without any way to wake up. This PR attempts to ensure that only a valid RTC GPIO is selected before sending the device into deep sleep.


gfwilliams:
Looks great - thanks!

PR #2369, opened 2023-05-12
Add function deepSleepExt1 for EXP32

cwilling:
An array of RTC enabled GPIOs can be used to wake from deep sleep. Example code:

// Flash LED1 (while awake)
var on = false;
setInterval(function() {
  on = !on;
  LED1.write(on);
}, 500);

// Every 10secs, sleep until any of pinArray is pressed
setInterval(function() {
  var pinArray = [D4,D15];

  ESP32.deepSleepExt1(pinArray, 1);
}, 10000);

### DIFF
--- a/targets/esp32/jswrap_esp32.c
+++ b/targets/esp32/jswrap_esp32.c
@@ -138,21 +138,19 @@ as RTC GPIOs](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-re
 void jswrap_ESP32_deepSleep_ext1(JsVar *pinVar, JsVarInt mode) {
   uint64_t pinSum = 0;
   if (jsvIsArray(pinVar)) {
-    JsVarRef pinName = jsvGetLastChild(pinVar);
-    while (pinName) {
-      JsVar *pinNamePtr = jsvLock(pinName);
-      JsVar *pinPtr = jsvSkipName(pinNamePtr);
-      Pin pin = jshGetPinFromVar(pinPtr);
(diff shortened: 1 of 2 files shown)


gfwilliams:
Thanks! For the array, please could you use `jsvIterator` instead?

```
  JsvIterator it;
  jsvIteratorNew(&it, array, JSIF_DEFINED_ARRAY_ElEMENTS);
  while (jsvIteratorHasElement(&it)) {
   ... jsvIteratorGetValue(&it);
    jsvIteratorNext(&it);
  }
  jsvIteratorFree(&it);
```

There are some plans to change the way arrays are stored eventually (to make them a tree), so when doing that using LastChild would break.

gfwilliams:
That was quick! But sorry - just spotted some other minor things...

* Can `pinSum = (uint64_t)pow(2, pin);` be `pinSum = 1<<pin;` in the two places where it's done? There's no need to drag in double arithmetic for it when a shift will do.
* `Pin pin = jshGetPinFromVar(jsvIteratorGetValue(&it));` will actually cause a memory leak because the return value from jsvIteratorGetValue isn't unlocked. It's not a big deal as we're shutting everything down, but it's not good practise esp if someone copy/pastes it for something else. All you need to do is change it to `jshGetPinFromVarAndUnLock(jsvIteratorGetValue(&it));`

cwilling:
Done. Thanks for the tips.

gfwilliams:
Great! Thanks!

PR #2373, opened 2023-05-21
Add typings for `setTimeout`/`setInterval` and co.

bobrippling:
This uses branding to help to catch issues where we might pass the wrong number, or a function to `clearInterval` and related functions.


gfwilliams:
Thanks!

PR #2378, opened 2023-06-11
readJSON typescript: prevent `any` propagation

bobrippling:
This tweaks the `readJSON` type to indicate that (when we're in the object-or-exception case), we get back `unknown`, prompting the developer to cast the json to the appropriate type, rather than `any` making its way through their code.

In the non-exception case of `readJSON(..., 1)`, we also have `undefined` as a possible return value.


gfwilliams:
Thanks!

Issue #2383, opened 2023-06-17
btnHandlerCommon block when LCD off


d3nd3:
`btnHandlerCommon` blocking when screen off. I created a pull request: #2384

Bangle 1 user.

d3nd3:
Hi, I didn't see you posted here.

I understand that the default desire is that the watch functions best like that. That state is preserved, if lockTimeout is the same value as lcdPowerTimeout, the device locks and turns display off at same time.  I see that this gives more flexibility to the lockTimeout for users who do not want to use that default safer state.

Hope that it makes more sense.

d3nd3:
But the backlight stuff will solve this posts issue:
[https://forum.espruino.com/conversations/387477/#comment17014744](https://forum.espruino.com/conversations/387477/#comment17014744)

And another stab at explaining the need for the lock thing:
If I have lcd power save on (any of lockTimeout,backlightTimeout,lcdTimeout > 0 ) ... then the code at :
https://github.com/espruino/Espruino/blob/491288d9ffb88150e4213f94c4ed30c99380b90b/libs/banglejs/jswrap_bangle.c#LL1598C1-L1605C4 is hit.

So the only way to get unlock whilst screen is off, is to have all (lockTimeout,backlightTimeout,lcdTimeout == 0) , equivalent to setLCDTimeout(0).  I do not want to call setLCDTimeout(0), I want to invidually control the timeouts, eg. lcdTimeout 5000, lockTimeout 0.  But currently not possible, because of how the code is arranged.

gfwilliams:
I think this can be closed now?

PR #2384, opened 2023-06-18
allow non-wake-binded inputs to propogate when is screen off

d3nd3:
So to keep it short & brief, the idea is to allow inputs that aren't wake-key-binded inputs, to not be blocked when screen is off. Believing that the `if locked dont push` is already enough control.

Also put `lcdPowerTimeout &&` in some places where missing, so it consistently behaves the same, requiring a timeout if its going to perform any `auto-wake-action`.

Thought it'd be nice to have a backlight api to go along with the timeout.


d3nd3:
The first 2 commits redundant, changed method. Simply set inactivity timer to 0 if unlocked when dealing with non-wake-bind inputs.  This allows lockTimeout to be larger than LCDPowerTimeout, so there is input whilst display is off.

d3nd3:
I tried to apply the same simple logic to the touch handlers, including banglejs2's handler, you'd have to tweak some things because I do not have enough expert knowledge. But I hope this could save you some time if you like the idea.

d3nd3:
I have not tested any of these code changes, proper care needs to be taken to ensure its bug free.

d3nd3:
Had a commit hell, just cleaned it up lol.

gfwilliams:
> I have not tested any of these code changes, proper care needs to be taken to ensure its bug free.

Do you think you *could* test them?? As long as GitHub actions are enabled then the firmware is automatically built - for example https://github.com/espruino/Espruino/actions/runs/5307613873?pr=2384

libs/banglejs/jswrap_bangle.c
```diff
@@ -1562,6 +1574,7 @@ void backlightOffHandler() {
 #endif // !EMULATED
 
 void btnHandlerCommon(int button, bool state, IOEventFlags flags) {
+  bool pushEvent = true;
```
gfwilliams, line 1577: Is this var used?

d3nd3: Yes, I thought it made more sense to not use `return` statements in this function so that the `if (button == lcdWakeButton) {` code can be reached.  `if (bangleFlags&JSBF_LOCKED) pushEvent = false;` instead of `if (bangleFlags&JSBF_LOCKED) return` .  Not super important but makes the feeling of writing/changing the function more stress-free?

d3nd3:
> > I have not tested any of these code changes, proper care needs to be taken to ensure its bug free.
> 
> Do you think you _could_ test them?? As long as GitHub actions are enabled then the firmware is automatically built - for example https://github.com/espruino/Espruino/actions/runs/5307613873?pr=2384

Sure, I would love to. I only have a Bangle1 here atm, that might change soon if Poolitzer is gifting me his old one.

d3nd3:
Okay , I have tested it and I haven't found anything non-functional about it. The behaviour I wanted is working, the key presses go through when screen is off. 

The backlight api seems to be working fine too, it didn't change much so that is what I expected, the function was already there.  Unable to test on a bangle 2, perhaps you could do that, all you would look for is if the backlight behaves correctly from the 3 functions: .on("LCDPowerBacklight" , isBacklightOn() , setLCDPowerBacklight().  Expect to see the sun-read display separate from backlight.  But ye its a small change to code, so everything _should_ be fine.

libs/banglejs/jswrap_bangle.c
```diff
@@ -1275,19 +1283,23 @@ void peripheralPollHandler() {
...
-      if (bangleFlags&JSBF_LOCKED) {
+      if (lockTimeout && bangleFlags&JSBF_LOCKED) {
         bangleTasks |= JSBT_UNLOCK;
         handled = true;
       }
     }
+    // ignore if locked
+    if (bangleFlags & JSBF_LOCKED) handled = false;
+    // extend timeouts when unlocked
+    inactivityTimer = 0;
```
gfwilliams, line 1302: I'm a bit concerned that in this code, any 'tap' interrupt from the accelerometer will keep the bangle awake, regardless of settings? So if you're just driving along on a bumpy road and the Bangle gets bounced around a bit, it won't turn itself off...

Maybe this should be in `bangleFlags&JSBF_WAKEON_TOUCH` ?

gfwilliams:
Thanks for the explanation in #2383 - the backlight stuff looks great, and it's really good to have better control of that.

I misunderstood earlier - I thought you wanted to get events through even when the device was locked - but what you want is to be able to be unlocked with the LCD/backlight off?

So my only concern is the line I linked above - I think for the Q3 wake on tap (which uses the accelerometer) we probably only want to reset the inactivityTimer if we've got wake_on_tap enabled - plus, should `if (bangleFlags & JSBF_LOCKED) handled = false;` be the `=true`? as far as I can see setting handled=false actually pushes the event rather than ignoring it?

d3nd3:
Yes, it should be = true, and it also should be within the WAKE_TOUCH, correct x2. I did a force push ammed.

d3nd3:
I just had a realization that the on Locked code should not be on the WAKEON_TWIST, WAKEON_FACEUP or WAKEON_TOUCH. Those are user settings which ignore LOCK.

gfwilliams:
> I just had a realization that the on Locked code should not be on the WAKEON_TWIST, WAKEON_FACEUP or WAKEON_TOUCH. Those are user settings which ignore LOCK.

I'm not entirely sure I understand?

d3nd3:
I mean, every input like on tap, on touch, on twist, on faceup should be filtered by is Locked in general, but not for the WAKE part. That was only in the BtnCommonFunction because that function handles 3 button and some might not have setting wake active in settings.

d3nd3:
I think thats ok, then the people who want ultimate control will disable auto-lock, then they will hook into every `.on` and control locking themselves within javascript. What do you think?

gfwilliams:
> should the users be denied .on events, when they have locked set.
> I think thats ok, then the people who want ultimate control will disable auto-lock, then they will hook into every .on and control locking themselves within javascript. What do you think?

Yes, I think for inputs that interact with a UI they should only be available when unlocked, **even when they are the WAKEUP flags**. If you press the middle button to wake the watch, you don't want it to jump straight to the launcher - it should unlock the watch and the next press goes to the launcher.

But I think not all should be behind locked:

* Not when locked: buttons/tap/drag/touch/stroke
* All the time: twist/faceup/health/HRM/accel/etc
* Maybe all the time? gesture

I'm not entirely sure if that's what's happening with your changes at the moment?

gfwilliams:
> We have the choice to put the !(bangleFlags & JSBF_LOCKED) check before jshHadEvent()

No, don't do that - if you're setting bangleTasks or similar then you should always call jshHadEvent. If you don't then the next time the Bangle does go to the event look for something it'll execute the event then

d3nd3:
I mean like : 
```javascript
if (!(bangleFlags & JSBF_LOCKED)) {
    faceUpSent = true;
    bangleTasks |= JSBT_FACE_UP;
    jshHadEvent();
}
```
Or let the task created and put the locked check in the other function where the task is handled.

Also I had idea for future possibility, if we ever really need more direct access to the inputs even when locked, could be a .on("touch-raw" etc... which doesn't check locked condition. Probably will never be requested though, and those die-hards could always fall back to pin reads.

d3nd3:
https://github.com/espruino/Espruino/blob/491288d9ffb88150e4213f94c4ed30c99380b90b/libs/banglejs/jswrap_bangle.c#LL1447C1-L1457C8

Is it intentionally that WAKE_ON_TWIST doesn't absorb the event like buttons do?

d3nd3:
The workflow action auto-builder not applying anymore?

d3nd3:
The tap location for top bottom and right left are reverse to the docs description. Wanna double check?  Its crazy how many different ways there are to input with the device. I never knew there were so many tap directions.

libs/banglejs/jswrap_bangle.c
```diff
@@ -1990,7 +1997,13 @@ void jswrap_banglejs_setLCDPowerBacklight(bool isOn) {
 #endif // !EMULATED
 }
 
-
+#if 0
```
gfwilliams, line 2000: Please can you not add `#if 0` code to PRs that go into Espruino?

gfwilliams:
> I mean like :

Ahh, ok - but as I said above the faceup event doesn't want a locked check on it.

> Is it intentionally that WAKE_ON_TWIST doesn't absorb the event like buttons do?

Yes - you wouldn't normally use the twist event in the UI - you probably care more about receiving it all the time (maybe you want to show seconds on a watch face when twisted, even when the watch is locked).

> I think you mean ....

Yes, that sounds exactly right. So is that what

> The workflow action auto-builder not applying anymore? 

It's because you have conflicts with the master branch now (stuff changed in jswrap_bangle and the CHangelog) but I'll try and fix that

> The tap location for top bottom and right left are reverse to the docs description

I just checked and on Bangle.js 2 they are correct.

So please can you change https://github.com/espruino/Espruino/pull/2384/commits/4e1403f40e1b572ec770d5dd8afb57efef3cc4c6 to only change it for the Bangle.js 1, or you'll inadvertantly break the 2.

Also just to check - you're not running it in left-handed mode or anything? It looks like we don't take account of that currently.

Please can you try and keep the changes in each PR to a minimum? Just as I thought we were getting to a consensus on the event handling there's now this `tap` change which definitely breaks Bangle.js 2

libs/banglejs/jswrap_bangle.c
d3nd3: Ok.

d3nd3:
I am sorry about putting too much into one PR.  I dont know what left handed mode is, never knew it exist ,can't find reference to it, I move the left right back into the #ifdef BANGLEJS_Q3 area.

d3nd3:
inactivityTimer might be better inside that condition too.
```javascript
if (lastBangleTasks != bangleTasks) {
    jshHadEvent();
    // ensure we don't sleep if touchscreen is being used
    inactivityTimer = 0;
}
```

gfwilliams:
> inactivityTimer might be better inside that condition too.

Sounds like a great plan, yet

d3nd3:
What sort of inputs do you want to trigger the inactivityTimer to reset, or should it only be triggered for inputs that have WAKE_ON in settings ?

gfwilliams:
> What sort of inputs do you want to trigger the inactivityTimer to reset,

I think just buttons and touchscreen (not 'tap' as that uses accelerometer)

d3nd3:
Currently its reset like this :  can we remove exceptions?
Currently:
    TWIST - IF settings.WAKE_
    FACEUP - IF settings.WAKE_
    BUTTONS - ANY - settings.WAKE_ not required.
    Bangle1 - TOUCH - IF settings.WAKE_
    Bangle2 - ANY - settings.WAKE_ not required

gfwilliams:
> I think you want touch event and BTN event to reset inactivityTimer without requiring WAKEON_TOUCH or WAKEON_BUTX for both Bangle1 and Bangle2

Yes, that sounds like a good idea.

> But yet for TWIST/FACEUP, they require WAKEON_TWIST and WAKEON_FACEUP to reset the timer.

Yes, I think so - otherwise maybe you're doing some kind of exercise that moves your arms around a lot and your Bangle won't ever sleep.

d3nd3:
It should be described like :

FaceUp/Twist have their inactivityTimer measured from the last time the device `woke up`.

Touch/Buttons have their inactivityTimer measured from the time of last input.

So FaceUp/Twist have their inativityTimer reset in code like :
```javascript
bool wokeup = false;
if (lcdPowerTimeout && !(bangleFlags&JSBF_LCD_ON)) {
    jswrap_banglejs_setLCDPower(1);
    wokeup = true;
}
if (backlightTimeout && !(bangleFlags&JSBF_LCD_BL_ON)) {
    jswrap_banglejs_setLCDPowerBacklight(1);
    wokeup = true;
}
if (lockTimeout && (bangleFlags&JSBF_LOCKED)) {
    jswrap_banglejs_setLocked(false);
    wokeup = true;
}
if (wokeup) inactivityTimer = 0;
```

Touch/Buttons have their inactivityTImer reset like this:
```javascript
inactivityTimer = 0;
```

gfwilliams:
Yes, that sounds good to me

d3nd3:
> What sort of inputs do you want to trigger the inactivityTimer to reset,
> 
> I think just buttons and touchscreen (not 'tap' as that uses accelerometer)

The question  is equivalent to asking `What sort of inputs are allowed to extend the wake state`.
Isn't a tap a legitimate input for ui? The `.on.touch()` for bangle2 even uses it, so its considered touch for bangle2.  I think that tap should `extend the wake state`.  Is the justfication for not using it that it might never sleep if moving around?

gfwilliams:
Are we getting confused between Tap and Touch?

Tap comes from the accelerometer, so conceivably if you were in a very bouncy environment it'd get triggered. It's never used for UI input

d3nd3:
> Are we getting confused between Tap and Touch?
> 
> Tap comes from the accelerometer, so conceivably if you were in a very bouncy environment it'd get triggered. It's never used for UI input

Ah what I was referring to is :
```javascript
#ifdef BANGLEJS_Q3
    if ((bangleFlags&JSBF_WAKEON_TOUCH) && (tapInfo&2)) {
```
Which is just the WAKEON_TOUCH setting being associated with a tap event, where as on bangle1 it doesn't do that.

d3nd3:
I've created a table to make things a bit more clear.

| Event                | Ignores Lock? | Can Wake? | Extends Wake in Unlocked?  | Consumes Wake Event? |
|----------------------|---------------|-----------|----------------------------|----------------------|
| Drag                 | No            | No        | Yes                        | No                   |
| Swipe                | No            | No        | Yes                        | No                   |
| Stroke               | No            | No        | Yes                        | No                   |
| Tap                  | No            | No        | Yes                        | No                   |
| Twist                | Yes           | Yes       | No                         | No                   |
| Faceup               | Yes           | Yes       | No                         | No                   |
| Button               | No            | Yes       | Yes                        | Yes                  |
| Touch                | No            | Yes       | Yes                        | Yes                  |

gfwilliams:
Ahh - yes. Since we can't leave the touchscreen on due to power consumption on Bangle.js 2 we fake it with the accelerometer, so we have to treat that event differently on Bangle.js 1/2

On your table:

* 'Tap' should ignore Lock
* swipe/stroke are not super important for extending wake, since they will all generate a Touch/Drag at some point as part of them - but I guess adding the code doesn't hurt

But yes, apart from that it looks good

d3nd3:
https://github.com/espruino/Espruino/blob/8f6bb0fdfc915454ee4b3734db8a0dacfb236dca/libs/banglejs/jswrap_bangle.c#LL1646C1-L1650C4

Can you remember why this returns false here, instead of true. Why doesn't it consume the event if locked for BTN4/5 which are pseudo touch buttons.  I think i might have guessed at one point, but i forgot.

gfwilliams:
I can't remember - but why don't you leave it as-is if it works?

If you change it you might save a microsecond when someone presses something, but there's a good chance it might break something.

So are you confident that the code now does what's in that table?

d3nd3:
> I can't remember - but why don't you leave it as-is if it works?
> 
> If you change it you might save a microsecond when someone presses something, but there's a good chance it might break something.
> 
> So are you confident that the code now does what's in that table?

Not just yet. I am preparing it, give me a day.

d3nd3:
I am going to test it my device.  Take care when merging it because I have duplicated:
`JsBangleTasks lastBangleTasks = bangleTasks;`
```javascript
if (lastBangleTasks != bangleTasks) {
    jshHadEvent();
    inactivityTimer = 0; // extend wake - only unlocked reaches here
  }
```
from your other commit.  Because I've added inactivityTimer into there. And wanted it to test.

d3nd3:
I am confident that the code does what is in that table.  Its ready to merge.  If there is a 1% error, it should be spotted quickly and an extra commit can always fix it. Difficulty for programmers to be 100% certain ;)

libs/banglejs/jswrap_bangle.c
```diff
@@ -1412,6 +1423,24 @@ void peripheralPollHandler() {
     }
     if (faceUpTimer<TIMER_MAX) faceUpTimer += pollInterval;
     if (faceUpTimer>=300 && !faceUpSent) {
+      // bypass lock. wake does not consume. do not extend wake
+      if (faceUp && (bangleFlags&JSBF_WAKEON_FACEUP)) {
+        bool woke = false;
+        // LCD was turned off, turn it back on
+        if (lcdPowerTimeout && !(bangleFlags&JSBF_LCD_ON)) {
+          woke = true;
+          jswrap_banglejs_setLCDPower(1);
```
gfwilliams, line 1432: Is there a reason you don't just do `bangleTasks |= JSBT_LCD_ON;` here like for other events? Or is it just that you moved the code and didn't change it?

The problem is you moved this code from outside of an interrupt to inside an interrupt, and it's probably not safe to call these functions from an IRQ since they could be being called from outside an IRQ at the same time.

gfwilliams:
This looks good apart from that code I just mentioned in the wakeup handler... If there's no reason for calling those functions direct I'm happy to swap it to `bangleTasks |= JSBT_LCD_ON`/etc here.

libs/banglejs/jswrap_bangle.c
d3nd3: Well spotted.  I am going to force push with those changes, as to not spam commits.

d3nd3:
Yes, I moved the code and didn't realise the implication there. Should be `bangleTasks |= JSBT_LCD_ON` etc.

PR #2393, opened 2023-07-17
Guard against Bangle.setUI({ <no mode> })

bobrippling:
Following on from #2389, this makes it clear that while `setUI()` can be used to reset the UI, `setUI({})` cannot, and must always specify `mode` (amongst other options).


gfwilliams:
Great, thanks!

PR #2394, opened 2023-07-21
Add non-new constructors for Number, String, Date, Boolean & RegExp

bobrippling:
This allows typescript code such as:

```typescript
var n = Number("123.45");
```

and similarly for `String(...)`, `Date(...)`, etc


gfwilliams:
Thanks!

PR #2397, opened 2023-08-03
Add `setLCDOverlay()` clearing typescript overload

bobrippling:
To check - is this how we're meant to clear the LCD Overlay?


gfwilliams:
Yes, that's good - thanks!

PR #2400, opened 2023-08-14
ble: add method to resolve random addresses

ssievert42:
This adds a method `NRF.resolveAddress()`, that allows resolving a "random private resolvable address" of a bonded peer to that peer's actual address.

Sadly the peer manager function that is needed for this, `pm_address_resolve()`, only exists in nrf sdk v14.1 and above, and thus the only device that currently supports this is Bangle.js 2.

My usecase is whitelisting my smartphone on Bangle.js 2, which works nicely after modifying both the settings and bootloader apps to use this method.


gfwilliams:
It's odd that `pm_address_resolve` just calls `im_address_resolve`, and *that* exists on SDK12, so maybe we could use that?

This looks really good though - it'd make whitelist a lot better

ssievert42:
Using `im_address_resolve` seems to work, at least on Bangle.js 2 :+1:

gfwilliams:
Looks great, thanks! Works on SDK12 too. Just added a useful code example.

If you're able to contribute the Bangle.js whitelist code you have (if it can be backwards compatible to builds without this function) that'd be awesome.

However I realise that the Nordic SDK does have some whitelist code internally that we're not using, so I guess potentially it makes more sense to enable the use of that from Bangle.js rather than the slightly hacky solution at the moment?

ssievert42:
> If you're able to contribute the Bangle.js whitelist code you have (if it can be backwards compatible to builds without this function) that'd be awesome.

Alright, I'll do that in a moment :+1:

> However I realise that the Nordic SDK does have some whitelist code internally that we're not using, so I guess potentially it makes more sense to enable the use of that from Bangle.js rather than the slightly hacky solution at the moment?

Yup, I think so too! Although that would (if I'm not looking at the wrong code) limit us to 8 (see `BLE_GAP_WHITELIST_ADDR_MAX_COUNT`) peers and, most importantly, the hacky js only solution seems to work correctly. So :shrug:

PR #2401, opened 2023-08-14
ble: allow configuring device privacy

ssievert42:
This allows using a random private resolvable address to prevent being tracked.

There are now two new methods: `NRF.setPrivacy()` and `NRF.getPrivacy()`, to set and get the current privacy settings.

One thing that (currently) cannot be configured is our IRK. Should be possible to add this as well, but I'd rather not do that, because that would complicate things a bit.

I always get a bit nervous when working directly with pointers, hopefully I'm not leaking any memory and / or freeing memory too early.

The initial setup with Gadgetbridge can be kind of weird and create a bunch of devices that don't actually exist, but everything works without (noticable) issues after that.

Something that bugs me is that if we run this on startup in addition to `NRF.setSecurity()` we may trigger a softdevice restart twice.

ToDo:

- [x] add nicer docs to explain what exactly `NRF.setPrivacy()` expects.
- [ ] try to figure out why Gadgetbridge may add devices that don't exist on initial setup

### DIFF
--- a/libs/bluetooth/bluetooth_utils.c
+++ b/libs/bluetooth/bluetooth_utils.c
@@ -210,6 +210,131 @@ const char *bleVarToUUIDAndUnLock(ble_uuid_t *uuid, JsVar *v) {
   return r;
 }
 
+#if PEER_MANAGER_ENABLED
+bool bleVarToPrivacy(JsVar *options, pm_privacy_params_t *privacy) {
+  if (jsvIsObject(options)) {
+    bool invalidOption = false;
+    memset(privacy, 0, sizeof(pm_privacy_params_t));
+    privacy->privacy_mode = BLE_GAP_PRIVACY_MODE_OFF;
+    privacy->private_addr_type = BLE_GAP_ADDR_TYPE_RANDOM_PRIVATE_RESOLVABLE;
+    privacy->private_addr_cycle_s = 0; // use default address change cycle
+    privacy->p_device_irk = NULL; // use device default irk
+    // privacy mode
+    {
+      JsVar *privacyModeVar = jsvObjectGetChildIfExists(options, "privacy_mode");
+      if (privacyModeVar && jsvIsString(privacyModeVar)) {
+        if (jsvIsStringEqual(privacyModeVar, "off")) {
+          privacy->privacy_mode = BLE_GAP_PRIVACY_MODE_OFF;
+        } else if (jsvIsStringEqual(privacyModeVar, "device_privacy")) {
+          privacy->privacy_mode = BLE_GAP_PRIVACY_MODE_DEVICE_PRIVACY;
+        } else {
+          invalidOption = true;
+        }
+      } else {
+        invalidOption = true;
+      }
+      jsvUnLock(privacyModeVar);
+    }
+    // other options are only relevant if privacy_mode is something other than off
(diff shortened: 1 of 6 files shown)


gfwilliams:
I didn't even realise this was a thing - seems like a good idea. But there are definitely memory leaks from handling JsVars.

> if we run this on startup in addition to NRF.setSecurity() we may trigger a softdevice restart twice.

It's probably not a huge deal, but what if we just add these as options to setSecurity and getSecurityStatus? It feels like it'd be cleaner than having another set of API functions, and it avoids the restart issue

libs/bluetooth/bluetooth_utils.c
```diff
@@ -210,6 +210,105 @@ const char *bleVarToUUIDAndUnLock(ble_uuid_t *uuid, JsVar *v) {
...
+    return result;
+  }
+  jsvUnLock(result);
+  return 0;
+}
+
+bool blePrivacyVarEqual(JsVar *a, JsVar *b) {
+  return (jsvIsUndefined(a) && jsvIsUndefined(b)) ||
+         (jsvIsObject(a) && jsvIsObject(b) &&
+          jsvIsEqual(jsvObjectGetChildIfExists(a, "privacy_mode"),
```
gfwilliams, line 303: These will all leak as you're not unlocking the result of jsvObjectGetChildIfExists

ssievert42: Thanks for pointing that out!  
Should be fixed now.

targets/nrf5x/bluetooth.c
```diff
@@ -2731,18 +2731,43 @@ void jsble_advertising_stop() {
...
+#endif
+
 #ifdef NRF52_SERIES
    // Set MAC address
    JsVar *v = jsvObjectGetChildIfExists(execInfo.hiddenRoot, BLE_NAME_MAC_ADDRESS);
    if (v) {
      ble_gap_addr_t p_addr;
      if (bleVarToAddr(v, &p_addr)) {
+#if PEER_MANAGER_ENABLED
+       err_code = pm_id_addr_set(&p_addr);
```
ssievert42, line 2761: This change is here because [the docs](https://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.sdk5.v15.0.0%2Fgroup__peer__manager.html&anchor=ga780f7978bfd31ded77908e2b33e8d05e) state that, if the peer manager is used, `sd_ble_gap_addr_set` must not be called and `pm_id_addr_set` should be used instead.  
I don't know what might happen if we ignore that, but "must not be called" sounds pretty serious to me.  
Ideally this shouldn't break anything.

bobrippling:
I've been tinkering with BLE addresses recently and would be interested in this - is there much left to bring it to being mergeable?

ssievert42:
Superseded by #2506

ssievert42:
This can be closed now.

bobrippling:
I see, fair points about the name being shown and the brief window before the softdevice can be configured - thanks for the details. That lines up / helps answer some thoughts/questions with what I've been tinkering with.

I guess I'd not be able to use the random private addresses with OpenHaystack too, since it pretty much fixes the address as a public one, but I've not looked into OpenHaystack to see if we can advertise a public key as part of a private random address - an investigation in progress!

> There's also at least one philosophical question, which is which address `NRF.getAddress()` should return.

Certainly! I found this and ended up with #2503

PR #2403, opened 2023-08-15
ble: use peer manager function for iterating over known peers

ssievert42:
This swaps `pdb_next_peer_id_get` for `pm_next_peer_id_get`. They both do the same thing, but using `pm_next_peer_id_get` allows the build to work without other changes on SDK 15_3.

I should probably have already used the peer manager function in the initial PR...


gfwilliams:
Thanks!

PR #2407, opened 2023-08-25
allow NDEF records with length over 254 bytes

### LINKED ISSUE #2406: NRF.nfcRaw (and nfcUrl) limited to payload/message length 255
see also conversation https://forum.espruino.com/conversations/354213/#17099032

https://github.com/espruino/Espruino/blob/master/libs/bluetooth/jswrap_bluetooth.c#L2638
`flatStrPtr` is `uint8_t *` so the dataLen is truncated even if data buffer is allocated with full length here
https://github.com/espruino/Espruino/blob/master/libs/bluetooth/jswrap_bluetooth.c#L2626
and terminated  here
https://github.com/espruino/Espruino/blob/master/libs/bluetooth/jswrap_bluetooth.c#L2641
same bug/feature  is there for nfcUrl - two lines here
https://github.com/espruino/Espruino/blob/master/libs/bluetooth/jswrap_bluetooth.c#L2456C1-L2456C1

fanoush:
- use 3 byte length in NDEF TLV block when over 254 bytes
- store 16bit url payload length
- -add maximum lenght checking (type 2 tag data is up to 992 bytes)

this is first attempt to fix #2406 , works for me

### DIFF
--- a/libs/bluetooth/bluetooth.h
+++ b/libs/bluetooth/bluetooth.h
@@ -316,27 +316,39 @@ void jsble_setup_advdata(ble_advdata_t *advdata);
 
 #define TAG_HEADER_LEN            0x0A
 
-#define NDEF_HEADER "\x00\x00\x00\x00" /* |      UID/BCC      | TT = Tag Type            */ \
-                    "\x00\x00\x00\x00" /* |      UID/BCC      | ML = NDEF Message Length */ \
-                    "\x00\x00\xFF\xFF" /* | UID/BCC |   LOCK  | TF = TNF and Flags       */ \
-                    "\xE1\x11\x7C\x0F" /* |  Cap. Container   | TL = Type Legnth         */ \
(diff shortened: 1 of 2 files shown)


fanoush:
not sure about NDEF and type 2 tag terminology in the comments, feel free to fix/suggest
also the `NRF.nfcUrl` has some duplicated code with `NRF.nfcRaw`, it could be done like others - prepare just the payload and pass that to ` jswrap_nfc_raw`

fanoush:
After splitting jswrap_nfc_raw into two parts the second part can be more easily reused from all other methods without creating extra copy of the payload.

Maybe I should move the `nfc_raw_data_start` method elsewhere now? at least above the first use so the extra definition is not needed there?

gfwilliams:
This looks great, thanks! Really good news on the space saving too!

> Maybe I should move the nfc_raw_data_start method elsewhere now? at least above the first use

That sounds like a good plan - maybe just to where you put the forward decl for now?

gfwilliams:
That's excellent, thanks! Merging now :)

PR #2415, opened 2023-09-15
Fix BangleJS debug builds failing on bogus assert call

elima:
An assert with a bogus/outdated condition is failing debug builds for BangleJS. This patch fixes the condition by checking against `IOEvent->flags` field instead.


gfwilliams:
Thanks! Wow, not sure how that got in

PR #2418, opened 2023-09-22
ESP32: Regression fix, cannot build without Bluetooth.

GaryOtt:
Added `#ifdef BLUETOOTH` to overcome compilation error.

```
targets/esp32/main.c:86:5: error: implicit declaration of function 'gatts_sendNUSNotificationIfNotEmpty' [-Werror=implicit-function-declaration]
     gatts_sendNUSNotificationIfNotEmpty();
```


gfwilliams:
Thanks!

PR #2421, opened 2023-10-28
fix CENTRAL_LINK_COUNT default for SDK15

### LINKED ISSUE #2420: 52840/sdk15 boards - no output over BLE without  having -DCENTRAL_LINK_COUNT=1 in board file
This does not work for some reason
https://github.com/espruino/Espruino/blob/62acf92dd9cd718bfb4c1c013f67b0c80fc1a489/targets/nrf5x/app_config.h#L181C3-L181C3

with this patch
``` diff
diff --git a/boards/BANGLEJS2.py b/boards/BANGLEJS2.py
index 91641c0cf..1494e0797 100644
--- a/boards/BANGLEJS2.py
+++ b/boards/BANGLEJS2.py
@@ -47,7 +47,7 @@ info = {
      #'DEFINES += -DESPR_REGOUT0_1_8V=1', # this increases power draw, so probably not correct!
      'DEFINES += -DESPR_LSE_ENABLE', # Ensure low speed external osc enabled
      'DEFINES += -DNRF_SDH_BLE_GATT_MAX_MTU_SIZE=131', # 23+x*27 rule as per https://devzone.nordicsemi.com/f/nordic-q-a/44825/ios-mtu-size-why-only-185-bytes
-     'DEFINES += -DCENTRAL_LINK_COUNT=2 -DNRF_SDH_BLE_CENTRAL_LINK_COUNT=2', # allow two outgoing connections at once
+#     'DEFINES += -DCENTRAL_LINK_COUNT=2 -DNRF_SDH_BLE_CENTRAL_LINK_COUNT=2', # allow two outgoing connections at once
      'LDFLAGS += -Xlinker --defsym=LD_APP_RAM_BASE=0x3660', # set RAM base to match MTU=131 + CENTRAL_LINK_COUNT=2
      'DEFINES += -DESPR_DCDC_ENABLE=1', # Use DC/DC converter
      'ESPR_BLUETOOTH_ANCS=1', # Enable ANCS (Apple notifications) support
diff --git a/targets/nrf5x/app_config.h b/targets/nrf5x/app_config.h
index 40163138a..74ab1e04f 100644
--- a/targets/nrf5x/app_config.h
+++ b/targets/nrf5x/app_config.h
@@ -182,6 +182,7 @@
 #define CENTRAL_LINK_COUNT 1
 #else
 #define CENTRAL_LINK_COUNT 0
+#error Central is 0
 #endif
 #endif
 // SDK15+ (fixes BLE UART send when CENTRAL_LINK_COUNT>1)
```
I get `Espruino/targets/nrf5x/app_config.h:185:2: error: #error Central is 0` when building for BANGLEJS2

Not sure why PEER_MANAGER_ENABLED is undefined there,  it is also tested in NRF5X_SDK_12 section above and that maybe works?

Noticed because I have older board files without the -DCENTRAL_LINK_COUNT lines and with fresh build I see no console output when connecting and also `process.env.APP_RAM_BASE.toString(16)` was lower than expected. Adding explicit
`     'DEFINES += -DCENTRAL_LINK_COUNT=1 -DNRF_SDH_BLE_CENTRAL_LINK_COUNT=1',` helped.

fanoush:
fix CENTRAL_LINK_COUNT default to be 1 except for nrf51

fixes #2420


gfwilliams:
Thanks!

PR #2422, opened 2023-10-30
getPressure: add Promise<> and undefined to types

bobrippling:
Some small typescript and doc tweaks to `getPressure`


gfwilliams:
I guess it raises the question of whether we should actually make getPressure *always* return a promise - but that doesn't help firmwares that are out there already

bobrippling:
Yeah, although I suppose if we always return a promise, then for hardware without a barometer, we'd (probably) choose to resolve with `undefined`. So I think how things currently are is better, since then a caller will find out immediately whether they'll get a pressure reading, rather than when the promise resolves.

gfwilliams:
Yes, I was thinking we might reject the promise - but as you say that would be harder to track down as it won't have a stack trace

PR #2436, opened 2023-11-27
Fix RJSON's object parsing

bobrippling:
Found in espruino/BangleApps#3114, object parsing exits early for strings with an empty object:

```javascript
>JSON.parse('{"c":1}')
={  }
```

Or if we've parsed in a relaxed mode and hit a string, I'd expect us to exit with that object on hitting a string key, but I'm not sure how to enable the relaxed mode parsing from an espruino prompt.


gfwilliams:
Thanks! Wow, sorry - that was a really stupid bug to introduce!

bobrippling:
Thanks for the follow up too! I need to get setup for running locally, was flying blind with that one

gfwilliams:
:) No problem! If you're on Windows it's worth installing WSL - once that's installed, you can git clone it and it should 'just work' with `make`

Issue #2137, opened 2022-01-30
`Bangle.getPressure()` fails sometimes with I2C Error on Bangle.js 2


malaire:
This code fails sometimes with I2C error on Bangle.js 2. The "ERROR"-message from code is not output.

Failures seem somewhat random, I've seen failures after step 1, 9, 19 and 20 so far and never got beyond 20 steps.

```
var counter = 0;

function foo() {
  Bangle.getPressure().then(d => {
    counter++;
    Terminal.println("" + counter + " " + d.temperature);
    setTimeout(function() { foo(); }, 10 * 1000);
  }).catch(error => {
    // This is never reached.
    Terminal.println("ERROR");
  });
}

foo();
```

Error is

```
Uncaught Error: I2C Error: Arbitration (start)
 at line 1 col 20
Bangle.getPressure().then(d=>{counter++;Terminal.println(""+co...
                     ^
in function "foo" called from line 1 col 5
foo();
    ^
in function called from system
```

gfwilliams:
What firmware version are you using? Just last week I made some changes to increase the time given for the sensor to power on, and it should fix issues like that

malaire:
I had 2v11. I've now updated firmware to latest available, "Firmware Update" -app says
> Your current firmware version is 2v11.67 and bootloader is 2v11.52

This seems much better, but I'm still getting that error rarely, so this has not been fully fixed.

malaire:
This failure is much more rare with 2v11.67 but I've already got it twice with 2v11.67 by re-sending the above code to Bangle.js 2 after it reaches about 20 steps.

gfwilliams:
Thanks - the problem is it's hardware-dependent, so on all the devices I have here it works fine.

Since the last PR made it better by increasing the delay, I guess I'll just increase it some more and hopefully that'll fix it.

gfwilliams:
I pushed an update yesterday which bumps the power up delay from 750ms to 1000ms - hopefully that'll fix it for you now?

malaire:
Still not fixed in 2v11.71 - I needed just 7 minutes of testing with the above code to get first error.

Also I don't understand how increasing power up delay could help here as I never get error on first `Bangle.getPressure()`. First call always succeeds and then later one of the subsequent calls fails.

gfwilliams:
> Also I don't understand how increasing power up delay could help here

Do you have other suggestions for how to fix this then?

[...]

PR #2438, opened 2023-12-06
getPressure: handle error when we don't have an exception

bobrippling:
This handles an I2C error, originally ocurring in #2137, found again by chance in an [unrelated PR](https://github.com/espruino/BangleApps/pull/3104).

When we create a promise for `getPressure`, we now look at both `jspGetException` and then `jspHasError` (separately).
If the former is null but we have an error, we raise an exception about the I2C communication with the barometer.

Previous behaviour would raise an exception with no payload, leading to the `Uncaught Error: Unhandled promise rejection: undefined` seen in the BangleApps pull request above.

### DIFF
--- a/libs/banglejs/jswrap_bangle.c
+++ b/libs/banglejs/jswrap_bangle.c
@@ -31,6 +31,7 @@
 #include "jswrap_heatshrink.h"
 #include "jswrap_espruino.h"
 #include "jswrap_terminal.h"
+#include "jswrap_error.h"
 #include "jsflash.h"
 #include "graphics.h"
 #include "bitmap_font_6x8.h"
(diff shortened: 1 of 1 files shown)


libs/banglejs/jswrap_bangle.c
```diff
@@ -5073,27 +5073,38 @@ JsVar *jswrap_banglejs_getPressure() {
...
+    JsVar *r = promisePressure;
+    promisePressure = 0;
+    return r;
+  }
+
   bool hadError = jspHasError();
   if (hadError) {
-    JsVar *exception = jspGetException();
+    JsVar *msg = jsvNewFromString("I2C barometer error");
+    JsVar *exception = jswrap_internalerror_constructor(msg);
```
bobrippling, line 5088: I believe I have the locking correct here - `jswrap_internalerror_constructor` unlocks `msg`, but would appreciate a second pair of eyes

bobrippling, line 5088: > ```
> libs/banglejs/jswrap_bangle.c:5088:24: error: implicit declaration of function 'jswrap_internalerror_constructor'; did you mean 'jswrap_typedarray_constructor'? [-Werror=implicit-function-declaration]
> ```

I could include the header, but want to confirm - is this expected or should I be going through another function for this?

gfwilliams:
Looks like there's a build error here as well? https://github.com/espruino/Espruino/actions/runs/7112131428/job/19361472245?pr=2438

libs/banglejs/jswrap_bangle.c
gfwilliams: I don't think `jswrap_internalerror_constructor` does unlock msg - if it's a function called from JS (which I'm pretty sure it is as it's the constructor) then the caller is responsible for unlocking

gfwilliams: I'd add that to the headers - I think it's ok. Usually we'd create an exception with `jsExceptionHere` but that actually fires off the exception, which we don't want to do, so I think your method is fine

bobrippling: Agreed - I've included the header, all sorted

bobrippling: I might've misread the code - I think [here it unlocks](https://github.com/espruino/Espruino/blob/deedeafd91683359807fe77aa2bfb2b37d56d219/src/jswrap_error.c#L57-L57) in `jsvObjectSetChildAndUnLock`

https://github.com/espruino/Espruino/blob/deedeafd91683359807fe77aa2bfb2b37d56d219/src/jswrap_error.c#L51-L62

(⬆️  `msg` here)

https://github.com/espruino/Espruino/blob/deedeafd91683359807fe77aa2bfb2b37d56d219/src/jsvar.c#L3195-L3198

(⬆️  `child` here)

gfwilliams: Ahh - `jsvObjectSetChildAndUnLock(d, "message", msg); ` does unlock `msg` but on the line above, `msg = jsvAsString(msg);` replaces `msg` with a newly locked version of `msg` that's guaranteed to be a string. I guess for readability it should really be written to a new variable instead.

bobrippling: That makes sense - I could make that change as part of this PR if you like?

gfwilliams: No, it's ok - i'll do it separately now...

gfwilliams: but you still need to unlock `msg`...

bobrippling: Ah yes - I was assuming `jsvAsString` unlocked the value it was given, with us replacing it. Thanks for the pointer!

bobrippling:
This is ready for another look btw :)

PR #2441, opened 2023-12-13
Python3 build docs

brendena:
Made the build_doc script work with python3.  Mostly just migrating the libraries to newer version. 

Library change  
I needed to switch out mdx_urlize for markdown2 because this extension doesn't support the latest markdown python library.  It seems that it will never.  Somebody did a pull request in 2022 with the fix for it and it still hasn't been accepted.  https://github.com/r0wb0t/markdown-urlize/pull/17

Affected libraries.  EspruinoDocs will need to be told to use python not python2.7
https://github.com/espruino/EspruinoDocs/blob/master/build.sh#L31

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -785,7 +785,7 @@ endif
 
 docs:
 	@echo ================================== Generating Board docs
-	$(Q)python2.7 scripts/build_docs.py $(WRAPPERSOURCES) $(DEFINES) -B$(BOARD)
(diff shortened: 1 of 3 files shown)


gfwilliams:
Hi - I'm not sure if it was a typo but you'd removed `importlib` from `common.py` and while docs would have built, that stopped any of the code building. I've just re-added it and everything seems to work again but let me know if there was some reason

PR #2447, opened 2024-01-13
Bangle.getPressure() redesign

### LINKED ISSUE #2445: Bangle.getPressure() fails intermittently
**EDIT**: This code was initially shared without the `(function())()` wrapping, so the promise callbacks were acting weird due to line by line upload of the webIDE. I think the test has to be re-ran to **validate** how much errors we got. (Potentially None.) Latest undefined returns from 2v20 = 3/100 attempts.

My firmware = 2v20
``` javascript 
let sleepProm = function(delay) {
  return new Promise((resolve) => {
    setTimeout(()=> {
      resolve();
    },delay);
  });
};

(function() {

let pressurePromise = Bangle.getPressure();
for (let i = 0; i < 60; i++) {
  pressurePromise = pressurePromise.then((data)=>{
    if ( typeof(data) === "undefined" )
      print("undefined");
    else console.log(data);
    return sleepProm(500).then(()=>{
      return Bangle.getPressure();
    });
  });
}
})();
  ```
It doesn't error nearly as much, if correctly ran from file or wrapped in (function(){ })() for RAM.

d3nd3:
Fixes #2445 

I decided to keep the timeout as the comments suggest there need to be a minimum delay after turning on.  But still utilized the code in the periPollHandler to resolve the promise.

### DIFF
--- a/libs/banglejs/jswrap_bangle.c
+++ b/libs/banglejs/jswrap_bangle.c
@@ -4243,7 +4243,7 @@ bool jswrap_banglejs_idle() {
     JsVar *o = jswrap_banglejs_getBarometerObject();
     if (o) {
       jsiQueueObjectCallbacks(bangle, JS_EVENT_PREFIX"pressure", &o, 1);  
-      if ( getPressureReady ) {
+      if (getPressureReady) {
         getPressureReady = false;
         // disable sensor now we have a result
         JsVar *id = jsvNewFromString("getPressure");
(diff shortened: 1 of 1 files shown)


d3nd3:
You have a choice whether a `Bangle.on("pressure")` callback fires from a `Bangle.getPressure()`, currently it does.

d3nd3:
Okay. I am fairly happy with the code.

- When Bangle.setBarometerPower(1) prior to using Bangle.getPressure(), it has fast response as described in the docs, albeit the value isn't 'fresh' necessarily, as expected.
- If user uses streaming Bangle.on("pressure'), he gets 'fresh' values, they come as fast as they can whilst being fresh.
- Decide if you like it that Bangle.on("pressure") triggers during the temporary activation during Bangle.getPressure().

Feedback welcome.

gfwilliams:
Thanks - I'll post some comments in the issue...

PR #2449, opened 2024-01-15
trim DFU bootloader, enable BANGLE1 SPI flashing

fanoush:
disable  noncritical errata at startup (will get applied when main application starts)
disable unused protobuf decoders
disable gcc startup code
enable SPI flashing for BANGLEJS 1 (may not work for now, fits with 8 bytes free)


fanoush:
I did test this with other boards that it does not break DFU itself and bootloader still works as before.
I did **not** test with Bangle 1 at all as I don't have Bangle 1 board with SWD available.

gfwilliams:
This looks great - thanks!

I'll wait until I have some time to test before merging though - given it's such a pain to reflash Bangle.js 1 I really want to make sure this doesn't break the bootloader!

While this is really neat, I'm honestly not sure if adding it to the Bangle.js 1 is something I want to do... I'm not aware of anyone that asked for this feature, and realistically on Bangle.js 1 for most people the result of a failed firmware update is going to be a bricked watch...

It just feels like the risk of this being a support nightmare might be higher than any potential benefit to Bangle.js 1 users.

Obviously if you have devices you're using this with then I'm all for keeping your changes in the repo - I'm just not sure if we should enable it for Bangle.js 1 by default.

gfwilliams:
Thanks! I'll just add a change to comment out the lines in the python file, with a comment about what they do if you put them in - the rest is fine, and I'll merge :)

Issue #2450, opened 2024-01-15
Promises: resolve a new Promise


d3nd3:
``` javascript
(function() {
  let p = new Promise((r)=>{
    console.log("resolved");
    r(new Promise((r2)=>{
      setTimeout(()=>{
        console.log("settled")
        r2();
      },2000);
    }));
  });
  
  p.then((v)=>{
    console.log("settledr1");
  });
})();
```
`settledr1` is printed instantly
### expected:
outer promise settles/fulfills upon inner promise being resolved.

---
``` javascript
(function() {
  let p = new Promise((r)=>{
    console.log("resolved");
    r(new Promise((r2)=>{
      setTimeout(()=>{
        console.log("settled")
        //r2();
      },2000);
    }));
  });
  
  p.then((v)=>{
    console.log("settledr1");
  });
})();
```
`settledr1` is printed instantly
### expected:
settledr1 should not be printed because r2 *inner promise* never resolves.

gfwilliams:
Is that new behaviour? I always assumed that when you resolved a promise, you resolved with the value you wanted to return.

So if you resolve but pass another promise as the value that you're resolving, that then gets added to the chain?

gfwilliams:
I can't face trying to fix this at the moment, so I've just made it throw an exception if you try and resolve with a promise so at least you know it's not doing what you expect.

d3nd3:
I am not sure if its new or not.  But its based on the distinction between `resolve` and `fulfilled`.   Would you like me to try and fix it? 

I don't mind implementing it.

gfwilliams:
Thanks! Yes, it'd be great if you could fix it - but ideally the less code you end up changing in the process the better. Hopefully it can be done by just linking the new promise in in the right place

d3nd3:
Hey, I am about to test some code regarding this, but I want to test on the emulator instead of flash to my watch so often. Do you some tips how I can update the emulator to use the firmware I push via github?

gfwilliams:
> Ah it seems I can just run the linux version?

Yes, totally! That's the best way to do it. It has some promise tests built in already - and will be able to check for memory leaks too

d3nd3:
I will intend to fix this bug too, its related to .then() not always returning a new promise, when it needs to because they can spawn independent chains.
Do you prefer to do it in separate pull requests or all here?
``` javascript
(function() {
  p1 = Promise.resolve();
  let p2 = p1.then(()=>{
    return new Promise((y,n)=>{
[...]

Issue #2227, opened 2022-06-23
Promise implementation isn't correct


mariusGundersen:
While trying to implement `Promise.finally` I have tried to understand the implementation of promises in espruino, and it does not appear to work correctly. The chaining of promises can't work right, since promises form a DAG, not a chain. I think the promise implementation could be simplified and made to work correctly if the `chain` code isn't used.

Here is an example that fails in today:

```js
function delay(ms){
  return new Promise(r => setTimeout(r, ms));
}

var p = delay(1000);

p.then(() => 'a').then(x => console.log(x)); //logs a after a second
p.then(() => 'b').then(x => console.log(x)); //should log b after a second, but logs a instead
```

This is probably because [jswrap_promise_then](https://github.com/espruino/Espruino/blob/master/src/jswrap_promise.c#L418) calls on [jswrap_promise_get_chained_promise](https://github.com/espruino/Espruino/blob/master/src/jswrap_promise.c#L392) which checks if the promise has been chained already and uses the chained promise if it has. Therefore the two `then`s return the same promise, which they obviously shouldn't:

```js
var p = new Promise();

var a = p.then(() => 'a');
var b = p.then(() => 'b');

console.log(a === b); // logs true
```

They should both return a new unresolved promise

mariusGundersen:
This has been fixed in #2454

PR #2454, opened 2024-01-23
Increases promise alignment with spec

d3nd3:
#2450
#2227

This code aims to make promise behave closer to ES6 specs.

### DIFF
--- a/src/jswrap_promise.c
+++ b/src/jswrap_promise.c
@@ -12,7 +12,9 @@
  *
  * ES6 Promise implementation
  * ----------------------------------------------------------------------------
+ * See https://github.com/espruino/Espruino/pull/2454 for better understanding.
  */
+
 
 #include "jsutils.h"
 
@@ -27,8 +29,16 @@
 #define JS_PROMISE_CATCH_NAME JS_HIDDEN_CHAR_STR"cat"
 #define JS_PROMISE_REMAINING_NAME JS_HIDDEN_CHAR_STR"left"
 #define JS_PROMISE_RESULT_NAME JS_HIDDEN_CHAR_STR"res"
-#define JS_PROMISE_RESOLVED_NAME "resolved"
+#define JS_PROMISE_ISRESOLVED_NAME JS_HIDDEN_CHAR_STR"resolved"
+#define JS_PROMISE_PROM_NAME JS_HIDDEN_CHAR_STR"prom"
+#define JS_PROMISE_VALUE_NAME JS_HIDDEN_CHAR_STR"value"
+#define JS_PROMISE_STATE_NAME JS_HIDDEN_CHAR_STR"state"
 
+enum JS_PROMISE_STATE_ENUM {
+    JS_PROMISE_STATE_PENDING,
+    JS_PROMISE_STATE_REJECTED,
+    JS_PROMISE_STATE_FULFILLED
+};
 
 /*JSON{
   "type" : "class",
@@ -39,9 +49,8 @@
 This is the built-in class for ES6 Promises
 */
 
-void _jswrap_promise_queueresolve(JsVar *promise, JsVar *data);
-void _jswrap_promise_queuereject(JsVar *promise, JsVar *data);
-void _jswrap_promise_add(JsVar *parent, JsVar *callback, bool resolve);
+void _jswrap_promise_resolve(JsVar *prombox, JsVar *data);
+void _jswrap_promise_reject(JsVar *prombox, JsVar *data);
 
 bool _jswrap_promise_is_promise(JsVar *promise) {
   JsVar *constr = jspGetConstructor(promise);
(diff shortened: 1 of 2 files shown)


gfwilliams:
Thanks - sorry for the delay, I was away last week and I'm still catching up a bit - I wasn't sure if you thought this was quite ready for review.

At first glances it seems hard to narrow down exactly what's changed because it looks like everything has :) I need to take some time and build this and test it out.

Does this run through the existing local tests (`bin/espruino --test-all`) and pass them still? I know you did this to fix some issues with the existing implementation - is it possible to get them added to the tests that are there so we can make sure that they don't regress again in future if something changes?

gfwilliams:
Just looked into this and ran `--test-all` myself to check, but it looks like this breaks `tests/test_promise3.js` and `tests/test_promise4.js`, both of which work in Node.js. Was that expected?

And as above, it'd be really nice to have tests added for the issues that you fixed to ensure they don't end up broken in the future.

d3nd3:
When you ran them in node.js , what was the output?

For me, I get Unhandled promise rejection, which is correct and due to the line:
```javascript 
p.then(function(value) { // 1
}
);
```

calling p.then() again on the initial promise, sets up a new chain, and because the second chain doesn't have a catch handler, it has to error. Even though the first chain does have a catch handler. 

test_promise4.js has the same idea in it, D1 is correct, but since it errors, the 'result' variable is not set, so the test fails.

Those two tests should be fixed for that condition.

gfwilliams:
Thanks! I'd previously run them in Node by copy/pasting them into the REPL so I could check the value of `result`, and I'd obviously missed the exception that appeared at that point (and the end result was as expected if it kept executing after the exception). As you say running them as files in node breaks. I've just updated the tests and added 3 new ones for the outstanding issues with promises.

However now I see fails on:

* `test_promise4.js` - reports `r1Cundefined` and not `r1D1` (appears to be executing `then` even when it's been rejected)
* `test_promise15.js` - we get `Resolving a Promise with a value that is a Promise is not currently supported` - this is what happens in #2450 which I think you might have thought was fixed? It's possible this was because of a merge issue though as that change went in on 2024-01-15?
* `test_promise17.js` -  returns `a` twice when it should work as in #2227

So it seems this still might not fix the original reported 2 issues fully?

d3nd3:
Yes it must be because of the merge issue because I tested them here, and they pass.

gfwilliams:
Well I have no idea what happened but I've been unable to merge this without introducing errors interpreting promises - so I just took the file directly from this PR and overwrote what's there before. But this is now merged, and yes, the tests do all appear to pass.

Thanks for the PR - but it would have been nice if it included the tests that it was attempting to fix.

d3nd3:
Yes, sorry about that, I didn't have the time and when I did, I had forgot most of the details of the implementation.

I want to just say that the crux of the issue with the previous implementation was that, the 'chain' pointer (linked list) was used solely for resolve/reject pass-through when there is no 'callback' defined.  Yet as pointed out in #2227 , this will not work because one promise can have multiple 'children' (when calling .then on it).

So by instead using the return value of .then() for the 'resolve/reject passthrough when no cb defined (catch search)', it works correctly.

As for #2450 , its similar issue, the chain needs to to be an array chain, instead of linear chain. So I implemented it like : 

image

So now there can be multiple callbacks and chain for each Promise.  I hope that helps understand the changes.

gfwilliams:
Reopening because of https://github.com/espruino/Espruino/issues/2495 - I've reverted this for now as at least it's just a single file change, and I don't want to have literally every use of Promises in the Espruino APIs broken

d3nd3:
sad thing is, I can't remember why I am using a container object, I need time to remember.  Its possible we don't need it?

d3nd3:
I have remembered the reasoning for using the container object. Its because: 
https://github.com/d3nd3/Espruino/blob/8859819a0fe1305c8567e79ef2eb3cb82622ae5c/src/jswrap_promise.c#L187

When resolving a promise, it needs a unique instance of "resolved" , that is why resolved sits inside the prombox next to the promise.  So think of it as spawning a new instance of 'resolved', which is needed for the recursion, yet re-using the promise.

d3nd3:
```c
// otherwise the returned promise will be fulfilled with the value.
  JsVar *promBox = jspromise_create();
  if (!promBox) return 0;
  promise = jsvObjectGetChildIfExists(promBox, JS_PROMISE_PROM_NAME);
  if (promise) {
    jspromise_resolve(promBox, data);
  }
  jsvUnLock(promBox);
  return promise;
```
In the `jswrap_promise_resolve` function I had already fixed it, so I guess this is one solution.  But I don't really want to copy paste that everywhere, so having another function seems nicer.

mariusGundersen:
> So when javascript land calls resolve or reject via constructor executor, the prombox is passed via first 'this' argument, that is how the container doesn't need its reference stored.

Why can't you use the promise directly? Why is the wrapping box needed?

mariusGundersen:
Ok, it's to handle the scenario where you resolve a promise with another promise:

```js
new Promise((resolveOuter) => {
  resolveOuter(
    new Promise((resolveInner) => {
      setTimeout(resolveInner, 1000);
    }),
  );
});
```

I'm this case the outer promise will be resolved but not yet fulfilled (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#description).

gfwilliams:
If there's any way around it I'd really rather not have two separate types of promise object kicking around external to jswrap_promise.c (if it's used inside, fine) - it feels like it's going to cause all kinds of headaches in the future where in code someone creates the wrong promise object inside the Espruino codebase, or thinks that they can call promise_resolve on something passed from JS land, or pass the promise from promise_create into JS land. At the very least I guess we should rename `promise_X` to `promBox_X` in all the code and rename `blePromise`/etc to `blePromBox` (but that's quite a lot of changes).

I'm not quite sure I understood from the diagram you drew, but looking at the code the Box stores a single value for the resolved state, and a link to the Promise? But the link to the promise is never updated once set, so every box is linked to just one promise... and you have multiple boxes, one for each time a promise is resolved?

But it feels like the next item in the promise 'tree' can't even be executed until the first one has resolved, so presumably only one 'box' is active at a time... And when you do `.then` you get a new promise object created anyway, which would have its own box too?

Is there a reason we can't have the 'current' Box linked from the promise itself? So then we only pass the promise around, but when we need to set resolved/etc we can just follow the link to Box and set that?

I'm still not sure I get it - **Do you think you could give me an example of when it wouldn't work for some code in Espruino to use promise_create and promise_resolve on the actual promise rather than the box?**

I know there wasn't much documentation before (although there were a few in-code comments), but it would be really nice to have some kind of code documentation at the end of this, even if it's just what you said above at the top of the file explaining what's going on with promBox. Right now there are even less comments than there were before.

I've spent the last hour trying to dig through this and get my head around it, and I'm still not there - and when the original author says "I had forgot most of the details of the implementation" it's worries me about how maintainable this is going to be in the future.

d3nd3:
The code is heavily inspired from : https://github.com/humanwhocodes/pledge
and the tutorials he provides:
https://humanwhocodes.com/blog/2020/09/creating-javascript-promise-from-scratch-constructor/
https://humanwhocodes.com/blog/2020/09/creating-javascript-promise-from-scratch-resolving-to-a-promise/
https://humanwhocodes.com/blog/2020/10/creating-javascript-promise-from-scratch-then-catch-finally/
https://humanwhocodes.com/blog/2020/10/creating-javascript-promise-from-scratch-promise-resolve-reject/

also: 
https://www.ecma-international.org/ecma-262/11.0/index.html#sec-promise-objects

Its not like i am completely reinventing the wheel here.  But I could adjust it so that it doesn't use a box, it seems possible.

gfwilliams:
I'm just looking at some of the promise test cases and how they work, and I'm wondering whether everything is easier than I expected...

In the original code I was trying to have an array as a chain of promises, because I hadn't realised that `.then` should return a brand new promise each time - so everything ended up being a bit of a hack.

So...

* Promises would have a state, value (if resolved/rejected) and a list of 'children'
* `new Promise()` creates a promise,
* promise.then/catch *create a new promise* and add it to the 'children' list of the first promise (if the first promise is already resolved, it queues a resolve immediately)
* Now when something resolves/rejects, it just goes through all its children and queues for them to be executed.
* If something resolves returning a promise as data, we just move the children of the current promise onto that new promise and we're good (I think?)

Does that sound right, or am I missing something obvious? Is there a need for the boxes in here?

d3nd3:
For my code to not use box, I would make the resolved be children of the 'resolve' function object instead, as the spec does, I think that could work?

gfwilliams:
Ok, looks like I posted this just as you posted your response!

Thanks - those links look really helpful. I'm not entirely sure I understand what you mean about making them fields of the function objects, but if you think it could work that would be perfect. If it matches what happens in the spec more then that's great.

>  I was under the impression you preferred less comments

Oh no, I like comments! Not:

```
function getFoobar() // this gets the foobar
```

But actually explaining what's going on is very helpful - especially here where the jswrap_promise has been a bit of a minefield in the past :)

d3nd3:
So given a promise `p1`.

It can have many `reactions` attached to it via many separate `.then` calls. Chained or unchained.

When `p1` resolves, all `pX` callbacks/promises attached to it (lets call these reactions, which are a promise and a callback(can be rejectCB or fullfillCB)) are fired, the return value of the fired CB in turn will resolve `pX`.  `pX` is the promise that is returned from `p1.then()`. The CB function is the function passed to `.then`, : 
`p1.then(fullfillCB(){},rejectCB(){})`

d3nd3:
Okay, so I have remembered why I used a promBox instead of doing as the spec does by making the `prom and is_resolved` children of the resolve function.  The reasoning is because the resolve function doesn't always exist in javascript land as a function. It is sometimes a promise created in C land, where there is no corresonding JsVar resolve function, just the `_jswrap_promise_resolve` and `_jswrap_promise_reject` functions, but when the promise constructor is called in javascript, the functions are converted into JsVars:
```javascript
JsVar *jsResolve = jsvNewNativeFunction((void (*)(void))_jswrap_promise_resolve, JSWAT_VOID|JSWAT_THIS_ARG|(JSWAT_JSVAR<<JSWAT_BITS));
    JsVar *jsReject = jsvNewNativeFunction((void (*)(void))_jswrap_promise_reject, JSWAT_VOID|JSWAT_THIS_ARG|(JSWAT_JSVAR<<JSWAT_BITS));
```

So I didn't want to create two native functions just for the sake of the c land manual resolve cases, and I wouldn't know where to store them.  So because I lacked imagination on memory management in c land, that is when I came up with the idea of the prom box.

What are your thoughts now?

d3nd3:
https://esdiscuss.org/topic/es6-promises-alreadyresolved-needed

This seems interesting, so the part we don't fully understand is why it needs a separate isResolved, but mb this sheds some light? I think we don't support this 'locking in' feature anyway, so mb its not needed.

d3nd3:
image
 
This resolve call, when the inner promise resolves, has to succeed, even though the outer promise has its `is_resolved` field already set.

gfwilliams:
Wow, thanks for digging into this - that's great!

So in the 'outer promise' case, you can call `resolve` with a promise, but the outer promise will still show as pending until that inner promise resolves, even though you can't resolve it a second time because of alreadyResolved.

My initial thought is that having a 4th state on the promise `JS_PROMISE_STATE_PENDING_INNER_PROMISE` would fix this, but I think it'd then fail if a promise returned a promise which returned a promise?

So I guess we're stuck with the box, *or* what I'd mentioned:

> If something resolves returning a promise as data, we just move the children of the current promise onto that new promise and we're good (I think?)

But I feel like what you've done being in line with the spec makes a lot more sense.

So do you think it's possible to make promise_create just return a promise, and jspromise_resolve/reject to take a normal promise as before? It's possible they can just use a fake `prombox` with `alreadyResolved=false` as we know they'll only be called from Espruino where we won't double-resolve anyway.

All the rest of the your code can be basically the same

d3nd3:
> If something resolves returning a promise as data, we just move the children of the current promise onto that new promise and we're good (I think?)

I don't think this is that unique from what is already done.  You still would have to call resolve eventually on the outer promise, which already was resolved.

Resolved field is to prevent firing of resolve function twice ( prevent triggering execution of the children callbacks multiple times ).  Imagine in a constructor 
```javascript
new Promise((resolve,reject)=>{
  resolve();
  resolve();
  return;
})
```
State field is used primarily in the `jswrap_promise_then` function to check if the promise has already fullfilled or rejected, in which case the callback should instantly be queued because there is no other event that would do it, if it already occured.

d3nd3:
Minor changes, what do you think?

d3nd3:
> Looking at where jspromise_create_prombox is used I guess a minor optimisation might be to give it a JsVar **promise argument where it would return the new promise it created, so you don't have to immediately search in the returned object for it?

Thats a great idea!

d3nd3:
XD, pushing slightly too often, I need to be more patient. But ye compile errors fixed.

d3nd3:
Gotta fix a lock or two, then should be ready.

d3nd3:
tests/manual/bangle2_fill.js
tests/test_262_FAIL.js
tests/test_new_nested2_FAIL.js
test_number_constructor_FAIL.js

Only tests that fail, but I guess unrelated?

gfwilliams:
Looks great, thanks!

> Only tests that fail, but I guess unrelated?

Yes, that's fine - I tagged the ones we expect to fail with `_FAIL` and the ones in `manual` shouldn't really be executed

gfwilliams:
Argh, what a pain - it seems that this increases the code size by 800 bytes, so now the Puck and Pixl builds fail :(

I've had a bit of a look through and I've tried to reduce some duplication but I only managed to save 250 bytes, so now at least the Pixl build is working but the Puck one fails.

If we could save another 300 bytes it'd work, but I think I'll probably have to pull some features out to make this build again now

d3nd3:
we could remove the support for thenables and less strict error checking in certain places.

gfwilliams:
Thanks - yes, I was a bit surprised too! 800b is quite a lot all things considered.

I had a bit of a dive through and I found some other places I could save a bit of space, so everything is now compiling again.

Good point about the names though - I'll lower them - I believe if we can get under 4 chars in total then key+value get stored in a single JsVar usually so they same RAM too.

Removing thenables is a good idea - although I'm sure someone will come along and complain at some point, but I'll comment it out for now and see.

gfwilliams:
Actually we do have a test in there for thenables, so I guess we'll keep them for now.

With the other changes I made (reducing string sizes in the Puck.js lib and other stuff) we're back down to normal flash usage now

Issue #2455, opened 2024-01-26
using newer compiler for ARM (13.2.Rel1)


fanoush:
Creating this issue to track more issues and discuss rationale.

Current 8-2018-q4-major  8.2.1 compiler downloaded by provision script is quite old but produced shorter code than newer ones. However when testing latest one 13.2.rel1 gcc 13.2.1 from
https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads I see that it produces smaller code both for bootloader and main firmware. However there are some issues:

- There is a breakage with newer gcc compilers in bootloader build - it cannot handle  code in ".data" section, gives "Error: leb128 operand is an undefined symbol". https://www.mail-archive.com/gcc-bugs@gcc.gnu.org/msg718258.html I have a fix based on https://gcc.gnu.org/bugzilla/show_bug.cgi?id=103698#c4 that seems to work, so I can create a PR, it works with old compiler too

- When linking main firmware it produces set of (likely harmless) warnings like
```
arm-gnu-toolchain-13.2.Rel1-x86_64-arm-none-eabi/bin/../lib/gcc/arm-none-eabi/13.2.1/../../../../arm-none-eabi/lib/thumb/v7e-m+fp/softfp/libc_nano.a(libc_a-closer.o): in function `_close_r':
closer.c:(.text._close_r+0xc): warning: _close is not implemented and will always fail
```
Discussed here https://community.arm.com/support-forums/f/compilers-and-libraries-forum/53519/arm-gcc-11-3-rel1-newlib-nano-linker-warning-_read-is-not-implemented-and-will-always-fail
Looks like it works fine and the code calling this is not present in final binary but it is a bit strange. I did not succeed to make it go away when messing with nosys specs as mentioned in various places. However I don't see this in bootloader build (maybe something from stdio is not pulled in there at all, in main FW it may be there but optimized out).

- For BANGLE2 I have tested both bootloader and main firmware and in general it seems to work for me - Espruino works, I see no issue with any apps I use. Bootloader can start Nordic DFU, can also flash update from storage file however there is one issue - the screen output/font rendering in bootloader is somehow garbled most probably  when printing space character (?), font letters can be read but there are random pixels set between that. I briefly checked font rendering but don't see how space is rendered, any tips? Can be different layout after linking or alignment issue or some timing issue however other font letters are OK and screen works perfectly fine in main firmware.

fanoush:
[...]

PR #2458, opened 2024-01-26
for code in RAM use different section than .data

fanoush:
fixes newer gcc compilers, inspired by https://gcc.gnu.org/bugzilla/show_bug.cgi?id=103698#c4

as discussed in #2455 

WIP, maybe not all linker files are fixed

### DIFF
--- a/targetlibs/nrf5x_12/nrf5x_linkers/banglejs_dfu.ld
+++ b/targetlibs/nrf5x_12/nrf5x_linkers/banglejs_dfu.ld
@@ -71,6 +71,15 @@ SECTIONS
 SECTIONS
 {
   . = ALIGN(4);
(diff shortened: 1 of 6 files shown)


gfwilliams:
Thanks for this! *Actually* I'm pretty sure I gave up on running the code from RAM - the idea was the bootloader could update itself but I never got it working well and it seemed too dangerous.

So we could probably just remove all of the `__attribute__( ( long_call, section(".data") ) )` in the bootloader...

Not sure what you think?

fanoush:
> So we could probably just remove all of the __attribute__( ( long_call, section(".data") ) ) in the bootloader...

Yes, makes sense, so then just forget about this PR and close it :-)

Issue #2456, opened 2024-01-26
SEGV on unknown address in jsvStringIteratorPrintfCallback


Ye0nny:
###### Espruino revision
Commit: https://github.com/espruino/Espruino/commit/fcc9ba4f12d79710f93d47ea37987cde0efce410
Version: 2v20

###### Build platform
Ubuntu 20.04.5 LTS (Linux 5.4.0-144-generic x86_64)

###### Build steps
```sh
export CCFLAGS='-g -fsanitize=address -fno-omit-frame-pointer'
export LDFLAGS='-fsanitize=address'
DEBUG=1 make
```

###### Test case
<details><summary>testcase</summary>
<p>

```javascript
d = new Date ( - 2147483648 , 32 ) ; 
if ( " Invalid Date " != d . toString ( ) ) throw ' Expected " Invalid Date " , but got : " ' + d + ' " ' ;
```
</p>
</details>

```javascript
// poc.js
d = new Date ( - 2147483648 , 32 ) ; 
d . toString ( )
```

###### Execution steps & Output
```
$ ./espruino poc.js

 ____                 _
|  __|___ ___ ___ _ _|_|___ ___
|  __|_ -| . |  _| | | |   | . |
|____|___|  _|_| |___|_|_|_|___|
         |_| espruino.com
 2v20 (c) 2023 G.Williams

Espruino is Open Source. Our work is supported
only by sales of official boards and donations:
http://espruino.com/Donate

AddressSanitizer:DEADLYSIGNAL
=================================================================
==1936175==ERROR: AddressSanitizer: SEGV on unknown address 0x5595cc335f38 (pc 0x5595cf7bb33e bp 0x7fffd42959f0 sp 0x7fffd42959e0 T0)
==1936175==The signal is caused by a READ memory access.
#0 0x5595cf7bb33d in jsvStringIteratorPrintfCallback src/jsvar.c:1864
#1 0x5595cf7c94c1 in vcbprintf src/jsutils.c:818
#2 0x5595cf7bb570 in jsvVarPrintf src/jsvar.c:1890
#3 0x5595cf832863 in jswrap_date_toString src/jswrap_date.c:699
#4 0x5595cf7ca79e in jsnCallFunction src/jsnative.c:222
#5 0x5595cf7cc85e in jspeFunctionCall src/jsparse.c:656
#6 0x5595cf7ce45a in jspeFactorFunctionCall src/jsparse.c:1322
[...]

PR #2459, opened 2024-01-28
Bugfix for #2456 (PR #2230) - Date library bug fixes

deirdreobyrne:
There were two things wrong with the Date library

1. The algorithm required divisions to round down, but integer division rounds-towards-zero. This was creating artifacts like negative month numbers.
2. There was no range checking on the inputs. Inputs large enough in magnitude could cause overflows in some of the integer operations.

Also the spelling on the function "fromCalendarDate" has been corrected. And the documentation for the "setMonth" method was wrong.

Date now implements the Gregorian calendar for over 1.2 million years either side of the present. Inputs outside that range will throw an error. Note that the Gregorian calendar actually came into existence in 1582, but memory is too tight on some platforms (notably PICO_R1_3) to do anything intelligent with that fact.

### DIFF
--- a/boards/PICO_R1_3.py
+++ b/boards/PICO_R1_3.py
@@ -46,6 +46,7 @@ info = {
      'DEFINES+=-DUSE_USB_OTG_FS=1  -DPICO -DPICO_1V3',
      'DEFINES+=-DPIN_NAMES_DIRECT=1', # Package skips out some pins, so we can't assume each port starts from 0
      'DEFINES += -DESPR_USE_STEPPER_TIMER=1', # Build in the code for stepping using the timer
+     'DEFINES += -DESPR_LIMIT_DATE_RANGE', # not enough code memory left for the full range of Date()
      'STLIB=STM32F401xE',
      'PRECOMPILED_OBJS+=$(ROOT)/targetlibs/stm32f4/lib/startup_stm32f401xx.o'
    ]
--- a/libs/filesystem/jswrap_fs.c
+++ b/libs/filesystem/jswrap_fs.c
@@ -358,7 +358,7 @@ JsVar *jswrap_fs_stat(JsVar *path) {
       date.month = (int)(((info.fdate>>5)&15)-1);  // TomWS: Month is 0 based.
       date.day = (int)((info.fdate)&31);
       TimeInDay td;
-      td.daysSinceEpoch = fromCalenderDate(&date);
+      td.daysSinceEpoch = fromCalendarDate(&date);
       td.hour = (int)((info.ftime>>11)&31);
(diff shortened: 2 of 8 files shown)


deirdreobyrne:
All I did was add in the test case I was sent by email - no changes were made to the base code - and PICO_R1_3 is failing to build due to lack of memory.

Converting to draft, and seeing if I can save a few bytes :(

gfwilliams:
Thanks! I'm amazed those changes saved enough space!

Just thinking - if you made INTEGER_DIVIDE_FLOOR into a function, could that save enough space that you didn't have to ifdef it for the Pico?

deirdreobyrne:
Doubtful - as INTEGER_DIVIDE_FLOOR is simply (a/b) on the Pico (in other words, it actually divides-towards-zero, which is why the Date() library has to be limited to after 1601 on that platform), and the STM32F401CDU6 has a divide instruction, so all the moving and shaking that needs to happen for a function call would probably take up more space.

However since I've realised one of the tests will *definitely* fail on the Pico (the one you sent me - new Date(-12,8)) I'll make some changes later and see. Although the cross-compiler I use for the Pico generates code far bigger than whatever github uses, so I'll have to use github to see what, if any, difference it makes.

deirdreobyrne:
I believe the numbers show I was right - 327616 vs 327640 bytes for the Pico.

However splitting the INTEGER_DIVIDE_FLOOR into a function does save bytes on all other platforms (as the integerDivideFloor function is reasonably complicated, so there are fewer bytes to call the function than to perform it).

gfwilliams:
> However splitting the INTEGER_DIVIDE_FLOOR into a function does save bytes on all other platforms 

Yes - what I meant really was maybe with INTEGER_DIVIDE_FLOOR as a function, enough bytes were saved you wouldn't have to do anything special for the Pico at all and could just use the normal code? 

No problem about pico testing - but if you think you really do still need ESPR_LIMIT_DATE_RANGE even with the INTEGER_DIVIDE_FLOOR function then you could I guess at least build a platform that you're confident with, with the `ESPR_LIMIT_DATE_RANGE` define in it

deirdreobyrne:
> Yes - what I meant really was maybe with INTEGER_DIVIDE_FLOOR as a function, enough bytes were saved you wouldn't have to do anything special for the Pico at all and could just use the normal code?

Doing so consumed 96 more bytes on my compile of the pico code, but there are only 34 bytes left!

I've done the pico testing as you suggested - all is good.

And note that I extended the Date() range for the Pico back to 1500 - back to before the introduction of the Gregorian calendar.

I think it's about as good as we can make it.

gfwilliams:
Ok, great - thanks for this!

PR #2469, opened 2024-02-23
added the defines to the platform file

brendena:
Based on the topics [here](https://github.com/espruino/Espruino/pull/2468), I've edited the code to put the defines in the platform_config.h

**Key points**

**How**
So what about this: we modify the Makefile to call scripts/build_platform_config.py $(BOARD) $(HEADERFILENAME) $(DEFINES) and then modify build_platform_config to pick up those defines from the command line and create a block like out define.

**No side effects**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -716,7 +716,7 @@ OBJS = $(PRECOMPILED_OBJS) $(SOURCEOBJS)
 # -nodefaultlibs -nostdlib -nostartfiles
 
 # -fdata-sections -ffunction-sections are to help remove unused code
-CFLAGS += $(OPTIMIZEFLAGS) -c $(ARCHFLAGS) $(DEFINES) $(INCLUDE)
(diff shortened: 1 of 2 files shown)


brendena:
Oooo, doesn't like ().  Lets try this again

fanoush:
I would never guessed the ESPR_AUTOCOMPLETE_NOT_REQUIRED does what it does from its name. what about something similar to BOARD_DEFINES_ALREADY_DEFINED? And also some comment somewhere in makefiles and/or generated header would be nice to explain why this hack exists (like  "these are defines passed already in command line, helps vscode to parse them from here")

And does it actually work in vscode now?

gfwilliams:
Great - thanks!

The define thing was my fault - merging now and changing to `ESPR_DEFINES_ON_COMMANDLINE` with a comment.

Also I'm wondering if we can pass the defines unescaped and unquoted (like is done for GCC) and save that fiddling in the Makefile (for the folks that insist on using MacOS some of the built-in unix functions are subtly different - it may be fine here but the less I can rely on them the better :)

gfwilliams:
Right, I think this is all merged in now - not sure why GitHub hasn't picked it up (it usually does).

PR #2472, opened 2024-03-01
Disabling unused warnings

brendena:
Removing warning is usually a bad thing, but for a project that supports so many devices and pieces of hardware i think these kinds of errors are hard to avoid.  

So personally i remove these warnings to make potentially larger more problematic problems easier to see.  

But this is just a suggestion.


brendena:
Disabling warning is usually a bad thing so i understand.  

Since you're already dealing with this.  I'll follow suit.

PR #2473, opened 2024-03-01
Building EMSCRIPTEN boards on the latest EMCC

brendena:
So this pull is updated to work with the latest EMCC.  So there's just two changes.  

* One is a name change on the build
* The other is a javascript function definition.  

I'm not entirely what changed to necessitate this function needing to be defined.  But it seemed to fix the problem.   

Error

error: undefined symbol: emscripten_memcpy_js (referenced by root reference (e.g. compiled C/C++ code))


gfwilliams:
Great - thanks! Still seems to work with the old EMCC too

PR #2476, opened 2024-03-20
Deprecate currency information in `locale`

atjn:
Closes https://github.com/espruino/BangleApps/issues/3269
Analogous to https://github.com/espruino/BangleApps/pull/3278

I am sorry but it is quite overwhelming to try and figure out how I can test this. Could you maybe point me to some docs on how to quickly build an emulator for the runtime? Then I will happily test it :)


atjn:
BTW is it on purpose that the `number` implementation is different between this package and the one in the app loader? It seems like a pretty big problem to me.

gfwilliams:
Do you actually have a Bangle.js device? Because if so, you can just go to the GitHub actions tab, download the built firmware and flash it.

I'm not super happy with adding getters/setters in here though - it's just complicating things instead of simplifying (which I guess was the idea of deprecating currency?). Maybe just have the single change to `currency`?

atjn:
> Do you actually have a Bangle.js device? Because if so, you can just go to the GitHub actions tab, download the built firmware and flash it.
Yes thanks I'll try that

gfwilliams:
Thanks - I'll get this in.

As for the number thing - yes, that's a good point. Just one of those things where someone updated the app but didn't think to push it to Bangle.js. I'll update that now as well.

PR #2477, opened 2024-03-21
locale: fix trailing space on date < 10

thyttan:
The trailing whitespace would make the date string look uncentered on e.g. `antonclk` clock app on Bangle.js.

Possibly similar change should be made to the `locale` library of espruino/BangleApps - but I'm not sure.

Also not sure if other entries could benefit from a `trim()` as well.

I have NOT built and flashed this to try it out.


gfwilliams:
Thanks! That's interesting actually. On Node.js:

```
d = new Date("8 Feb 2020")
d.toString().substr(4,11)
='Feb 08 2020'
```

so it looks like Bangle.js is perhaps doing it wrong by not padding the day number out. Either way adding `trim` won't break anything if this gets fixed in the future

PR #2478, opened 2024-03-22
Esp32 cleanup jswarn

MaBecker:
for details see comments with first commit

### DIFF
--- a/libs/network/esp32/jswrap_esp32_network.c
+++ b/libs/network/esp32/jswrap_esp32_network.c
@@ -512,7 +512,7 @@ static esp_err_t event_handler(void *ctx, system_event_t *event)
     jsvObjectSetChildAndUnLock(jsDetails, "mac", jsvNewFromString(temp));
     sprintf(temp, "%d", event->event_info.disconnected.reason);
     jsvObjectSetChildAndUnLock(jsDetails, "reason", jsvNewFromString(temp));
-    jsvObjectSetChildAndUnLock(jsDetails, "msg", jsvNewFromString(wifiReasonToString(event->event_info.disconnected.reason))))
+    jsvObjectSetChildAndUnLock(jsDetails, "msg", jsvNewFromString(wifiReasonToString(event->event_info.disconnected.reason)));
     sendWifiEvent(event->event_id, jsDetails);
     return ESP_OK;
   } // End of handle SYSTEM_EVENT_STA_DISCONNECTED
--- a/libs/network/esp32/jswrap_esp32_network.c
+++ b/libs/network/esp32/jswrap_esp32_network.c
@@ -512,8 +512,7 @@ static esp_err_t event_handler(void *ctx, system_event_t *event)
     jsvObjectSetChildAndUnLock(jsDetails, "mac", jsvNewFromString(temp));
     sprintf(temp, "%d", event->event_info.disconnected.reason);
     jsvObjectSetChildAndUnLock(jsDetails, "reason", jsvNewFromString(temp));
-    sprintf(temp,"%s",wifiReasonToString(event->event_info.disconnected.reason));
-    jsvObjectSetChildAndUnLock(jsDetails, "msg", jsvNewFromString(temp));
+    jsvObjectSetChildAndUnLock(jsDetails, "msg", jsvNewFromString(wifiReasonToString(event->event_info.disconnected.reason))))
     sendWifiEvent(event->event_id, jsDetails);
     return ESP_OK;
   } // End of handle SYSTEM_EVENT_STA_DISCONNECTED
--- a/libs/network/esp32/jswrap_esp32_network.c
+++ b/libs/network/esp32/jswrap_esp32_network.c
@@ -1552,8 +1552,9 @@ void jswrap_wifi_setHostname(
   // now update mDNS
   startMDNS(hostname);
 
-  if (jsvIsFunction(jsCallback))
+  if (jsvIsFunction(jsCallback)) {
     jsiQueueEvents(0, jsCallback, 0, 0);
+  }
 }
 
 static uint8_t seq_no;
--- a/libs/network/esp32/jswrap_esp32_network.c
+++ b/libs/network/esp32/jswrap_esp32_network.c
@@ -305,7 +305,7 @@ static char *wifiErrorToString(esp_err_t err){
     case 0x300D: return "WiFi is in sleep state(RF closed) and wakeup fail";
     case 0x300E: return "The caller would block";
     case 0x300F: return "Station still in disconnect status";
-    default: return "unkown WiFi error, see wifiErrorToString";
+    default: return "unknown WiFi error, see wifiErrorToString";
   }
 }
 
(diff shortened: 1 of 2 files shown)


libs/network/esp32/jswrap_esp32_network.c
```diff
@@ -169,6 +169,7 @@ static char *htModeToString(wifi_second_chan_t htMode) {
  * Convert a Wifi reason code to a string representation.
  */
 static char *wifiReasonToString(uint8_t reason) {
+  jsDebug("wifiReasonToString %d",reason);
```
gfwilliams, line 172: Is this right? shouldn't it be `jsDebug(DBG_INFO,` ? This might break something

libs/network/esp32/jswrap_esp32_network.c
```diff
@@ -252,57 +253,59 @@ static char *wifiModeToString(wifi_mode_t mode) {
...
+    case SYSTEM_EVENT_MAX: return "MAX";
+    default: return "unknown event, see wifiEventToString";
   }
 }
 
 /**
  * convert WiFi error to a string value.
  */
 static char *wifiErrorToString(esp_err_t err){
+  jsDebug("wifiErrorToString %d", err);
```
gfwilliams, line 291: same...

gfwilliams:
Otherwise this looks good though - it'll be nice to avoid those annoying messages

MaBecker:
Thanks for checking

MaBecker:
Please do not merge yet, will let you when ready to merge

MaBecker:
@gfwilliams  can you please share how to handle data for callback functions.


Is this how to?

```js
  JsVar *tmp = jsvNewObject();

  // Schedule callback if a function was provided
  jsvObjectSetChildAndUnLock(tmp, "hostname", jsvNewFromString(hostname));
  if (jsvIsFunction(jsCallback)) {
    JsVar *params[2];
    params[0] = jsvNewWithFlags(JSV_NULL);
    params[0] = tmp;
    jsiQueueEvents(NULL, jsCallback, params, 1);
    jsvUnLock(params[0]);
  }
```

gfwilliams:
Yes, that looks good - but some typos in there....

* the second `params[0]` should be `[1]`?
* jsiQueueEvents last argument should be 2 because 2 arguments
* `tmp` should be unlocked at the end of the function in the scope it was allocated in

MaBecker:
I have no clue what 

params[0] = jsvNewWithFlags(JSV_NULL);

is for

gfwilliams:
Did that code come from somewhere? All that line does is make the first argument `null` - we were trying to keep the API the same between different ways to connect Espruino to the net, and also to keep it with a Node.js-ish API.

In some Node.js APIs they have a callback with an `error` argument. It's null of there's no error, or a String if there is an error: https://github.com/espruino/Espruino/blob/8f3a9cb52/libs/network/wiznet/jswrap_wiznet.c#L325

MaBecker:
> Yes, that looks good - but some typos in there....
> 
> * the second `params[0]` should be `[1]`?
> * jsiQueueEvents last argument should be 2 because 2 arguments
> * `tmp` should be unlocked at the end of the function in the scope it was allocated in

like this? 

```js
JsVar *tmp = jsvNewObject();

  // Schedule callback if a function was provided
  jsvObjectSetChildAndUnLock(tmp, "hostname", jsvNewFromString(hostname));
  if (jsvIsFunction(jsCallback)) {
    JsVar *params[2];
    params[0] = jsvNewWithFlags(JSV_NULL);
    params[1] = tmp;
    jsiQueueEvents(NULL, jsCallback, params, 2);
    jsvUnLock(params[1]);
  }
  jsvUnLock(tmp)

```

gfwilliams:
Nope - you changed `jsvUnLock(params[0]);` to `jsvUnLock(params[1]);` so now you're not unlocking the first arg but are unlocking the second one twice.

This should be fine:

```
JsVar *tmp = jsvNewObject();

  // Schedule callback if a function was provided
  jsvObjectSetChildAndUnLock(tmp, "hostname", jsvNewFromString(hostname));
  if (jsvIsFunction(jsCallback)) {
    JsVar *params[2];
    params[0] = jsvNewWithFlags(JSV_NULL);
    params[1] = tmp;
    jsiQueueEvents(NULL, jsCallback, params, 2);
    jsvUnLock(params[0]);
  }
  jsvUnLock(tmp);
```

gfwilliams:
So do you think this is all ready now?

libs/network/esp32/jswrap_esp32_network.c
```diff
@@ -508,6 +512,8 @@ static esp_err_t event_handler(void *ctx, system_event_t *event)
     jsvObjectSetChildAndUnLock(jsDetails, "mac", jsvNewFromString(temp));
     sprintf(temp, "%d", event->event_info.disconnected.reason);
     jsvObjectSetChildAndUnLock(jsDetails, "reason", jsvNewFromString(temp));
+    sprintf(temp,"%s",wifiReasonToString(event->event_info.disconnected.reason));
```
gfwilliams, line 515: I think `wifiReasonToString` now returns strings that are over 35 chars in length with the new error messages? That means it'll overflow `temp` because `sprintf` doesn't check bounds - `temp` needs to be bigger

gfwilliams: Actually scratch that - you can just do `jsvObjectSetChildAndUnLock(jsDetails, "msg", jsvNewFromString(wifiReasonToString(event->event_info.disconnected.reason))))` ?

libs/network/esp32/jswrap_esp32_network.c
```diff
@@ -542,23 +548,23 @@ static esp_err_t event_handler(void *ctx, system_event_t *event)
...
 
   if (event->event_id == SYSTEM_EVENT_STA_START) {
-    s_retry_num = 5; // Try to connect 5 times
-      // Perform an esp_wifi_connect
-      esp_err_t err = esp_wifi_connect();
-      if (err != ESP_OK) {
-        jsError( "Wifi: event_handler STA_START: esp_wifi_connect: %d(%s)", err,wifiErrorToString(err));
-        return NULL;
-      }
+    JsVar *jsDetails = jsvNewObject();
```
gfwilliams, line 551: I'm not convinced this is ever used or freed? So it's a memory leak - you probably don't want this

libs/network/esp32/jswrap_esp32_network.c
```diff
@@ -252,57 +253,59 @@ static char *wifiModeToString(wifi_mode_t mode) {
...
+    case 0x3007: return "WiFi internal control block of station or soft-AP error";
+    case 0x3008: return "WiFi internal NVS module error";
+    case 0x3009: return "MAC address is invalid";
+    case 0x300A: return "SSID is invalid";
+    case 0x300B: return "Password is invalid";
+    case 0x300C: return "Timeout error";
+    case 0x300D: return "WiFi is in sleep state(RF closed) and wakeup fail";
+    case 0x300E: return "The caller would block";
+    case 0x300F: return "Station still in disconnect status";
+    default: return "unkown WiFi error, see wifiErrorToString";
```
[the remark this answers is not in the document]
MaBecker: Thanks

libs/network/esp32/jswrap_esp32_network.c
MaBecker: Nice, just changed it

gfwilliams:
Great - thanks!

MaBecker:
Thanks for checking, tried to fix your findings. Would say good to go for now.

PR #2494, opened 2024-04-23
NRF.updateServices() needs fixing (fix2292)

MaBecker:
many thanks to fanoush


gfwilliams:
Great - thanks! I guess this must have come up with a compile warning - we should probably try and have a run through and see what other stuff like this the compiler is warning about?

PR #2496, opened 2024-04-25
typescript: allow Promise.resolve() [no args]

bobrippling:
To permit creating the equivalent of `async () => {}` via `() => Promise.resolve()`


gfwilliams:
thanks!

PR #2503, opened 2024-05-24
NRF: `getAddress(1)` returns the current BLE address

bobrippling:
This allows callers of `NRF.getAddress()` to get the current address, for example when `setAddress()` has been called - such as for [Open Haystack](https://banglejs.com/apps/?id=openhaystack)

### DIFF
--- a/libs/bluetooth/jswrap_bluetooth.c
+++ b/libs/bluetooth/jswrap_bluetooth.c
@@ -707,14 +707,20 @@ void jswrap_ble_eraseBonds() {
     "class" : "NRF",
     "name" : "getAddress",
     "generate" : "jswrap_ble_getAddress",
+    "params" : [
+      ["current", "JsVar", "If true, return the current address rather than the default"]
+    ],
     "return" : ["JsVar", "MAC address - a string of the form 'aa:bb:cc:dd:ee:ff'" ]
 }
-Get this device's default Bluetooth MAC address.
+Get this device's default or current Bluetooth MAC address.
 
 For Puck.js, the last 5 characters of this (e.g. `ee:ff`) are used in the
 device's advertised Bluetooth name.
 */
-JsVar *jswrap_ble_getAddress() {
+JsVar *jswrap_ble_getAddress(JsVar *current) {
+  if (jsvGetBool(current)) {
+    return jsvObjectGetChildIfExists(execInfo.hiddenRoot, BLE_NAME_MAC_ADDRESS);
+  }
 #ifdef NRF5X
   uint32_t addr0 =  NRF_FICR->DEVICEADDR[0];
   uint32_t addr1 =  NRF_FICR->DEVICEADDR[1];
--- a/libs/bluetooth/jswrap_bluetooth.c
+++ b/libs/bluetooth/jswrap_bluetooth.c
@@ -719,7 +719,8 @@ device's advertised Bluetooth name.
 */
 JsVar *jswrap_ble_getAddress(JsVar *current) {
   if (jsvGetBool(current)) {
-    return jsvObjectGetChildIfExists(execInfo.hiddenRoot, BLE_NAME_MAC_ADDRESS);
+    JsVar *addr = jsvObjectGetChildIfExists(execInfo.hiddenRoot, BLE_NAME_MAC_ADDRESS);
+    if (addr) return addr;
   }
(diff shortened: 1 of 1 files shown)


bobrippling:
I suppose there's no way to check if this feature exists (except firmware version), perhaps a separate function would be better

gfwilliams:
I think this is ok (same function) - you can always check firmware version.

However I think it still needs work - it looks like if you call getAddress(1) when you haven't set an address, it returns undefined, when it should fall through and return the default address?

bobrippling:
Good spot, thanks - updated the PR

gfwilliams:
Thanks! Looks good - ideally I'd use `bool current` and set the type to `bool` in the JSON but I'll just merge this now and make the change myself.

gfwilliams:
Just FYI, if you search for jswrap_ble_getAddress you'll see it is used in a bunch of places that weren't updated, so it's getting called with random values for `current`.

I'm honestly not sure why the build didn't fail - maybe I need to look at making warnings about calling functions with incorrect parameters into full on errors.

bobrippling:
> Just FYI, if you search for jswrap_ble_getAddress you'll see it is used in a bunch of places that weren't updated, so it's getting called with random values for `current`. I've just put in a fix for it I hope.
> 
> I'm honestly not sure why the build didn't fail - maybe I need to look at making warnings about calling functions with incorrect parameters into full on errors.

Ah yeah, that's a [C thing](https://cigix.me/c17#6.7.6.3.p14) - this would've caught it when it was a no-argument function:

```diff
-JsVar *jswrap_ble_getAddress();
+JsVar *jswrap_ble_getAddress(void);
```

There may be similar bugs, to catch them you can use [`-Wstrict-prototypes`](https://gcc.gnu.org/onlinedocs/gcc-13.1.0/gcc/Warning-Options.html#index-Wstrict-prototypes).

I've done a [quick check](https://github.com/bobrippling/Espruino/actions/runs/9553564689/job/26332656923?pr=1) and [`jswrap_graphics_theme`](https://github.com/espruino/Espruino/blob/49b8ac8e97efaac9f562ad879e7cf3f927a57dd8/libs/graphics/jswrap_graphics.c#L4575-L4575) appears - there may be more but `make` isn't invoked with `-k` so it stops at the first error.

```
libs/graphics/jswrap_graphics.c:4575:8: error: conflicting types for 'jswrap_graphics_theme'; have 'JsVar *(JsVar *)' {aka 'struct JsVarStruct *(struct JsVarStruct *)'}
 4575 | JsVar *jswrap_graphics_theme(JsVar *parent) {
      |        ^~~~~~~~~~~~~~~~~~~~~
In file included from libs/graphics/jswrap_graphics.c:16:
libs/graphics/jswrap_graphics.h:94:8: note: previous declaration of 'jswrap_graphics_theme' with type 'JsVar *(void)' {aka 'struct JsVarStruct *(void)'}
   94 | JsVar *jswrap_graphics_theme(void);
      |        ^~~~~~~~~~~~~~~~~~~~~
```

gfwilliams:
Thanks! That's interesting - I always assumed `(void)` vs `()` was effectively the same!

I've just fixed `jswrap_graphics_theme`. 

I'd prefer not to have to change to `(void)` everywhere if there's a choice - it just seems like it'll be a pretty huge PR... although looking at https://github.com/espruino/Espruino/compare/master...bobrippling:Espruino:fix/unspec-args you only had to actually change the header files?

Don't suppose you know if there's a compiler switch to enable the behaviour without having to change it?

bobrippling:
Yeah you've got two options really:

1. Compile the codebase as [C23](https://en.wikipedia.org/wiki/C23_(C_standard_revision)) (`-std=c2x`), to gain this:
> Functions with no arguments listed in the prototype void foo() are understood as taking no arguments (see removal of K&R function declarations)

The downsides are that there may be other things which need fixing to align with the standard upgrade, but given that the codebase is already C11 (or the GNU variant), I don't think it's a real issue:
https://github.com/espruino/Espruino/blob/49b8ac8e97efaac9f562ad879e7cf3f927a57dd8/Makefile#L238-L238

And this option is very cheap to try out.

Or:

2. Get gcc to point out all the problematic declarations with `-Werror=strict-prototypes`. This will still require changing of the prototypes themselves, but it's only the header files that need to change, except for file-local/`static` functions

PR #2504, opened 2024-05-24
Object.defineProperty can be called on functions and arrays (and the root)

### LINKED ISSUE #2501: `Object.defineProperty(<function>, ...)`
For debugging, I'd like to be able to do
```javascript
Object.defineProperty(Bangle, ...)
```

At the moment I'm told it's because `Bangle` is not an object:

https://github.com/espruino/Espruino/blob/01467c2a539a737e2204af3f759348d0d1a7326e/src/jswrap_object.c#L601-L604

If it seems reasonable, I'll make a PR to adjust the check to permit functions - I believe this is ok to be done on all functions:

https://github.com/espruino/Espruino/blob/01467c2a539a737e2204af3f759348d0d1a7326e/src/jsvar.c#L206-L209

bobrippling:
This permits `Object.defineProperty(Bangle, ...)`, for example

Closes #2501

### DIFF
--- a/src/jswrap_object.c
+++ b/src/jswrap_object.c
@@ -607,8 +607,8 @@ will be ignored.
 
  */
 JsVar *jswrap_object_defineProperty(JsVar *parent, JsVar *propName, JsVar *desc) {
-  if (!jsvIsObject(parent)) {
-    jsExceptionHere(JSET_ERROR, "First argument must be Object, got %t", parent);
+  if (!jsvIsObject(parent) && !jsvIsFunction(parent) && !jsvIsArray(parent)) {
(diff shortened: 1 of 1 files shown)


bobrippling:
Would appreciate an opinion on the `jsvIsRoot()` part - I guess that's for the global object but am not certain.

gfwilliams:
I'm pretty sure jsvIsRoot(parent) isn't needed? I thought root was treated as an object so jsvObject should return true?

gfwilliams:
@bobrippling ? any thoughts on this?

bobrippling:
Ah yes you're right - thank you

gfwilliams:
Thanks!

PR #2510, opened 2024-06-07
fix PRECOMPILED_OBJS for all SDKs to be .o files like it already is for SDK12

fanoush:
allows gcc to use default compiler rules = honor ASFLAGS


gfwilliams:
Are you sure this works ok and it does still compile the `.S` though? I thought we had to explicitly tell it we needed to compile the `.S` files?

fanoush:
Well, yes, it is already like that for SDK12, see last unmodified line of the patch
```
PRECOMPILED_OBJS += $(NRF5X_SDK_PATH)/components/toolchain/gcc/gcc_startup_nrf52.o
```
Not sure why did you put .S into **OBJS** for other SDKs than 12, maybe that was some kind of workaround for something(?), but for me it seems to work fine with .o.  It finds .S file and additionally honors ASFLAGS so the board file can have this
```
     'ASFLAGS += -D__STARTUP_CLEAR_BSS -D__START=main',
     'LDFLAGS += -nostartfiles',
```
also for SDKs other than 12.

Also all build checks are green.

PR #2511, opened 2024-06-11
Fix NUCLEOF411RE binary

AkosLukacs:
Looks like this commit https://github.com/espruino/Espruino/commit/29a243226ea2ec15a68f8aed9bd3c7ff736c185c broke the Nucleo F411 binary, details here: https://forum.espruino.com/conversations/396990/#comment17415645 

Reverted the mentioned commit's changes to the F411 board file, and changed `'STLIB=STM32F401xE'` to `'STLIB=STM32F411xE'` since it's a 411 :) No idea if that makes any difference... 
Did some basic testing, so far it seems to work fine.

Edit: don't have a 401 Nucleo, so I have no idea if whether that binary works...

### DIFF
--- a/boards/NUCLEOF411RE.py
+++ b/boards/NUCLEOF411RE.py
@@ -27,16 +27,15 @@ info = {
  'build' : {
    'optimizeflags' : '-O3',
    'libraries' : [
+     'NET',
      'GRAPHICS',
      'NEOPIXEL'
    ],
    'makefile' : [
-     'DEFINES+=-DSAVE_ON_FLASH_MATH', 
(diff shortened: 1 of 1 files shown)


gfwilliams:
That's a bit concerning. Can you confirm if SAVE_ON_FLASH_MATH causes the problem? It shouldn't, so maybe leave it in?

We use ESPR_PACKED_SYMPTR in other places and it seems to work fine - I wonder why it doesn't in this case...

As far as I remember, those changes were made because I was having trouble getting the binary size down enough that it would fit in the chip. Have you been able to test the firmware with these changes?

It seems like re-adding network support, and then also removing the two features that would have saved storage space is almost certainly going to push the firmware size back up to the point I'm constantly fighting to get a build.

PR #2512, opened 2024-06-12
clean gcc startup code also for nrf5x SDK >=15

fanoush:
For SDK15 and up the gcc startup code is  in different location.
When not properly cleaned it breaks when gcc startup code is built with different flags for different boards.
This is follow-up fix to PR #2510 which makes building with different flags possible.


gfwilliams:
Ahh - thanks!

PR #2518, opened 2024-06-20
Add note of missing RegExp features to RegExp class documentation

esdeboer:
Add Regular Expression note from http://www.espruino.com/Features to the RegExp class doc.

Didn't test if it works, Doxygen should support the markdown, but the config in the project didn't seem to include any of the class documentations


gfwilliams:
Thanks! This looks perfect!

PR #2521, opened 2024-06-21
fix overlays being cut off from the top

### LINKED ISSUE #2520: overlays are cut off from the top instead of landing at desired y position
Overlays are being cut off from the top, by about as many pixels as their y position should be, instead of landing at the desired y position.

To reproduce on a Bangle.js 2 (code copied from the docs about overlays):
```js
g.setBgColor(0, 1, 0).clear().reset();

var ovr = Graphics.createArrayBuffer(100,100,2,{msb:true});
ovr.transparent = 0;
ovr.palette = new Uint16Array([0,0,g.toColor("#F00"),g.toColor("#FFF")]);
ovr.setColor(1).fillRect({x:0,y:0,w:99,h:99,r:8});
ovr.setColor(3).fillRect({x:2,y:2,w:95,h:95,r:7});
ovr.setColor(2).setFont("Vector:30").setFontAlign(0,0).drawString("Hi",50,50);
Bangle.setLCDOverlay(ovr,38,38, {id: "myOverlay", remove: () => print("Removed")});
```

Expected outcome:
good

Actual outcome:
bad

A git bisect, using the emulator, leads to this being the first bad commit: fdbf0bfd2590bf241eabe3877b283a78391035ef

Moving the following line inside the for loop above it seems to fix things
https://github.com/espruino/Espruino/blob/31c5d60eb743587d9b505b5e7403eb4d8567eda1/libs/graphics/lcd_memlcd.c#L347
but that would revert part of 5a99fbdafedf5b8a551ac53f2370c76f63be9523
I've got absolutely no idea when partial updates happen, but I guess moving that iterator call inside the for loop would probably break something else ¯\\\_(ツ)_/¯

ssievert42:
Previously, overlays were always positioned at y=0 and the y pos was cut off from the top part of the buffer.

That behaviour was introduced by: fdbf0bfd2
Since that commit, the coordinates of a `GfxDrawImageLayer` were changed to be the position in pixels, shifted to the left by 8 bits.
The overlay y position now has to be shifted to the left by 8 bits as well, before assigning it to the y start coordinate of a `GfxDrawImageLayer`.

Closes #2520 (See that issue for screenshots and some more info.)

### DIFF
--- a/libs/graphics/lcd_memlcd.c
+++ b/libs/graphics/lcd_memlcd.c
@@ -319,7 +319,7 @@ void lcdMemLCD_flip(JsGraphics *gfx) {
     // initialise image layer
     GfxDrawImageLayer l;
     l.x1 = 0;
-    l.y1 = ovY;
+    l.y1 = ovY<<8;
     l.img = overlayImg;
     l.rotate = isRotated180 ? 3.141592 : 0;
     l.scale = 1;


ssievert42:
Superseded by #2522

PR #2526, opened 2024-07-01
discussion: SetUI Q3 changes

thyttan:
@gfwilliams

Related issues:
https://github.com/espruino/BangleApps/issues/3435
https://github.com/espruino/BangleApps/pull/3452

I've tweaked setUI for Bangle.js 2 to try and  achieve goals discussed previously in the issues linked above, regarding the opinion that hw button edge should be rising - and that setting custom handlers on any mode of setui should be possible.

Here I tweaked `spotrem`  to take advantage of these setUI changes to move all input logic inside setUI: https://github.com/espruino/BangleApps/pull/3485

**NOT THOROUGHLY TESTED!**

### DIFF
--- a/libs/js/banglejs/Bangle_setUI_Q3.js
+++ b/libs/js/banglejs/Bangle_setUI_Q3.js
@@ -85,28 +85,28 @@
       setWatch(Bangle.showLauncher, BTN1, {repeat:1,edge:"falling"})
     ];
   } else if (mode=="custom") {
-    if (options.clock) Bangle.CLOCK=1;
-    if (options.touch)
-      Bangle.touchHandler = options.touch;
-    if (options.drag) {
-      Bangle.dragHandler = options.drag;
-      Bangle.on("drag", Bangle.dragHandler);
-    }
-    if (options.swipe) {
-      Bangle.swipeHandler = options.swipe;
-      Bangle.on("swipe", Bangle.swipeHandler);
-    }
-    if (options.btn) {
-      Bangle.btnWatches = [
-        setWatch(function() { options.btn(1); }, BTN1, {repeat:1,edge:"falling"})
-      ];
-    } else if (options.clock) {
-      Bangle.btnWatches = [
-        setWatch(Bangle.showLauncher, BTN1, {repeat:1,edge:"falling"})
-      ];
-    }
   } else
     throw new Error("Unknown UI mode "+E.toJS(mode));
+  if (options.clock) Bangle.CLOCK=1;
+  if (options.touch)
+  Bangle.touchHandler = options.touch;
+  if (options.drag) {
+    Bangle.dragHandler = options.drag;
+    Bangle.on("drag", Bangle.dragHandler);
+  }
+  if (options.swipe) {
+    Bangle.swipeHandler = options.swipe;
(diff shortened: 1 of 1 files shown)


libs/js/banglejs/Bangle_setUI_Q3.js
```diff
@@ -82,31 +82,37 @@
...
+  Bangle.touchHandler = options.touch;
+  if (options.drag) {
+    Bangle.dragHandler = options.drag;
+    Bangle.on("drag", Bangle.dragHandler);
+  }
+  if (options.swipe) {
+    Bangle.swipeHandler = options.swipe;
+    Bangle.on("swipe", Bangle.swipeHandler);
+  }
+  if (options.btn) {
```
thyttan, line 101: Particularly interested in what you think about this solution to letting the user choose button edge (between rising/falling/both).

libs/js/banglejs/Bangle_setUI_Q3.js
```diff
@@ -82,31 +82,37 @@
...
+    if (Bangle.btnWatches) Bangle.btnWatches.forEach(clearWatch);
+    var e = "rising";
+    if ("object"==typeof options.btn) {
+      e = options.btn.edge;
+      options.btn = options.btn.fn;
+    }
+    Bangle.btnWatches = [
+      setWatch(function() { options.btn(1); }, BTN1, {repeat:1,edge:e})
+    ];
```
thyttan, line 110: @gfwilliams I'm particularly interested in what you think about this solution to letting the user choose button edge (between rising/falling/both). Since it's different from what you proposed with callbacks.

thyttan: Here it's used from `spotrem`: 
https://github.com/espruino/BangleApps/pull/3485/files#diff-f23a88305664ab4df5d78b95d4770ca5e752014d9742f2d36c1d7529500d61a6R109-R115

gfwilliams: Thanks, but honestly, I think I prefer the callback solution mentioned before - the main reason is that this was isn't backwards compatible. At least the old one would silently continue but not add the watch - if someone ran spotrem on an old firmware it'd just fail with an error and not do anything

gfwilliams: ... also I feel like the implementation is actually marginally tidier:

```
if (options.btn) Bangle.btnWatches.push(setWatch(options.btn.bind(options), BTN1, {repeat:1,edge:"rising"}))
if (options.btnRelease) Bangle.btnWatches.push(setWatch(options.btnRelease.bind(options), BTN1, {repeat:1,edge:"falling"}))
```

gfwilliams:
Just to add I think this still needs a lot of testing.

We've gone from in https://github.com/espruino/BangleApps/issues/3435 having an app that enables this for users that want it, to enabling it for everyone.

Personally I'd quite like to see this as an app that overrides `setUI` and then let users install it and get some feedback before we put this in for good, because right now I think the only other person to look at it is @bobrippling?

thyttan:
@gfwilliams thanks! I'll refactor to the callback solution and package it as an app.

PR #2534, opened 2024-07-29
Bangle_showClock: Set Bangle.CLOCK=1

thyttan:
Background:
  When moving from an app where e.g. `widanclk` (analog clock widget) is
  in use to the clock with a Bangle.showClock call the widget will not
  hide immediately.

The fix:
  Makes it so widgets that should hide on clock are hidden immediately
  when fast loading the clock, at the same time as the red back button
  widget.

### DIFF
--- a/libs/js/banglejs/Bangle_showClock.js
+++ b/libs/js/banglejs/Bangle_showClock.js
@@ -1,3 +1,4 @@
 (function() {
+  Bangle.CLOCK=1;
   Bangle.load(".bootcde");
 })
--- a/libs/js/banglejs/Bangle_showClock.min.js
+++ b/libs/js/banglejs/Bangle_showClock.min.js
@@ -1,1 +1,1 @@
-(function(){Bangle.load(".bootcde")})
+(function(){Bangle.CLOCK=1;Bangle.load(".bootcde");})


gfwilliams:
By 'not hide immediately' do you mean there's a glitch/flicker as the clock loads, or that it takes multiple seconds/minutes?

Because I'd have thought that when you load a clock, the clock itself should call `Bangle.setUI(` which will set CLOCK=1, and then it will call `Bangle.load/drawWidgets` after which should redraw. So if that's not happening it would be better to find out why as it might be some other issue.

Also, I'd have thought `Bangle.load` should delete `Bangle.CLOCK` anyway really - because if you're loading something that's not a clock then you don't want that to stay set

thyttan:
> By 'not hide immediately' do you mean there's a glitch/flicker as the clock loads, or that it takes multiple seconds/minutes?

It stays on for some tenth of a second longer than the red back button.

> Because I'd have thought that when you load a clock, the clock itself should call Bangle.setUI( which will set CLOCK=1, and then it will call Bangle.load/drawWidgets after which should redraw. So if that's not happening it would be better to find out why as it might be some other issue.

Yes, that's not the issue. It just seems there is some time before everything is ready to redraw - at least on with `edgeclk`.

> Also, I'd have thought Bangle.load should delete Bangle.CLOCK anyway really - because if you're loading something that's not a clock then you don't want that to stay set

I don't know how this work currently.

## Videos to illustrate

Current behavior:

https://github.com/user-attachments/assets/b1a6e25e-7d0d-4024-b06f-8b8a2dc52e83

With `Bangle.CLOCK=1` added to `showClock` (disregard the drawing bug of large sections of `edgeclk`):

https://github.com/user-attachments/assets/4e07fe53-1e70-40e5-9fcb-4e765f6b4dc2

thyttan:
Thinking some more, maybe instead the red back button should stay around until the `dtlaunch` screen is cleared and `edgeclk` is drawn. That would also mitigate the widgets hiding at different times I think. I get unsure of what the best approach is.

PR #2540, opened 2024-08-15
Update `locale` with support for meridians

atjn:
Mirroring changes in https://github.com/espruino/BangleApps/pull/3535


gfwilliams:
Ok, thanks! Shame it looks like this pushes the Bangle.js build just above the allowed filesize, but I'll find some other way to same some bytes

atjn:
Oh, I see. Sorry about that :/

PR #2548, opened 2024-08-28
allow to switch between USB_CDC and UART for ESP32C3 boards

MaBecker:
'DEFINES+=-DUSB_CDC', # Disable this line to use serial interface

Tested with both ESP32C3 boards. Please let me know if this is ok or you want me to change anything.

### DIFF
--- a/boards/ESP32C3_IDF4.py
+++ b/boards/ESP32C3_IDF4.py
@@ -74,6 +74,7 @@ info = {
      'DEFINES+=-DESP_STACK_SIZE=25000',
      'DEFINES+=-DJSVAR_MALLOC', # Allocate space for variables at jsvInit time
      'DEFINES+=-DUSE_FONT_6X8',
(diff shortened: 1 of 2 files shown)


gfwilliams:
Looks great - thanks!

SimonGAndrews:
hi @MaBecker , I'm curious how the uart knows what pins to use for tx and rx.   I can see in jshardwareUart.c , in the case where serial device is EVSERIAL1,   line 65 that inituart()  is being called with tx and rx gpio set to -1.   Is that the way the tx and rx are disabled ?   That may have been the desired state when using the usb. 

Screenshot from 2024-08-29 08-46-59

_(the screen shot shows this call and  the inituart() function on line 34. )_

gfwilliams:
Espruino tries to search for supported UART pins if none are specified, so it might come from: https://github.com/espruino/Espruino/blob/master/boards/ESP32C3_IDF4.py#L162-L163

PR #2549, opened 2024-08-29
Close comment in linker script

bobrippling:
Presumably this hasn't had an effect since we'd not see code booting up if the entry point was missing - a default might've been used.


gfwilliams:
interesting - thanks!

PR #2552, opened 2024-09-06
ignore esp-idf-4

MaBecker:


MaBecker:
I thought this should be in esp-idf a subfolder ......

gfwilliams:
Thanks! It's just that to put it in esp-idf/4/etc we'd have to change the build system for old IDFs, which I didn't want to do initially

PR #2564, opened 2024-10-06
E_showScroller_Q3: comment out prints to console

thyttan:
- [ ] Update minified code before merge (Whats the best way to do it?).


gfwilliams:
Wow, what a pain - sorry, not sure how those got left in there. I'll minify here - there's a script in `libs/js` for it which needs EspruinoDocs

PR #2565, opened 2024-10-08
E_showMenu_/E_showScroller_Q3: forward touch event

thyttan:
The goal was to make it possible for menus to distinguish between short
    and long touches when the entry maps to a function.

With this alarms in the `alarm` app could be made togglable by longpressing
    corresponding menu entries.


libs/js/banglejs/E_showMenu_Q3.js
```diff
@@ -38,11 +38,11 @@
           g.setColor(g.theme.fg2).setFont("12x20").setFontAlign(-1,0).drawString((item.format) ? item.format(v,1) : v, r.x+12, r.y+H/2);
           g.drawImage(/* 20x20 */atob(v==item.value?"FBSBAAH4AH/gHgeDgBww8MY/xmf+bH/jz/88//PP/zz/88f+Nn/mY/xjDww4AcHgeAf+AB+A":"FBSBAAH4AH/gHgeDgBwwAMYABmAAbAADwAA8AAPAADwAA8AANgAGYABjAAw4AcHgeAf+AB+A"), r.x+r.w-32, r.y+H/2-10);
         },
-        select : function(idx) {
+        select : function(idx, touch) {
           if (idx<0) return; // TITLE
           Bangle.buzz(20);
           item.value = item.min + idx*step;
-          if (item.onchange) item.onchange(item.value);
+          if (item.onchange) item.onchange(item.value, touch);
```
thyttan, line 45: I had overlooked to tweak this line in https://github.com/espruino/BangleApps/pull/3604

libs/banglejs/jswrap_bangle.c
```diff
@@ -5904,6 +5904,8 @@ On Bangle.js there are a few additions over the standard `graphical_menu`:
     menu is removed
   * (Bangle.js 2) `scroll : int` - an integer specifying how much the initial
     menu should be scrolled by
+  * `onchange : function(value, touch) {}` - on Bangle.js 2 this function can
```
gfwilliams, line 5907: I don't think this is in the right place - I think it means the `"":{...}` object here?

gfwilliams:
Thanks - looks good - I think it's just that comment?

libs/banglejs/jswrap_bangle.c
thyttan: Aha - I see what you mean.

libs/js/banglejs/E_showMenu_Q3.js
```diff
@@ -133,19 +133,19 @@
...
+      if ("function" == typeof item) item(l, touch);
       else if ("object" == typeof item) {
         // if a bool, just toggle it
         if ("number" == typeof item.value) {
           showSubMenu(item, keys[idx]);
         } else {
           if ("boolean"==typeof item.value)
             item.value=!item.value;
-          if (item.onchange) item.onchange(item.value);
+          if (item.onchange) item.onchange(item.value, touch);
```
thyttan, line 148: @gfwilliams should we also pass the scroller (`l`) here to mimic the capabilities of when the item is a function on L140?

It feels a little hacky/involved adding these things to `onchange`. Maybe there should be a separate function inside the `item` object named "func", "run" or similar that took the same role as if `item` had been a function?

thyttan: I'll do a small refactor and then ask you to comment when I pushed it to here.

thyttan: @gfwilliams 

For now I landed on not introducing a new extra function beside `onchange` inside entry objects. I believe that is better for backwards compatibility. 

(I go back and forth in my head on whether the extra function would be a good thing - please tell if you do like the idea)

libs/js/banglejs/E_showMenu_Q3.js
thyttan: removed this change. It didn't make sense when thinking it through.

libs/js/banglejs/E_showMenu_Q3.js
gfwilliams: Yes, I think it's best not to introduce more functions unless you absolutely need it too.

libs/js/banglejs/E_showMenu_Q3.js
```diff
@@ -133,19 +133,19 @@
...
       else if ("object" == typeof item) {
-        // if a bool, just toggle it
         if ("number" == typeof item.value) {
           showSubMenu(item, keys[idx]);
         } else {
+          // if a bool, just toggle it
           if ("boolean"==typeof item.value)
             item.value=!item.value;
-          if (item.onchange) item.onchange(item.value);
+          if (item.onchange) item.onchange(item.value, l, touch);
```
gfwilliams, line 148: I feel like including `l` probably isn't super important since it's returned by `showMenu` anyway, but if you think it should be included that's fine.

I'd just say in that case maybe you should include it in the other 2 calls to `item.onchange`? Sending `touch` for those might not be appropriate I guess as it won't be a touch on the main menu, but a different menu.

thyttan: > I feel like including `l` probably isn't super important since it's returned by `showMenu` anyway, but if you think it should be included that's fine.

Makes me wonder why `l` is passed on for `if ("function"==typeof item)` in the first place then?

> I'd just say in that case maybe you should include it in the other 2 calls to `item.onchange`? Sending `touch` for those might not be appropriate I guess as it won't be a touch on the main menu, but a different menu.

Hm.. Maybe that makes sense (and maybe `touch` as well). I get unsure of myself every time I think about this again and from different perspectives.

I've taken a step back for now. Might think/code on this later in the week.

gfwilliams: > why l is passed 

Yeah, I think the idea was the calling function would get the scroll level out of the menu easily (but it could have done that anyway). One bad side-effect is if you do `"menitem":showNewMenu` you get a memory leak as the old menu is passed as an argument, but I think realistically we can't easily remove it now unless it breaks something.

... so thinking about the memory leak I guess my gut feeling is to not pass `l`.

thyttan: I think there's no problem with removing `l` from being sent into the callback function, as I find in https://github.com/espruino/Espruino/issues/2567 .

thyttan:
Yes I think so to!

gfwilliams:
Thanks!

PR #2568, opened 2024-10-10
typescript: all boolean parameters can be `0` / `1`

bobrippling:
This permits the more common use of `Bangle.setBacklight(1)` or `Bangle.setLCDPower(1)`, etc


gfwilliams:
Thanks!

PR #2574, opened 2024-11-06
typescript: add menu scroller

bobrippling:
... for banglejs2 only, the easiest way to do this being to indicate it as optional


gfwilliams:
Thanks!

PR #2579, opened 2024-11-25
crypto: add support for AES CCM

ssievert42:
This adds support for AES CCM and enables it on Linux, Bangle.js 2 and the Bangle.js 2 emulator.

Can be used for creating encrypted BTHome advertisements.
Example that follows https://bthome.io/encryption to encrypt and decrypt a BTHome advertisement:
https://gist.github.com/ssievert42/dd979303b228cbd1c52733a3f63004cf

WIP BTHome module that creates encrypted advertisements (planning on submitting this soon™; check first comment below the gist for creating and setting an encryption key):
https://gist.github.com/ssievert42/1305e4a036d8a3d4a6e14229c2eb01e5


ssievert42:
Should this be enabled on Puck.js?
It does use up some precious flash, but might be handy for someone wanting to use BTHome with authenticated encryption.

gfwilliams:
Thanks - just looking at this, I'm not quite sure what the differences are but is there any way code can be reused from jswrap_crypto_AEScrypt so we're not having quite so much duplication?

Or at the very least, doing like we do in jswrap_crypto_AEScrypt where we have one function we call from jswrap_crypto_AES_encrypt and jswrap_crypto_AES_decrypt, as it looks like there's a bunch of duplicated code in jswrap_crypto_AES_ccmDecrypt and jswrap_crypto_AES_ccmEncrypt?

ssievert42:
Turns out the Bangle.js 2 build can be 1424 B smaller (still an increase compared to current master though) by not configuring mbedtls to enable all AES stuff if the board doesn't use AES.

gfwilliams:
Thanks for that! Combining the two functions still helps a lot.

> not configuring mbedtls to enable all AES stuff if the board doesn't use AES.

Great! Is that done now?

Some description in the JSON header for ccmEncrypt/Decrypt would be really handy. I know we don't seem to have anything for AES but ideally a small comment that it's for CCM encryption and a very small bit of example code would be perfect.

So what's your main reason for doing this? BTHome?

Before merging, have you tested and got it working with BTHome? It'd be good to ensure we've got everything you need for it to be useful.

ssievert42:
> Great! Is that done now?

Should be as of my [last push](https://github.com/espruino/Espruino/compare/3c98dd4ad5a971d3e0fc9d24f313b6384eba96c9..5423dd295cce73d5376e4352daf7057df8296eda) :+1:

> Some description in the JSON header for ccmEncrypt/Decrypt would be really handy

Oops, totally forgot about that :see_no_evil: I'll try to come up with something.

> So what's your main reason for doing this? BTHome?

Yup: Sending encrypted BTHome advertisements in the general direction of Home Assistant.
But apart from BTHome I (currently) don't have another use case for this in mind.

I discovered BTHome like four days ago, but didn't want to use something wireless without some form of security - hence this crypto stuff.
I've actually been trying to use my Bangle.js 2 to talk to Home Assistant for some time now. Ideally bidirectional and without needing a phone, but something that also works while connected to a phone.
There's even a (mostly failed) attempt at using ZigBee somewhere on my computer. I just kind of gave up when I couldn't get multiprotocol to work properly - Home Assistant and a Seeed XIAO BLE, that I turned into a ZigBee sniffer, could see another Seeed XIAO BLE running Espruino, but for the most part not the other way around. Probably because the ZigBee stack didn't get enough radio time to listen for data.

By the way, would you be open to adding the Seeed XIAO BLE to the list of supported boards?
(If I were to create a PR.)
I've been using these [with my own board definition and some changes to directly build an uf2](https://github.com/ssievert42/Espruino/tree/xiaoble) (based on one found [here](https://forum.espruino.com/conversations/370630/)).
For me, personally, these are currently the best way to test stuff on real hardware, because flashing is super quick and requires no software-setup on the host - just double-press the reset button, copy the uf2 to the usb-drive that appears and you're done!
(A way to get to the bootloader from software would be nice though. Since ruining and having to replace the button on my Bangle.js 2, I've got slight paranoia about wrecking tiny buttons.)

> have you tested and got it working with BTHome?

Just tested this again; sending encrypted BTHome advertisements and having Home Assistant do something seems to work nicely :+1:
To create advertisements I'm using [this modified version of the BTHome module](https://gist.github.com/ssievert42/1305e4a036d8a3d4a6e14229c2eb01e5) (also mentioned in the first comment here).
I've also written a [script](https://gist.github.com/ssievert42/dd979303b228cbd1c52733a3f63004cf) that follows https://bthome.io/encryption to test if encrypting and decrypting data works as expected. (Uses the same data as in the BTHome docs.)
What I didn't try is receiving and decrypting BTHome advertisements, but since decrypting works that should also work.

gfwilliams:
Sorry for the delay, I missed this...

> Is there a nice built-in way to fill an ArrayBuffer with data from a hex-String?

Actually no (just the usual JS)... for base64 there's `arraybuffer.set(atob(...))`. Not sure if there's anything in Node.js/browser that we could implement here? There is https://nodejs.org/api/buffer.html#static-method-bufferfromstring-encoding which might make sense, but I feel like it might end up being too big to keep in for all boards if all the other functionality started getting added.

> I discovered BTHome like four days ago, but didn't want to use something wireless without some form of security

My feeling with it was if someone can get close enough to read the BLE advertisement with my Office's temperature, they're probably close enough to look in the window anyway :)

Still, it's nice to have the option...

> By the way, would you be open to adding the Seeed XIAO BLE to the list of supported boards?

Yes, that'd be good, thanks!

ssievert42:
> Sorry for the delay, I missed this...

No worries :)

I'm still gonna create a PR soon-ish with some docs for this, probably with something along the lines of `let key = new Uint8Array([0xde,0xad,0xbe,0xef])` for initializing ArrayBuffers.

> if someone can get close enough to read the BLE advertisement with my Office's temperature, they're probably close enough to look in the window anyway :)

I hadn't thought of it that way :see_no_evil:
My usecase is more on the controlling stuff side; I guess somebody living in the apartment next to me could potentially toggle my lights or something.
But yeah, having the option is still nice, especially for someone as paranoid as I am about sending unencrypted data over the air.

gfwilliams:
> I guess somebody living in the apartment next to me could potentially toggle my lights or something.

Yeah, I live out in a village so we're spaced out enough that Bluetooth won't get far. Even if it did I'm not too worried about hacking attempts from our 90 year old neighbour.

It's easy to forget that if you're in a flat or more built up area the signal can be viewed by a bunch of people you may not know!

PR #2581, opened 2024-12-03
ESP32: add setIP and setAPIP

MaBecker:
similar to what I did for ESP8266


libs/network/esp32/jswrap_esp32_network.c
```diff
@@ -92,6 +92,8 @@ static bool g_isStaConnected = false;
 #define EXPECT_CB_EXCEPTION(jsCB)   jsExceptionHere(JSET_ERROR, "Expecting callback function but got %v", jsCB)
 #define EXPECT_OPT_EXCEPTION(jsOPT) jsExceptionHere(JSET_ERROR, "Expecting Object, got %t", jsOPT)
 
+// Global data structure for setIP and setAPIP
+tcpip_adapter_ip_info_t info;
```
gfwilliams, line 96: Please could you move this into the setIP function? Looking at tcpip_adapter_set_ip_info I'm pretty sure it doesn't have to be a global variable so having it just in the fn will free up a bit of RAM

gfwilliams:
Looks really good, thanks! Please can you just move the variable mentioned above and I'll get this merged?

MaBecker:
Ups, yes it is no longer needed - thanks for checking!

gfwilliams:
Great, thanks!

PR #2584, opened 2024-12-09
BLE: always return privacy info from NRF.getSecurityStatus()

ssievert42:
This fixes `NRF.getSecurityStatus()` only returning privacy info while connected and makes it return privacy info regardless of connection state.

I somehow missed this when adding BLE private address support.


ssievert42:
Just rebased onto master.

gfwilliams:
Thanks! Sorry I missed merging this earlier.

Issue #2578, opened 2024-11-25
regression: calling builtin js libs fails with "Uncaught SyntaxError" since b8634fd8e0040c09946e5b6e0eb0bf4b6587acec


ssievert42:
On both Bangle.js 2 and the Bangle.js 2 emulator, calling builtin js libs like `Bangle.setUI()` doesn't work anymore.

The following code:
```js
Bangle.setUI();
```
only results in:
```
Uncaught SyntaxError: Got '*' expected EOF
 at line 1 col 1
*k(){1{Bangle.buzz(30)}catch(b){}},a={};#(object
^
in function called from line 1 col 14
Bangle.setUI();
```

but at least Bangle.loadWidgets()` is affected as well.

I bisected the issue, and it looks like  b8634fd8e0040c09946e5b6e0eb0bf4b6587acec is the first bad commit.

No idea how to fix that though :sweat_smile:

To reproduce, try a firmware / the emulator built from b8634fd8e0040c09946e5b6e0eb0bf4b6587acec that doesn't work, and compare it with the firmware / emulator from 617e1a29753cd4dd5a103f528d3bbfbcc22fbf32 which should be working.

gfwilliams:
You're sure you've tested on an actual device? Because I've been using firmwares with this change in  on a device on my wrist for ages now without issues.

Just tried again with `Bangle.setUI` on 2v24.183 (downloaded as a 'cutting edge' build from https://www.espruino.com/binaries/travis/master/) and there are no errors - it seems to 'just work'.

Are these builds you're doing yourself? What platform are you building on? The minified JS files in `libs/js` now include non-ASCII characters, so I wonder whether git has checked them out in the wrong format?

I've had of issues with Git changing line endings between windows/unix, but not other things... But it might not be a huge surprise if a platform is trying to interpret the binary file as UTF8 for some reason.

Maybe check the MD5sum of the file? this is what it is at current git head 1aed93498f00565a45a5786932a8f0c60eddda4b

```
md5sum libs/js/banglejs/Bangle_setUI_Q3.min.js 
2be6d6d2bba8f9559327b212e7a47f14  libs/js/banglejs/Bangle_setUI_Q3.min.js
```

ssievert42:
> You're sure you've tested on an actual device? Because I've been using firmwares with this change in on a device on my wrist for ages now without issues.

Yup :+1: I've had to use dfu to reflash my watch after building from current master, because it would only show that error and I couldn't get into the recovery menu.
[...]

PR #2591, opened 2025-01-05
Add board: Seeed XIAO BLE

ssievert42:
This adds support for the [Seeed XIAO BLE](https://www.seeedstudio.com/Seeed-XIAO-BLE-nRF52840-p-5201.html), a tiny development board with a nRF52840 and 2 MB of external flash.

The board definition file is based on [this one by Espruino forums user parasquid](https://forum.espruino.com/comments/17036990/).

Some goodies that are included as well:
- the build generates an UF2 image, that can be flashed to the board when it is in bootloader mode, without needing to install anything on the host (connect via USB, double press the reset button, then copy the file to the USB drive that appears)
- automagically built UF2 image with GitHub actions
- `E.rebootToDFU()` triggers a reboot to UF2 bootloader mode

### DIFF
--- a/boards/XIAOBLE.py
+++ b/boards/XIAOBLE.py
@@ -55,6 +55,7 @@ info = {
             # 'DEFINES += -DPIN_NAMES_DIRECT=1', # Package skips out some pins, so we can't assume each port starts from 0
             "LDFLAGS += -Xlinker --defsym=LD_APP_RAM_BASE=0x2ec0",  # set RAM base to match MTU
             'DEFINES += -DESPR_BLE_PRIVATE_ADDRESS_SUPPORT',
+            'DEFINES += -DESPR_HAS_BOOTLOADER_UF2',
             'NRF_SDK15=1',
         ],
     },
--- a/boards/XIAOBLE.py
+++ b/boards/XIAOBLE.py
@@ -24,6 +24,11 @@ make clean && BOARD=XIAOBLE RELEASE=1 make
 # The board should automatically disconnect after copying is finished, and reboot into Espruino,
 # which turns on the red led for a short time after starting up.
 
+# If you accidentally put some stuff in .boot0 that prevents you from interacting with Espruino:
+# Pull pin D1 to 3.3 V / high (important: make sure you use 3.3 V; higher voltages may cause damage) and reset the board,
+# by either pressing the reset button or cutting the power and powering it up again.
+# This behaviour is archived by configuring pin D1 as BTN1.
+
 import pinutils
 info = {
     "name": "Seeed Xiao BLE",
@@ -87,6 +92,7 @@ devices = {
     "LED1": {"pin": "D11"},
     "LED2": {"pin": "D13"},
     "LED3": {"pin": "D12"},
+    "BTN1": {"pin": "D1", "pinstate" : "IN_PULLDOWN"},
     "BAT": {
         "pin_charging": "D23",
         "pin_voltage": "D32",
--- a/boards/XIAOBLE.py
+++ b/boards/XIAOBLE.py
@@ -1,0 +1,170 @@
+#!/bin/false
+# This file is part of Espruino, a JavaScript interpreter for Microcontrollers
+#
+# Copyright (C) 2013 Gordon Williams <gw@pur3.co.uk>
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# ----------------------------------------------------------------------------------------
+# This file contains information for a specific board - the available pins, and where LEDs,
+# Buttons, and other in-built peripherals are. It is used to build documentation as well
+# as various source and header files for Espruino.
+# ----------------------------------------------------------------------------------------
+
+# To build and flash run:
+"""
+source ./scripts/provision.sh XIAOBLE
+make clean && BOARD=XIAOBLE RELEASE=1 make
+"""
+# Then connect the board to your computer and press the reset button twice.
+# You should now see a USB storage device called "XIAO-SENSE" (no matter which version you have).
+# Copy the "espruino_*_xiaoble.uf2" file from the "bin" folder to that drive.
+# The board should automatically disconnect after copying is finished, and reboot into Espruino,
+# which turns on the red led for a short time after starting up.
+
+import pinutils
+info = {
+    "name": "Seeed Xiao BLE",
+    "link": ["https://www.seeedstudio.com/Seeed-XIAO-BLE-nRF52840-p-5201.html"],
+    "default_console": "EV_USBSERIAL",
+    "variables": 14000,  # How many variables are allocated for Espruino to use. RAM will be overflowed if this number is too high and code won't compile.
+    "binary_name": "espruino_%v_xiaoble.uf2",
+    "build": {
+        "optimizeflags": "-Os",
+        "libraries": [
+            "BLUETOOTH",
+            # "GRAPHICS",
+            "CRYPTO","SHA256","SHA512",
+            "AES_CCM",
+            # "NFC",
+            # "TENSORFLOW",
+            "NEOPIXEL",
+            "JIT",
+        ],
+        "makefile": [
+            "DEFINES += -DESPR_LSE_ENABLE",  # Ensure low speed external osc enabled
+            "DEFINES += -DCONFIG_GPIO_AS_PINRESET",  # Allow the reset pin to work
+            "DEFINES += -DNRF_USB=1 -DUSB",
+            "DEFINES += -DNEOPIXEL_SCK_PIN=33 -DNEOPIXEL_LRCK_PIN=34",  # nRF52840 needs LRCK pin defined for neopixel
+            "DEFINES += -DBLUETOOTH_NAME_PREFIX='\"XIAOBLE\"'",
+            "DEFINES += -DSPIFLASH_READ2X",  # Read SPI flash at 2x speed using MISO and MOSI for IO
+            "DEFINES += -DESPR_UNICODE_SUPPORT=1",
+            "DEFINES += -DNRF_SDH_BLE_GATT_MAX_MTU_SIZE=131",  # 23+x*27 rule as per https://devzone.nordicsemi.com/f/nordic-q-a/44825/ios-mtu-size-why-only-185-bytes
+            # 'DEFINES += -DPIN_NAMES_DIRECT=1', # Package skips out some pins, so we can't assume each port starts from 0
+            "LDFLAGS += -Xlinker --defsym=LD_APP_RAM_BASE=0x2ec0",  # set RAM base to match MTU
+            'DEFINES += -DESPR_BLE_PRIVATE_ADDRESS_SUPPORT',
+            'NRF_SDK15=1',
+        ],
+    },
+}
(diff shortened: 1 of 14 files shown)


gfwilliams:
Thanks - this looks great!

Do you think it'd be possible to change  `jswrap_espruino_rebootToUF2` just to use `jswrap_espruino_rebootToDFU` though? There's already the stuff in there for it so all you'd have to do is rename `jshRebootToUF2Bootloader` to `jshRebootToDFU` in `jshardware.c` and change the `"ifdef" : "STM32F4",` line to `"#if" : "defined(STM32F4) || defined(ESPR_HAS_BOOTLOADER_UF2)",`

ssievert42:
Thanks for the feedback!
I've now dropped `E.rebootToUF2()` in favor of `E.rebootToDFU()`, mentioned the UF2 bootloader mode in the docs of `E.rebootToDFU()` and changed the other stuff as you suggested - no need to have two functions with slightly different names that basically do the same thing :)

ssievert42:
Does the board definition file need to use `pinutils.generate_pins()`?
I don't remember why I decided to change `get_pins()` to not use `pinutils.generate_pins()`, but I think the reason was to have pin names that are the same as the ones printed on the board.

boards/XIAOBLE.py
```diff
@@ -0,0 +1,172 @@
...
+# The board should automatically disconnect after copying is finished, and reboot into espruino,
+# which turns on the red led for a short time after starting up.
+
+import pinutils
+info = {
+    "name": "Seeed Xiao BLE",
+    "link": ["https://www.seeedstudio.com/Seeed-XIAO-BLE-nRF52840-p-5201.html"],
+    "default_console": "EV_USBSERIAL",
+    "variables": 14000,  # How many variables are allocated for Espruino to use. RAM will be overflowed if this number is too high and code won't compile.
+    "binary_name": "espruino_%v_xiaoble.hex",
```
gfwilliams, line 33: What happens if this is .uf2? It may be you don't need the `ifdef XIAOBLE ... proj: $(PROJ_NAME).uf2` line if it is...

targets/nrf5x/jshardware.c
```diff
@@ -2887,6 +2887,14 @@ void jshReboot() {
   NVIC_SystemReset();
 }
 
+#ifdef ESPR_HAS_BOOTLOADER_UF2
```
gfwilliams, line 2890: This looks great, thanks! Please could you add a comment in https://github.com/espruino/Espruino/blob/master/README_BuildProcess.md#infomakefile-definitions though?  I'm trying to keep all the defines documented somewhere

src/jsflash.c
```diff
@@ -1374,6 +1374,7 @@ JsVar *jsfGetBootCodeFromFlash(bool isReset) {
 
 bool jsfLoadBootCodeFromFlash(bool isReset) {
   // Load code in .bootFirst at first boot UNLESS BTN1 IS HELD DOWN (BTN3 for Dickens)
+#ifndef ESPR_NO_BOOT_JS
```
gfwilliams, line 1377: It's a shame the board doesn't have a button so this is needed - but I feel like it's very unlikely anyone would make a custom build that includes this?

Maybe you could:

* Make the UF2 include the flash area used for saved code so it always overwrites it
* Make a special UF2 that *just* overwrites the saved code area to allow you to clear it if you want to?

gfwilliams:
Thanks for those changes to UF2 reboot - that looks really good.

> Does the board definition file need to use `pinutils.generate_pins()`?  ... I think the reason was to have pin names that are the same as the ones printed on the board.

No, that's fine - generate_pins was just a utility function to generate a bunch of contiguous pins. If you want to renumber the pins so they match the board what you've done is the right way to handle it.

boards/XIAOBLE.py
ssievert42: Hardcoding only the XIAOBLE to create a UF2 in that makefile does seem rather hacky...
Turns out I needed to change a bit more; something kept adding ".hex" to the end of the binary name.
See d40b0ca6829fd5d614553debfd23dcb8861d94f5 for what I did to make this work :)

targets/nrf5x/jshardware.c
ssievert42: Added :+1:
Reminder for future me: Do this for BLE private address support as well.

src/jsflash.c
ssievert42: > but I feel like it's very unlikely anyone would make a custom build that includes this

Yeah, you're probably right.
The board does have a (reset) button; pressing it just immediately resets the board...

> Make the UF2 include the flash area used for saved code

Probably not an option if saved code is on SPI flash though :see_no_evil:
(And I really want to use those 2 MB of external flash.)

> Make a special UF2 that just overwrites the saved code area

I'd rather have something that allows recovering other data saved to storage, but I guess this will be the way to go if you don't like the "pull a magic pin to ground on boot" thing I added with my latest push.

ssievert42:
Thanks for the review!
I think I've addressed most of your comments;
The changes I just pushed do the following:
- if the "binary_name" in a board definition file ends with ".uf2", we build an UF2 image and pass `chip.part` (in the case of the XIAOBLE `NRF52840`) to the `uf2conf.py` script as the "familyID" argument
- drop `ESPR_NO_BOOT_JS` from my previous stuff in this PR
- add `ESPR_SKIP_BOOT_JS_PIN` to set a pin that can be pulled to ground on boot to skip running boot js (seems to work reasonably well, even without adding a delay between setting the pin mode and reading the pin value)

If you think that the whole skipping boot js stuff is not that good of an idea I'll just drop that for now and look into the other recovery option you suggested, where we'd have an "I messed up and want to reset my board" build.

gfwilliams:
Thanks! This is looking really good. 

> ESPR_SKIP_BOOT_JS_PIN 

As D1 is configured as the button in the bootloader anyway, could we not just add `BTN1' : { 'pin' : 'D1' , 'inverted' : 1}` to the BOARD.py file, and this would have exactly the same effect, without having to have ESPR_SKIP_BOOT_JS_PIN? There are some other things we do (like clearing bonded peers) if booted with BTN1 held down so it might be preferable to do that route?

ssievert42:
Minimal code changes to achieve the same thing, sounds neat!
However, for BTN1 to have any effect on whether boot js is run or not, I had to add a `#if defined(XIAOBLE)` to `jsflash.c`. (See the latest commit that I pushed.)

Do you think having D1 / BTN1 as IN_PULLDOWN or IN_PULLUP would be better?
I figured that it's probably IN_PULLDOWN, as having (weak) ground on a pin can probably cause less weird behavior from something connected to that pin than pulling that pin high.

jj12:
Thank you both for your work 😄

When trying the firmware from the build artifacts, I noticed that some pins are not working. D6 to D10 have a wrong `num` assigned: instead of 11-15 must have values 43-47, as per [variant.cpp](https://github.com/Seeed-Studio/Adafruit_nRF52_Arduino/blob/master/variants/Seeed_XIAO_nRF52840/variant.cpp).

Also: The the pin definitions >= D12 have different pin numbers assigned than mentioned in variant.cpp or the [pinout sheet](https://files.seeedstudio.com/wiki/XIAO-BLE/XIAO-nRF52840-pinout_sheet.xlsx). This might be mostly cosmetic, but poses an inconsistency with Seeed's docs.

(Unfortunately, I'm not yet able to get the build process running myself, via dockerfile. Compiling any board fails with `Syntax error: Unterminated quoted string`, will have to keep looking.)

ssievert42:
@jj12 Thanks, nice catch!
The pin definitions should now match Seeed's docs; I was going purely by the [shematic](https://files.seeedstudio.com/wiki/XIAO-BLE/Seeed-Studio-XIAO-nRF52840-Sense-v1.1.pdf) and apparently can't really read those (and assumed that P0.N or P1.N both meant pin N) and just assigned pin numbers to the leftover pins at will  :see_no_evil:
(I've checked pins D0 through D10 with a multimeter and `digitalWrite(D0, 1)` just to be sure though, and apart from D1, where you've got to do `D1.mode("output")` first since it's defined as BTN1, everything seems fine :) )

About your build issues: Just that message isn't a lot to go by, but can you build from `master`, or does this PR introduce those errors?
I've had to deal with some [weird build behaviour](https://github.com/espruino/Espruino/issues/2578) myself, maybe you can find something that helps in that issue.
Just a stab in the dark: maybe it's an issue with Git changing line endings?

jj12:
Yup, can confirm, all pins work now. Thanks!

Building is not related to your PR, it was based on `master` and the provided Dockerfile on RPi4. Also now tried your steps from the other issue, but didn't work (new error this time...). Might try on x86, and if that does not work, I will open a separate issue.

gfwilliams:
> Do you think having D1 / BTN1 as IN_PULLDOWN or IN_PULLUP would be better?

I think pulldown is probably best as well - otherwise we'd have to get it to 'invert' the state in software as well.

> adding ifdef to jsflash

Have you tried without those IFDEFs? It might have been a bit of a red herring - because normally we detect the button press at boot through:

https://github.com/espruino/Espruino/blob/master/targets/nrf5x/main.c#L34
https://github.com/espruino/Espruino/blob/master/src/jsinteractive.c#L889-L897

So I think if you wrote code to the device and then booted it while holding the button, it wouldn't be executed even without those ifdefs in jsflash.

We also have this idea of `.bootrst` and `.bootcde` (https://www.espruino.com/Saving) - sometimes folks want to force code to always run even when the button is pressed (eg if in a product you don't want a customer to be able to boot up with nothing running), but we make it nontrivial to write to  `.bootrst` in the IDE so it's ok if it can make it hard to recover from (it's the same in any board).

I think the ifdefs for BANGLEJS are in there because in that case we wanted a way to boot into the recovery menu and ensure that absolutely nothing ever ran - but that's kind of an exception because we have an actual UI that can be displayed for recovery

> Docker

Yes, the Docker build is a bit annoying and honestly I'm tempted to just remove that file. It was contributed but then not maintained, and given Github builds are just so easy, I think it's probably just wasting people's time having something that's not working in the repo.

gfwilliams:
> `.boot0`

I think this is exactly as expected. It's a bit like what I said with `.bootrst` but the process is all documented at: https://www.espruino.com/Saving#boot-process

When you save to flash, the IDE saves to `.bootcde` which really is only executed when the button isn't held down, so you're safe.

If you write to `.boot0` deliberately then yes, that is always executed - but there are reasons some people want that, and you'd have to deliberately choose to write to it instead of the IDE's defaults.

So I'll take out those extra IFDEFs so this port behaves like all the others and merge this in - you could maintain your own version with this change but I'd argue it's probably  a lot easier just to write to `.bootcde` :)

> Docker

Thanks for spotting about the compiler download. Actually if arm-gcc exists on the path then provision won't try and download it, so probably the easiest is just to ensure that `apt install gcc-arm-none-eabi` comes before provision. Having said that different GCC versions have staggeringly different binary sizes and performance, which is why we specifically choose one - so even if you *can* build using docker, you may find that building for some devices (where storage space is very tight) still fails.

jj12:
> Docker

@ssievert42 Thanks a lot for the detailed information on your Dockerfile and steps, I was able to create my own build.
The firmware is indeed bigger than the github artifact, for the NRF52840 no issue, but for the NRF52832 (MDBT42Q) it results in code and storage overlap.

> Some goodies

Finally also tried `E.rebootToDFU()`, this is so much nicer than trying to hit the mini reset button!

PR #2593, opened 2025-01-11
sdl: Add support for input

pavelmachek:
I find SDL support in Espruino useful. And it would be even more useful if it has input ... so I did that. I'm now using it for developing applications for Bangle.js2 without hardware and without emulator, and it is rather useful.

Please review. There's probably better way to do this, but I'll need pointers.

### DIFF
--- a/bench.js
+++ b/bench.js
@@ -1,0 +1,128 @@
+#!bin/espruino
+
+eval(require("fs").readFile("sdl.js"));
+
+R = Bangle.appRect;
+
+function introScreen() {
+  g.reset().clearRect(R);
+  g.setColor(0,0,0).setFont("Vector",25);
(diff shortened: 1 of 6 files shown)


pavelmachek:
Yeah, so ... peek and poke are "traditional" methods of accessing hardware, right? :-). I then used javascript to glue it all together.

Thanks for pointers, let me take a look.

pavelmachek:
Thanks for pointers, I have something with better interface. I'll likely close this and open new merge request once ready.

gfwilliams:
Great, thanks! If you can keep the `E.touch` event in the same format I'm hoping that eventually we can build in `E.showMenu/etc` that will work with it - which would be really handy for boards like CYD

PR #2596, opened 2025-01-15
sdl: Add support for input

pavelmachek:
I find SDL support in Espruino useful. And it would be even more useful if it has input ... so I did that. I'm now using it for developing applications for Bangle.js2 without hardware and without emulator, and it is rather useful.

It now generates "touch" event on E, so should be compatible with other hardware.

### DIFF
--- a/libs/graphics/lcd_sdl.c
+++ b/libs/graphics/lcd_sdl.c
@@ -91,6 +91,7 @@ void lcdInit_SDL(JsGraphics *gfx) {
     SDL_Quit();
     exit(1);
   }
+  SDL_WM_SetCaption("Espruino", NULL);
 }
 
 void lcdIdle_SDL() {
--- a/libs/graphics/lcd_sdl.c
+++ b/libs/graphics/lcd_sdl.c
@@ -97,9 +97,10 @@ void lcdInit_SDL(JsGraphics *gfx) {
 }
 
 void lcdIdle_SDL() {
+  static bool down;
+  extern void nativeQuit();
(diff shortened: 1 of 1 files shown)


gfwilliams:
Looks great, thanks! Please can you just remove the `fflush(stdout);` line and I'll merge this?

pavelmachek:
Sorry, debugging leftover. It should be fixed now.

I would like to handle "window close" event from sdl. Can I just exit(0) from C or should I do something more complex. Would it make sense to add keyboard handling?

gfwilliams:
> Can I just exit(0) from C or should I do something more complex.

Since this is for Linux only you could do:

```
void nativeQuit();
nativeQuit();
```

Which basically causes it to quit loopin and exit nicely: https://github.com/espruino/Espruino/blob/master/targets/linux/main.c#L564-L566

> Would it make sense to add keyboard handling?

You could add `jshPushIOCharEvent(EV_USBSERIAL, (char)ch);` for each typed key - so that it goes to the console as if the terminal window had focus?

I don't think going further really helps as it's not something exposed on other platforms - so it wouldn't be any good for emulating/etc

pavelmachek:
Ok, thanks for the hints. I added quit handling as suggested.

I guess translating keyboard between sdl and serial would not be trivial, lets postpone that for now. (SDL sees left arrow but serial expects escape sequence... etc.)

gfwilliams:
Ok, thanks! This all looks good then - just merging!

Issue #2589, opened 2025-01-03
ESP32 IDF4 Build - I2C.Setup function results in Invalid Arguments on clock speed


SimonGAndrews:
An error occurs using espruino in the ESP32-C3 IDF 4 build. when setting up the i2C.   
![GetImage (2)](https://github.com/user-attachments/assets/a82b0c60-335c-4d27-a1e5-c2e9dc646547)
In more detail:  When building with the command line: 
`BOARD=ESP32C3_IDF4 DEBUG=1 make flash  `

The error occurs when executing the espruino command: 
`I2C1.setup({scl:D9,sda:D8}); `

 The error text is : 
```
i2c: i2c_param_config(684): i2c clock choice is invalid, please check flag and frequency 
Uncaught Error: jshI2CSetup: Invalid arguments at line 1 col 27 
I2C1.setup({scl:D9,sda:D8}); 
```

SimonGAndrews:
This [link](https://esp32.com/viewtopic.php?t=21030) seems to identify the problem. 
And tracing the error message back into the code shows that the ESP IDF i2c driver (home/Espruino/esp-idf-4/esp-idf/components/driver/i2c.c)  function i2c_get_clk_src() , inside i2c_param_config() is expecting to be passed . clk_flags in the config structure from espruino jshardwareI2c.c  
![Screenshot from 2025-01-03 11-19-40](https://github.com/user-attachments/assets/aeeea1a8-129c-45f6-b4ed-a55c2f3911c3)

i2c_get_clk_src() proides [clock selection functionality](https://docs.espressif.com/projects/esp-idf/en/v4.4/esp32c3/api-reference/peripherals/i2c.html#source-clock-configuration) in the IDF V4.4 for the specific device..

SimonGAndrews:
testing is WIP

gfwilliams:
Thanks! This all looks good - if you could get your changes as a PR (or just a patch I can apply) that'd be great.

On the I2C pin front, on STM32 where we have a similar situation of hard-coded pins, we have a checkPinsForDevice function which calls `jshFindPinForFunction(device, functions[i]);` which works out what pin to used based on the auto-generated `jspininfo.c` file: https://github.com/espruino/Espruino/blob/master/targets/stm32/jshardware.c#L2532-L2535

So I think you could do:

```
if (!jshIsPinValid(info->pinSCL)) info->pinSCL = jshFindPinForFunction(device, JSH_I2C_SCL);
[...]

PR #2600, opened 2025-01-19
Fix for issue #2589  - i2Csetup for ESP 32-C3

SimonGAndrews:
See comments in the issue #2589 regarding the details of the change and the testing completed. 
Note 
- the ESP32-S3 build was not tested but hopefully and should  be ok. 
- the element of this change, which reverses default pins for SDA & SCL for the ESP32 build (as discussed in the issue) will be a breaking change for anyone depending upon the I2C1.setup() function without specifying SDA and SCL in a config for the an ESP32 setup.   However the ESP32 build will now be inline with common documentation.


gfwilliams:
This looks excellent - thank you! I think it's very unlikely folks will be using I2C.setup without specifying pins so I'm pretty sure we're good.

Issue #2601, opened 2025-01-23
ESP32 - S3 and C3 build (ESP IDF 4) - Error Initialising SPI1


SimonGAndrews:
See [Forum](https://forum.espruino.com/conversations/402116/) Where issues identified by @devsnd result in a restart of the ESP32 - C3 when  
`SPI1.setup({ sck:D4, mosi:D5 });` 
is executed in a script.   
The issue does not occur when this code is used:
`var spi = new SPI();   
spi.setup({ sck:D4, mosi:D5 });
`

The restart is preceeded with the message:

![Image](https://github.com/user-attachments/assets/161f1e26-b866-443c-998b-df636f5ca030)

spi_bus_initialize is called in Espruino from targets/esp32/jshardwareSpi.c  
Investigation to follow.

MaBecker:
My guess: This could be some hard coded  dma channel or an sdconfig issue.

SimonGAndrews:
Thanks, yep will check out  the use of SPI_DMA_CH_AUTO
And look at the pins.  The i2c2 pins need defining also.  I’m not as familiar with SPI so doing a little home work this am.   Should be able to sort.

gfwilliams:
> I’m not as familiar with SPI so doing a little home work this am

Thanks! If you want a quick test, connect MOSI and MISO pins together, and you should then be able to do `SPI.send("Hello")` and get the same characters back!

SimonGAndrews:
Nice Tip.  Thanks.
Just to confirm this issue also occurs in the ESP32C3_IDF4  (ESP32-C3 target) build so will fix for both targets.

SimonGAndrews:
Using  SPI_DMA_CH_AUTO seems to be the answer.  Testing under way including making sure ESP32 still works.

SimonGAndrews:
The Change to fix the issue was to use the SPI_DMA_CH_AUTO flag in the DMA parameter in the ESPIDF spi_bus_initialize () function in  file /targets/ESP32/jshardwareSpi.c
This file also used hardcoded SPI pin and SPI Peripherial Host IDS which were limited to the ESP32 and did not include those for the ESP32C3 and ESP32S3.   Changes to accomodate the additional boards and the hardcoding were made by using Pin configurations in the relevent Boards.py file for the targets.

SimonGAndrews:
The board files ESP32.Py , ESP32C3_IDF4.py and ESP32S3_IDF4 were modified as in this example for the S3.

[...]

PR #2603, opened 2025-01-28
ESP32 (IDF4) Fix for SPI Setup errors issue #2601

SimonGAndrews:
See notes on the change and the testing carried out in #2601


MaBecker:
Can you please edit the other ESP32 board files too?

gfwilliams:
Wow, thanks for all your work on this - it looks great! I'll copy the changes to other boards at this end after merging

gfwilliams:
Just done - I did notice that the naming of the SPI SCLK decl was wrong so it might not have auto-assigned - the naming of it as SPIx_SCK is annoying and not like everywhere else (where we use `SCLK`), but it's because it came from the STM32 csv files that were scraped from STM32 datasheets early on!

PR #2605, opened 2025-01-29
build: upload .uf2 image to espruino.com

ssievert42:
This adds `*.uf2` to the list of files the `ci_upload.sh` script tries to upload to espruino.com.  
And also changes the address from http to https (which ideally shouldn't break anything).

Reason for doing this: I'd like to have the XiaoBLE UF2 image show up [here](https://www.espruino.com/binaries/travis/master/) instead of the `.hex` file.

I don't really see any value in having both the `.hex` file and `.uf2` file, so ideally the `.hex` file wouldn't show up in GitHub CI builds or on espruino.com, but I've currently got no idea how to cleanly do that :see_no_evil:

This probably won't work without adjusting that `travis_upload.php` though...

### DIFF
--- a/scripts/ci_upload.sh
+++ b/scripts/ci_upload.sh
@@ -22,5 +22,5 @@ fi
 
 if [[ -n \"$UPLOADTOKEN\" ]]; then 
   cd bin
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks! Yes, I'll check and make sure.

PR #2608, opened 2025-02-02
ESP32 - fix jshCanWatch for ESP32 & ESP32S3

MaBecker:
setWatch on ESP32 and ESP32S3 retruns `WARNING: Unable to set watch. You may already have a watch on a pin with the same number (eg. A0 and B0),
or this pin cannot be used with watch` for vaild pins,


gfwilliams:
Thanks! And this definitely seems to fix it for you?

This is an odd one as looking back this change was made 8 years ago! I'm amazed nobody noticed this earlier - you must have used setWatch in the past on ESP32 surely and it worked?

gfwilliams:
Reverted. Turns out it didn't actually work.

PR #2620, opened 2025-03-27
Improve timer scheduling accuracy

ssievert42:
This is an attempt to improve timer scheduling accuracy, especially for repeating tasks.

Previously, a repeating task would be scheduled like this:
- schedule for wakeup in task interval time
- on wakeup, do some stuff
- schedule for wakeup in task interval time
- on wakeup, do some stuff
- schedule for wakeup in task interval time
- ...

Without accounting for doing some stuff on wakeup, scheduling to be woken up and being woken up potentially taking some time, which appears to result in some "drift"; everything is scheduled a tad to late and over time that slight error adds up.

This changes things up a bit and uses a start timestamp `utilTimerStartTime`.  
The period until we want to be woken up is then calculated as `utilTimerStartTime + utilTimerPeriod - currentTime`.  
When we are woken up, we add the last `utilTimerPeriod` to `utilTimerStartTime`, to prevent us from accumulating those slight errors.

I did this because I noticed that on Jolt.js, Waveform playback is noticeably too slow when connected via BLE, or not connected at all, but absolutely fine when connected via serial; this change seems to fix Waveform playback speed.

I've been running this on a Jolt.js, Puck.js, XiaoBLE and my Bangle.js 2 for the past day, and haven't noticed anything odd yet.
Still, hopefully this doesn't introduce any regressions :sweat_smile:

To verify that scheduling accuracy is better (apart from audio playback sounding a lot nicer on Jolt.js), I hooked up a logic analyzer to pin 0 on a XiaoBLE and ran the following code while connected via BLE:
```js
analogWrite(D0,0.1,{forceSoft:true,freq:1000});
```

Results without this patch:
withoutPatch

Results with this patch:
withPatch

And that's only for ten pulses; as time goes on the "drift" without this patch only gets worse, while we can comfortably keep up the 1 kHz with this patch applied :)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -25,6 +25,7 @@
             Fix potential overflow of locks - favour small memory leak over accidental free if this ever happens (fix #2616)
             micro:bit2: Ensure we don't initialise the I2C1 peripheral (we use software I2C internally)
             Unlock functionCode early in function execution to reduce locks needed during recursion (#2616)
+            Improve timer scheduling accuracy
 
      2v25 : ESP32C3: Get analogRead working correctly
             Graphics: Adjust image alignment when rotating images to avoid cropping (fix #2535)
--- a/src/jstimer.c
+++ b/src/jstimer.c
@@ -26,8 +26,8 @@ volatile unsigned char utilTimerTasksTail = 0;
 volatile bool utilTimerOn = false;
 /// When we rescheduled the timer, how far in the future were we meant to get called (in system time)?
 int utilTimerPeriod;
-/// The system time at which the util timer's period was last set
-JsSysTime utilTimerSetTime;
+/// The system time from which the timer was last scheduled - add utilTimerPeriod to this to get the time when the timer should be called next
+JsSysTime utilTimerStartTime = 0;
 /// Incremented with utilTimerPeriod - used when we're adding multiple items and we want them all relative to each other
 volatile int utilTimerOffset;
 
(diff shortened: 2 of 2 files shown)


gfwilliams:
Thanks - but I think if you look back at the commit history it's what we used to do originally. The reason it's not done like this now is that jshGetSystemTime runs from the RTC, so it's only accurate to 1/32768 sec. So for 8kHz audio it's fine, but when folks are using `digitalPulse` for short pulses or software PWM it can get pretty glitchy. 

It's frustrating as we're using the same timer for two different things - highish precision (more than we get from the RTC) but also with audio we need something a bit more accurate over time.

Currently we clear the timer before processing the interrupt, but I've just committed some code that leaves the timer running and so effectively takes account of the time that the interrupt takes to run, which hopefully fixes this. analogWrite definitely seems a lot better and the frequency isn't changed by having a BLE connection.

It ended up being a bit of a pain because you can't actually stop IRQs, so you can never be entirely sure that you set the capture/compare register up in time. Hope that helps!

ssievert42:
Impressive!

> The reason it's not done like this now is that jshGetSystemTime runs from the RTC

That probably explains some not particularly pretty glitchiness I had at higher frequencies.
There's definitely an audible quality improvement between my patch and the change you made :)

But seems like everything is still a tad too slow.

And doesn't changing the contract of `jshUtilTimerReschedule()`, to interrupt the last time the timer triggered plus `period`, instead of just interrupting after `period`, break `jstRestartUtilTimer()` if we've already got a running timer?

> It ended up being a bit of a pain

Another suggestion: Maybe we could, instead of trying to reschedule the timer, just go ahead an sneakily do `NVIC_SetPendingIRQ(TIMER1_IRQn);`?
(I tried this with `TIMER2_IRQn` (plus and interrupt handler for timer 2 and `NVIC_EnableIRQ(TIMER2_IRQn);`); setting the interrupt as pending seemed to work, both if we're currently not in an ISR and from within the timer 2 ISR.)

I had a go at changing stuff around to try to achieve maximum accuracy™; please let me know what you think: #2622

gfwilliams:
> And doesn't changing the contract of jshUtilTimerReschedule() ... break

I don't think so? Not looking at the places it was called from. I could be wrong though - did you notice any issues?

> Maybe we could, instead of trying to reschedule the timer, just go ahead an sneakily do NVIC_SetPendingIRQ(TIMER1_IRQn);?

I'll discuss on https://github.com/espruino/Espruino/pull/2622

PR #2622, opened 2025-04-06
nRF52: Further improve software timer accuracy

ssievert42:
This is yet another attempt to improve timer scheduling accuracy.

While 0b2724c3076f9f8a0c7df41e59ec2947f20bbd34 helped a lot, everything still seemed a tad too slow.

At least with a logic analyzer connected to D0 on a XiaoBLE and running:
```js
analogWrite(D0,0.1,{forceSoft:true,freq:10000});
```

I expected 100 us between rising flanks, but got 115 us when building from `master` and 100 us with this patch.

The major change this brings is to trigger the timer interrupt manually by calling `NVIC_SetPendingIRQ(TIMER1_IRQn);` when we notice that the timer has counted beyond the value we've written to the compare register.
This should also fix up `jshUtilTimerStart()` to actually interrupt after the passed period, by pretending that the last time the timer should have triggered is now.

Potentially problematic: This touches `jshGetTimeFromMilliseconds()` and changes it to round to the nearest time in system ticks, because `E.dumpTimers()` was reporting 99 us for 100000 Hz. (It now correctly says 100 us.)

Also, the whole wraparound stuff probably doesn't work on nRF51, where (it looks like) `NRF_TIMER_MAX` is only 16 bits instead of 32. But the same can be said about the code currently in `master` :see_no_evil:

This is by no means perfect; all feedback and suggestions welcome :)


gfwilliams:
>  Maybe we could, instead of trying to reschedule the timer, just go ahead an sneakily do NVIC_SetPendingIRQ(TIMER1_IRQn);?

In theory that sounds good.

Just so I'm sure, so the issue is when running at higher frequencies where it's more likely a bluetooth interrupt will come in and by the time the timer is to be scheduled next it's already passed? So at lower frequencies the existing code works reasonably accurately?

> jshGetTimeFromMilliseconds

So this is just so that you see it reported correctly, rather than it being for anything else?

I would have thought we could have done it better with something like:

```
return (JsSysTime) ((ms + (0.5*SYSCLK_FREQ/1000)) * (SYSCLK_FREQ / 1000.0));
// or

return (1+(JsSysTime) (ms * (SYSCLK_FREQ / 500.0))) >> 1; // probably a lot faster
```

My concern is I'm pretty sure this function (and jshGetMillisecondsFromTime) gets called a lot and inlined, and I've had a few issues with them - either using up way more memory than expected and being slow, or otherwise adding inaccuracy:

https://github.com/espruino/Espruino/commit/5be8869ae36b7d9fe955e1b1071b9f8a58a038cf
https://github.com/espruino/Espruino/commit/6fc3a99299cc5da46772a52b17fa4ce62d9465a1

So I want to be a bit careful around changes, specifically those that might add more double precision operations.

> the whole wraparound stuff probably doesn't work on nRF51

I did have that in mind and did try the current code on nRF51 and it appears to work, but I'm not sure if your changes will?

Why do you now use `utilTimerTriggerTime`? I thought just querying the CC0 channel was ok?

ssievert42:
Thanks for taking a look at this!

> In theory that sounds good.

Well then, let's hope that theory and practice don't diverge too much :)

>  running at higher frequencies where it's more likely a bluetooth interrupt will come in

Yup, that's probably the issue. I'm still toying with audio output, and at 8 kHz the playback speed difference is kind of noticeable.
(Not sure how I'd figure out if it actually is a bluetooth interrupt; I can't even get execution to continue after hitting a breakpoint when using SWD :see_no_evil: (openocd just says "external reset detected" and the board resets))

> jshGetTimeFromMilliseconds

It's not just to have it reported correctly; there's also a 1 us timing difference in the output when doing software PWM at 10 kHz.
I still remember running into the time accuracy problem on Bangle.js 2 that you fixed with 6fc3a99299cc5da46772a52b17fa4ce62d9465a1...
So it's probably not big enough of a problem to warrant messing with `jshGetTimeFromMilliseconds()` :sweat_smile:
I'm gonna go ahead and remove those changes for now.

> wraparound stuff

I kept your math magic; so this should in turn also work.

Except maybe for:
```c
uint32_t thisCC = utilTimerTriggerTime + period;
```
which should really be:
```c
uint32_t thisCC = (utilTimerTriggerTime + period) & NRF_TIMER_MAX;
```


But doesn't `if (((int32_t)thisCC - (int32_t)current) < 2)` break if we've got a 16 bit timer and we're close to wraparound?
For example with `thisCC = 0x0001` and `current = thisCC - 5` so `current = 0xFFFC`.
In that case the if statement should evaluate to `false` for 16 bit math, but if I'm not mistaken, evaluates to `true` instead.

> Why do you now use `utilTimerTriggerTime`?

(Also in response to [your comment here](https://github.com/espruino/Espruino/pull/2620#issuecomment-2796604977))

That's so that in `jshUtilTimerStart()` we can do:
```c
// schedule at the current time + period
// by pretending that the last time we wanted to be called at is now
NRF_TIMER1->TASKS_CAPTURE[1] = 1; // get current timer value
utilTimerTriggerTime = NRF_TIMER1->CC[1];
```
That's needed because when inserting a new task, we may do: `utilTimerInsertTask()` -> `jstRestartUtilTimer()` -> `jshUtilTimerStart()`, where we expect to be interrupted after `period`, whether we've got a running timer or not.

If we wrote to the CC0 channel instead, we'd potentially trigger the timer interrupt.
There's probably a better way to get `jshUtilTimerStart()` to interrupt after `period`, but adding a variable seemed like the most straightforward one.

ssievert42:
Alright, just removed the changes to `jshGetTimeFromMilliseconds()` and added a `& NRF_TIMER_MAX` to the timer target value calculation.

gfwilliams:
> I can't even get execution to continue after hitting a breakpoint when using SWD 🙈

There's a magic command to send (`(*(uint32_t*)0x4001050C)=1`) right after it breaks which will then stop the next steps from rebooting it!

> But doesn't if (((int32_t)thisCC - (int32_t)current) < 2) break if we've got a 16 bit timer and we're close to wraparound?

Ahh - probably! I didn't check that seriously (just that it still worked) so maybe because it's only 16 bit if it's missed it's not the end of the world as it doesn't take *that* long before it just wraps and re-executes anyway. `NRF_TIMER_MAX` seems like a great fix.

So one last thing - I'm just looking at this, and is it possible that if we do `jshUtilTimerStart(1)` (or a low value) outside an IRQ, that the first `nrf_timer_cc_write` causes an IRQ almost immediately, but then subsequent calls overwrite the value that was set by `jshUtilTimerReschedule` that was called by the IRQ (which occurs while jshUtilTimerStart is still executing), and then we maybe even force a call again with NVIC_SetPendingIRQ?

ssievert42:
> There's a magic command to send...

Thanks, that seems to help!
Although I keep getting resets after some time of inactivity after continuing from a breakpoint, the resets directly after continuing from a breakpoint are no more; that's debuggable enough for me :)

> if we do jshUtilTimerStart(1) (or a low value) outside an IRQ...

So I've been staring at the code for a while and came to the conclusion that while that absolutely would be an issue if `jshUtilTimerStart()` was called outside an IRQ with interrupts enabled while the timer is running, we never actually do that.

As long as `jshUtilTimerStart()` only gets called by `jstRestartUtilTimer()` which is in turn called from
- `utilTimerInsertTask()` while interrupts are disabled (and nothing bad can happen)
- `jsble_restart_softdevice()` after having disabled the timer with `jshUtilTimerDisable()`

we should probably be fine, because if the timer is disabled, we only do a single `nrf_timer_cc_write` at the very end of `jshUtilTimerReschedule()`.
(To verify that (and because I now can :) ) I stepped through the `jstRestartUtilTimer()` call from `jsble_restart_softdevice()`.)

Does `jshUtilTimerStart()` need to be callable outside of an IRQ if the timer is currently active?

gfwilliams:
> Does jshUtilTimerStart() need to be callable outside of an IRQ if the timer is currently active?

I don't think so, no... In that case jshUtilTimerReschedule is more likely. Realistically it should happen from utilTimerInsertTask

Thanks for looking into that and for all your work on this! I'm happy to merge it in then!

gfwilliams:
Sadly it looks like this has broken `Puck.IR` on the Puck.js devices for the 2v26 release. 

I'm not entirely sure why, but as reported at https://github.com/orgs/espruino/discussions/7774 it creates a `InternalError: Timeout on Utility Timer` error and subsequent attempts to use the timer after this fail until after calling a `reset()`

I'm reverting this for now...

PR #2624, opened 2025-04-07
Fix globbing and legacy shell syntax in provision.sh

jpcano:
As per building documentation, I executed this:

```shell
source scripts/provision.sh THINGY52
```
And I got this a `==== not found` error.

Selection_016

There are some legacy idioms in `provision.sh` script that can have undefined behaviour in some machines/shells (like mine) so I implemented the following recommendations from Shellcheck across the whole file:

https://www.shellcheck.net/wiki/SC2086
https://www.shellcheck.net/wiki/SC2006

### DIFF
--- a/scripts/provision.sh
+++ b/scripts/provision.sh
@@ -48,24 +48,24 @@ if [ "$BOARDNAME" = "ALL" ]; then
   PROVISION_EMSCRIPTEN=1
   PROVISION_EMSCRIPTEN2=1
 else
-  FAMILY=`scripts/get_board_info.py $BOARDNAME 'board.chip["family"]'`
+  FAMILY=$(scripts/get_board_info.py "$BOARDNAME" 'board.chip["family"]')
   if [ "$FAMILY" = "" ]; then
     echo "UNKNOWN BOARD ($BOARDNAME)"
     return 1
   fi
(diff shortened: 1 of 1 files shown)


gfwilliams:
Just so I know, what particular shell/OS are you using that this isn't working on?

Because for GitHub we use a ubuntu-24.04 base for builds and it's fine on there

jpcano:
> Just so I know, what particular shell/OS are you using that this isn't working on?
> 
> Because for GitHub we use a ubuntu-24.04 base for builds and it's fine on there

My shell is **zsh 5.8.1 (x86_64-ubuntu-linux-gnu)** and it is running on **Linux Mint 21.2 Cinnamon**. I think the OS is irrelevant here because `source` is a shell built-in command. The ShellCheck recommendations I implemented make the script less ambiguous and its interpretation more strict and therefore now zsh can execute it correctly.

gfwilliams:
Ok, thanks - just merging.

In general though I'd prefer not to have all these sweeping changes if they're not absolutely needed. Even if it works for you with zsh there's no guarantee that it won't cause issues on Macs.

It looks like your `==== not found` error could have been fixed just by adding `"` around the echo...

PR #2625, opened 2025-04-15
Improve documentation of the packet transfer protocol

rikkuness:
While trying to get my head around the packet transfer I made some notes, thought they were maybe worth contributing back. Is my understanding of it correct?


gfwilliams:
Thanks! Yes, that looks spot on! Just FYI this has come up: https://github.com/espruino/BangleApps/issues/3792#issuecomment-2804668109

I'm looking at it now but it might be that some changes are needed to the uart.js implementation

PR #2626, opened 2025-04-20
Bangle.setUI: fix back btn not drawn in leftmost position

thyttan:
The bug seems to have snuck in when the back widget was no longer set
with Object.assign in commit https://github.com/espruino/Espruino/commit/453def81645c08e46195ad0627c1b77c8a59ee2a .

Before this PR:
image

After this PR:
image

Not sure about the minified files. Looks like you may have used more aggressive minification before @gfwilliams? Please apply that again in that case.

Note:
When working this out, I first tried to do stuff with widgets `sortorder` property, but that didn't seem to have any effect?


gfwilliams:
Great, thanks! I'd wondered why `Object.assign` was used - I should have digged into that deeper! Yes, I use the `EspruinoDocs/bin/minify.js` IIRC but it was updated to do pretokenisation

PR #2627, opened 2025-04-23
Document how to build for different ESP32 Series

jpcano:
I was going to add support for the ESP32-S3 chip, as I thought this was not supported.  I found this by chance after digging in the boards' files. Thank you for the great work on this!

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ Main supported platforms are:
 * nRF52
 * nRF51
 * ESP8266
-* ESP32
+* ESP32, ESP32-S3, ESP32-C3
 * Linux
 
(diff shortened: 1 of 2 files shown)


gfwilliams:
Thanks - but do you think you could maybe just add some docs that say to replace ESP32 with ESP32S3_IDF4/etc?

Duplicating the same lines of code seems a bit over the top and just makes it more painful to maintain in the future.

After all right at the top of the file it says:

> To work out which 'Board name' to use, simply look in the boards folder for the right BOARD.py file. To build for boards/PUCKJS.py use `BOARD=PUCKJS RELEASE=1 make`

And we don't have separate sets of building instructions for all the other boards that are supported

jpcano:
I removed the duplications. Hopefully, now it is clearer, and this makes the discoverability of the supported boards a bit easier for someone reading the README for the first time.

gfwilliams:
Excellent - thank you!

PR #2636, opened 2025-05-26
Bangle_setUI_Q3: fix btn back handler was not always rm:ed correctly

thyttan:
Using the recent `messagegui` updates in combination with Fastload Utils exposed a bug with `Bangle.setUI` where back on hw button watches could end up not cleared after setting up a the UI of the clock after `unreadTimeout` of `messagegui` ran out. This seems to fix that, when testing this special case at least. I just finished the changes so have not tested for long.

@gfwilliams please say if we should change something before merge.

When this looks good I can fix the Bangle.js 1 implementation as well.

### DIFF
--- a/libs/js/banglejs/Bangle_setUI_Q3.js
+++ b/libs/js/banglejs/Bangle_setUI_Q3.js
@@ -10,11 +10,10 @@
     hadBackWidget = true; // if we had a back widget already, don't redraw at the end
     WIDGETS.back.remove(options.back); // only redraw when removing if we don't have options.back
   }
-  let clearBtnWatches = function() {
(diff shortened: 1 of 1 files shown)


gfwilliams:
Sorry - missed this last week!

gfwilliams:
I've just committed some changes - I think we don't need to worry about removing the listener in `.remove` since the only place we call it is setUI line 11, and we clear the button watches 2 lines later.

It makes the code much tidier - and it's probably something I should have done originally. Does that look good?

If so we can do the same on Bangle.js 1?

thyttan:
I think that will do :) let's go with that.

PR #2639, opened 2025-06-10
update micro:bit v2

fanoush:
- fix board info to reflect 128k of ram
- enable JIT
- enable MTU 58, 2 centrals, packed symbols, trimmed startup code  like in other nrf52 boards


gfwilliams:
That's great - thanks!

Issue #2641, opened 2025-07-10
Bangle.js buzz when back widget tapped


RKBoss6:
For me, the native back widget seemed to be quite finicky to use, and I had no idea why. Having a bit of haptic feedback (`Bangle.buzz()`) when the back widget is clicked would be a really lovely feature to give haptic feedback that you touched the button, hopefully preventing frustrated touches in the area until you get to it. I looked over the SetWatch function, and the back widget seems to be implemented in there, but as this is firmware related stuff, I've got no idea on how to test that part without needing to rebuild the firmware version each time. Does anyone know how to do this safely?
Thanks!

I also feel as though this might be part of a larger update, adding more haptic feedback, for instance, when you long press and release, a buzz to show that it's going to the clock would be good.

gfwilliams:
That sounds like it could be a good idea. You can change the built-in functions by overriding them in JS - there's a tutorial on it at https://www.espruino.com/Bangle.js+Modification which might be a good place to start and test things? Then when it's ok we can push the changes to firmware

I think you need to modify setUI - there is already some code in there for haptic feedback, it just doesn't appear to be rolled out to everything: https://github.com/espruino/Espruino/blob/master/libs/js/banglejs/Bangle_setUI_Q3.js#L38-L40

Handling the long-press is a bit trickier - that's possibly something that'd have to be in the firmware.

gfwilliams:
I'm not sure what you mean about the back function, but the touch handler is at https://github.com/espruino/Espruino/blob/master/libs/js/banglejs/Bangle_setUI_Q3.js#L132-L134

Literally just add `b()` to that function

RKBoss6:
@gfwilliams Alright, I attempted to create a PR for review.

RKBoss6:
I think an app could achieve this with more flexibility though, what do you think?

PR #2642, opened 2025-07-11
[Bangle.js] Add buzz on back widget pressed

RKBoss6:
This is in regards to [Issue 2641](https://github.com/espruino/Espruino/issues/2641)

### DIFF
--- a/libs/js/banglejs/Bangle_setUI_Q3.js
+++ b/libs/js/banglejs/Bangle_setUI_Q3.js
@@ -130,6 +130,7 @@
       var touchHandler = function(_,e) {
         if (e.y<36 && e.x<48) {
           e.handled = true;
+          b();
           E.stopEventPropagation(); // stop subsequent touch handlers from being called
           options.back();
         }


gfwilliams:
Great - and just to check, this is tested (via the tutorial mentioned in https://github.com/espruino/Espruino/issues/2641) and it works ok for you?

RKBoss6:
@gfwilliams Hi! I just tested this a few times, and it works for almost everything. The place where it doesn't buzz, however, is right after you open an app like settings, and press the back widget. It simply loads the clock, and does not buzz. Do you know why this might be? Thanks!

gfwilliams:
> Do you know why this might be? Thanks!

Ahh, yes - it'll be because it immediately resets the device to load the new app, so the vibration motor is turned off basically as soon as it's turned on. For it to work nicely `b` would have to return the promise, then you could do `b.then(() => options.back())`

RKBoss6:
@gfwilliams That works now, I did not update the `Bangle_setUI_Q3.min.js` file, should I do that?
Additionally, what is the difference between the `Bangle_setUI_Q3.js` and `Bangle_setUI_F18.js` files?

RKBoss6:
@gfwilliams Ok, thanks so much for the help!

gfwilliams:
Well, you can just install a cutting edge build and get it, so personally I don't think an app would be that much help?

PR #2645, opened 2025-07-18
Added IGNORE_SYSTEM_GCC option to provision.sh

### LINKED ISSUE #2644: Allow `provision.sh` to use embedded GCC and ignore system installed version
I have `arm-none-eabi-gcc` installed from system packages so the `provision.sh` ofc warns:
```
===== ARM
arm-none-eabi-gcc installed, version 14.2.0
*********************************************************************
*********************************************************************
***                                                               ***
*** ARM GCC IS INSTALLED ALREADY, BUT IS NOT THE EXPECTED VERSION ***
***                                                               ***
***   The build may work with your compiler version, but it       ***
***   is possible you will encounter errors, or issues with       ***
***   build size                                                  ***
***                                                               ***
*********************************************************************
*********************************************************************
      Expected 13.2.1
      Got      14.2.0
```

I would prefer to use the embedded version so versions match up. It'd be neat to have an env var that allows skipping the check for the existence in the `PATH` and then installs `gcc-arm-embedded` anyway. Then when it updates the `PATH` with the path to the extracted version, put it at the start of `PATH` instead of the end so that it takes preference over the system version.

rikkuness:
Fixes #2644 .

I also updated it to conditionally update `PATH` so re-running doesn't keep stacking up the same thing in there.

Also moved the path to the start of `PATH` which will mean all `arm-none-eabi-` binaries now will preferentially use the ones from the locally installed toolchain over any existing system paths.

Seems to work in all my local testing so far.

### DIFF
--- a/scripts/provision.sh
+++ b/scripts/provision.sh
@@ -273,7 +273,8 @@ if [ "$ARM" = "1" ]; then
     echo "===== ARM"
     EXPECTEDARMGCCVERSION="13.2.1"
     EXPECTEDARMGCCFILENAME="arm-gnu-toolchain-13.2.Rel1-x86_64-arm-none-eabi"
-    if type arm-none-eabi-gcc 2> /dev/null > /dev/null; then
+
+    if type arm-none-eabi-gcc 2> /dev/null > /dev/null && [[ -z $IGNORE_SYSTEM_GCC ]]; then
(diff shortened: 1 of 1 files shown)


gfwilliams:
Thanks! looks good to me :)

PR #2656, opened 2025-08-22
E_showPrompt_Q3: allow long press actions

thyttan:
as proposed at https://github.com/espruino/BangleApps/issues/3915#issuecomment-3212066440

### DIFF
--- a/libs/js/banglejs/E_showPrompt_Q3.js
+++ b/libs/js/banglejs/E_showPrompt_Q3.js
@@ -64,6 +64,9 @@
           draw(i); // highlighted button
           g.flip(); // write to screen
           E.showPrompt(); // remove
+          if (e.type===2 /*long press*/ && options.buttonsLong && options.buttonsLong[btns[i]]) {
+            resolve(options.buttonsLong[btns[i]]);
+          }
           resolve(options.buttons[btns[i]]);
         }
       });


RKBoss6:
If it works, and is solid, add it to `showAlert` as well for consistency?

gfwilliams:
IMO it's nicer to have:

```
          if (e.type===2 /*long press*/ && options.buttonsLong && options.buttonsLong[btns[i]])
            resolve(options.buttonsLong[btns[i]]);
          else
            resolve(options.buttons[btns[i]]);
```

While promises will ditch the second call to resolve, it's clearer when reading the code that we only expect one to be run?

But otherwise this looks good - I'm happy for it to go in, but it would be good if the docs change went in at the same time (and I don't know if we care about Bangle.js 1? maybe at this point we don't worry)

thyttan:
I agree with adding that `else`, will do!

I can look at adding to Bangle.js 1 if it makes sense?

gfwilliams:
Thanks! Are you able to test on Bangle.js 1 at all? I think it might be good just to keep things working the same across all devices. Then you'll still be able to get the snooze options even on that

thyttan:
Not able to test aside from on B1 emulator no 🫡

gfwilliams:
Hmm, actually I just looked and it's going to be hard to do, because right now it just uses the `updown` setUI mode, and everything responds on the rising edge anyway. Let's leave it for Bangle.js 1 then.

RKBoss6:
Should we add this for showAlert as well to keep things consistent?

gfwilliams:
> Let's add this for showAlert as well to keep things consistent, as it could also benefit from long-press actions...

showAlert should be find as it calls into showPrompt anyway. If you supply buttonsLong as an option it should 'just work': https://github.com/espruino/Espruino/blob/master/libs/js/banglejs/E_showAlert.js

thyttan:
This is ready to go in I think

thyttan:
@bobrippling added your suggestions and attributed to you. Thanks!

bobrippling:
> @bobrippling added your suggestions and attributed to you. Thanks!

Thanks!

RKBoss6:
@thyttan Not necessarily a bug maybe, but if the user doesn't provide a message, it defaults buttons to resolve true or false.
I might open up an issue later, but right now it's not an inconvenience, as prompts should have messages in them by default.

gfwilliams:
Thanks! Looks good - just merged

PR #2660, opened 2025-10-23
[Bangle.js 2] Add the option to set accent color through `E.showPrompt` options

RKBoss6:
Added an option `options.accentColor` which sets the accent color for the prompt to be shown (Buttons and title bar). It would be a niche use case, but I can see this used in further customization of the watch, color coded alarms, or urgent alerts. I'm debating adding this to Bangle.js 1, but I see no reason not to, so I'll probably add that.

Since this is my first commit here, what all steps do I need to take to adjust the .min file and changelog accordingly?

Examples:

download (1)
download

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,5 @@
-          : Add `E.internal` as a way to access the 'hidden root' containing Espruino internal variables that previously needed `global["\xff"]`
+          : Bangle.js1,js2: Add accentColor option to E.showPrompt(), which changes the color of the tile bar and buttons.
+			Add `E.internal` as a way to access the 'hidden root' containing Espruino internal variables that previously needed `global["\xff"]`
             Bangle.js: Fix back handler not removed when using E.setUI with a back button but without widgets (#2636)
             nRF52: Add 'slaveLatency' option to BluetoothRemoteGATTServer.connect, default changed from 2 to 4
             nRF52: Add 'phy' and 'mtu' events to NRF/BluetoothDevice when these change on a connection


gfwilliams:
Honestly, I feel like this is massive overkill. showPrompt/etc all use the current theme colours, so why can't you just use g.setTheme for the few times you want to do this?

RKBoss6:
Yes, you're right, it would annoy people with custom themes, but I don't imagine this to be a use all the time feature, rather for something like the app `phoneremind` where you want to display a critical alert that your phone was left behind, it would be good to see a visual of the urgency (color red). I imagine this to be a feature used by maybe a handful of apps that need the colors to signify different things, for example, or show urgency in some way. Instead of setting g.theme, which will override other theme colors as well, it's better to have it centralized for the rare use case where you need it.

gfwilliams:
I don't really see what's wrong with:

```
let oldTheme = g.theme;
g.setTheme({bgH:"#f00",bg2:"#f00"});
E.showPrompt("Hello").then(() => {
  g.setTheme(oldTheme);
});
```

If you change the theme to display one thing you can just revert it right after the menu disappears?

I'm just not too fond of changes like this which aren't really needed, are only for one app, and are undocumented. I've pulled in features like this before and it's pretty much always a maintenance nightmare and they rarely end up being used by anyone else.

gfwilliams:
Being devil's advocate for a second: If people are complaining about UI cohesiveness, making the colours *different for every app* might not actually be going in the direction they want?

> Changing the theme and resetting it upon leave is tricky because if a boot.js script shows a prompt, it is my understanding that the theme will stay?

Well, boot.js runs before the app, so should use the system theme if something is drawn at that point? Eg. `bootupdate.js` 

... if code runs later, and the app unloads itself properly and restores the theme in the `.remove` handler then the theme will update to what it was before, for example:

```
let oldTheme = g.theme;
g.setTheme({fg:"#f00",fg2:"#f00"});
E.showPrompt("Old Theme", {
  remove : () => { print("Remove"); g.setTheme(oldTheme); }
}).then(() => E.showPrompt("New Theme"));
```

To be honest this just makes me double down on my position on this:

We have a way of changing colours already, in one place *for all types of menu* - it's called themes. That's what they are there for, they work great, and we use them already (for example in Wave Clock) to ensure widget backgrounds are consistent.

The last thing I want is another way of changing colours that would have to be implemented again for all the different UI elements, and which has to be applied separately and in a slightly different way for every different UI function.

PR #2668, opened 2025-11-27
Fix accel limit bound

bobrippling:
[Spotted here](https://github.com/espruino/BangleApps/issues/4067#issuecomment-3583091172) - no repro but this looks like a typo


gfwilliams:
Thanks!

PR #2672, opened 2025-12-13
[E.showMenu] Add option to add separators

RKBoss6:
Added the option to add separators by adding this to your existing menu:
```javascript
"General Settings":"separator",
``` 

And it'll make a separator for that, for better organization.
I got the idea when looking at app settings pages, some of which were really long, and not that organized. Instead of adding different pages for each category, it would be nice to have it all scrollable and still organized.

Screenshots:

download (1)

\
\
s

Hopefully it'll help keep things organized in the future, and I think it looks pretty clean and modern as well.

### DIFF
--- a/libs/js/banglejs/E_showMenu_Q3.js
+++ b/libs/js/banglejs/E_showMenu_Q3.js
@@ -1,4 +1,4 @@
-E.showMenu=(function(menu) {
+(function(menu) {
   const H = 40;
   if (menu===undefined) {
     g.clearRect(Bangle.appRect);


thyttan:
>> > Perhaps this change becomes a customisation app instead, in a similar vein to promenu?

> The problem with that is it requires app implementation, not just an install that changes how the menus work.

Maybe you could detect sub-menus, bring them out into main menu and use the names of the sub-menus as header names? That way you would not need updates to apps.

gfwilliams:
Sorry, but right now I don't think this is a good idea. I don't see much of a sign of others wanting this other than you. People trying to bodge this using fake menu items as headings in `E.showMenu` might be a sign that there is demand, but I don't see that happening much. There's just so little space on the screen that I'm not sure having separators is helpful as it actively means more scrolling to tap on the item you want.

As I mentioned, we had some complaints over having too many menu items onscreen, and had to split things up. This seems like it's going backwards by encouraging developers to do exactly the thing that we know users complained about before and we worked hard to avoid.

... and I have a feeling that if merged, you will want to actively go through apps in the App Loader pulling submenus into one long menu with separators *which we know users objected to before*, and then we'll have long discussions in each of those PRs about it too.

There's also a long history of folks contributing firmware changes to deal with specific use cases, and then them never really being used and me being stuck having to maintain them - so I really do have to push back against it now. Take #2656 for example - how many uses of `buttonLong`? One, in sched, 4 months ago - and it'll probably never be used again.

The problem is it's not just this change. I'd want to tweak it slightly, I have to minify the files, test on Bangle.js 1 and 2, and then you'll want to modify existing apps to use it - and checking all those will be a pain for the BangleApps maintainers. After all that I can pretty much guarantee that there will be messages/complaints about those app changes, and *definitely* about the increased padding at the top.

I'm afraid I'm just not up for handling that at the moment

PR #2677, opened 2026-01-15
Fix for unlink not updating cluster count for file deletion.

AidansLab:
Not sure if this issue exists on other devices, but on the Pip-Boy, if a file is deleted, the free cluster count is not properly updated. I believe this is because the cluster count is only updated if a directory is deleted, and does not update if a file is deleted. So I moved the cluster update code appropriately so it updates for both.

### DIFF
--- a/libs/filesystem/fat_sd/ff.c
+++ b/libs/filesystem/fat_sd/ff.c
@@ -3485,9 +3485,10 @@ FRESULT f_unlink (
 			} else {
 				if (dir[DIR_Attr] & AM_RDO)
 					res = FR_DENIED;		/* Cannot remove R/O object */
+				else
+					dclst = ld_clust(dj.fs, dir);	/* Get cluster chain for files and directories */
 			}
 			if (res == FR_OK && (dir[DIR_Attr] & AM_DIR)) {	/* Is it a sub-dir? */
-				dclst = ld_clust(dj.fs, dir);
 				if (!dclst) {
 					res = FR_INT_ERR;
 				} else {					/* Make sure the sub-directory is empty */


CodyTolene (approved):

gfwilliams:
@rblakesley do you have any thoughts on this?

Does this happen on the default pip-boy build, or only when you build it yourself? We tend to do the official builds off a fork that had a newer FatFS library which might have this fixed (which I haven't been able to upstream yet because it increases teh binary size and breaks some Espruino builds).

I'm a bit concerned about changing FatFS though - it's an extremely well-used library and I'd imagine pretty well tested, so unless this is something that is in FatFS upstream I'd have to assume there's a reason it's not there.

rikkuness:
0.11 feels safe, a lot of the releases from 0.12 onwards were fixing bugs with 0.12 😄

AidansLab:
Agree 100%, upgrading 1 version is way better than modifying the library (though I am a tiny bit sad I won't have a piece of code I wrote in all Pip-Boys, would have been fun to brag about, lol). 

> hopefully without increasing the code size too much or introducing any breaking changes.

Just looked, ff.c from 0.11 is actually 2 kb **smaller** than 10c.

I'm glad we were able to figure this out. This has been annoying people for a very long time.

rblakesley:
> Just looked, ff.c from 0.11 is actually 2 kb smaller than 10c.

That's good news!

gfwilliams:
Thanks for checking it @AidansLab!

> we can just move to 0.11 and it should fix the issue

This would definitely be my preference as well. I just looked at the 0.15 code as it was easily on GitHub and the change is exactly like you have done: https://github.com/hhuOS/FatFs/blob/master/source/ff.c#L5020C6-L5020C35

So I'll merge this for now as it's not going to hurt either way.

>  ff.c from 0.11 is actually 2 kb smaller than 10c.

Brilliant - thanks for checking that. That's actually about the amount of memory I was finding I was over budget!

So yes, let's update FatFS at some point.

AidansLab:
You guys are awesome, really. I'm excited for the next update. I might hold up on releasing the Pip-Cam until the next official update is out, we'll see, still got a few bugs and kinks to work out anyway. (Also, I'm gonna send out a free Pip-Cam to one of the discord members who's been super kind, so if any of you guys want one, I'd truly be honored to send you guys one or even a few, it will be fully open source too).

AidansLab:
Yeah, I was looking at the source file since I was on my laptop and couldn't compile it. That's so awesome, this is definitely one of my proudest achievements now.

PR #2680, opened 2026-01-21
Add true rounded corners to E.showMenu

RKBoss6:
This PR just makes the corners in buttons truly rounded - before it was a poly with harsh edges. Hopefully this helps subtly make the interface feel a little more fresh and modern - without changing anything major. I know we've had some issues with changes like this @gfwilliams, so I've avoided changing anything that isn't related to the rounded corners at all, and most people probably won't notice/comment on it anyway. Overall, this PR is designed to be as minimal as possible, only using `g.fillRect()` calls, and hopefully helps to make the prompt interface feel a little fresher. 

### Screenshots:

download (3)

download (2)

### DIFF
--- a/libs/js/banglejs/E_showPrompt_Q3.js
+++ b/libs/js/banglejs/E_showPrompt_Q3.js
@@ -35,7 +35,7 @@
       btnPos.push({x1:x-2, x2:x+BW-2,
                    y1:y, y2:y+BH});
       var btnText = g.findFont(btn, {w:bw-4,h:BH-4,wrap:1});
-      var r = 10;
+      var r = 11;
       var isHL = (idx===highlightedButton);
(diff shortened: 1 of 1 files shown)


RKBoss6:
That could just be dithering+ actually being able to see the pixels. In the watch it looks really nice, and if you want you can pop it in as a boot.js script to test it out...

bobrippling:
> Is this really an improvement do you think? The corners look a little bold and uneven:
>
> @bobrippling @thyttan any thoughts?

Between the two:
> download (3) download (2)

I prefer the original as it's more blocky, like an old school game, rather than the rounded one, which shows pixelation but less intentionally - like what you say about it being bold and uneven:

> image

Not to diminish your effort on this though @RKBoss6, I appreciate your desire to improve things

PR #2704, opened 2026-04-11
Bangle.js 2: preserve nested redraw UI in uiRemove

stweedo:
Abort stale outer setUI work when uiRemove synchronously installs a redraw-capable replacement UI, such as E.showMenu. Add a regression test covering nested setUI during uiRemove.

This appears to match an existing class of menu/UI handoff workarounds already present in BangleApps, where apps defer E.showMenu or submenu transitions to avoid redraw/navigation glitches. This may not be a complete list, but current examples include:

  - wrkmem: explicit workaround comments about menu glitches and phantom buttons in apps/wrkmem/
    app.js:79 and apps/wrkmem/app.js:668
  - alarm: repeated deferred menu handoffs in apps/alarm/app.js:104
  - daisy: menu rebuilds after option changes in apps/daisy/settings.js:78
  - counter2: menu rebuild after changing menu shape in apps/counter2/settings.js:30
  - draguboard: delayed E.showMenu(appMenu) in apps/draguboard/settings.js:29
  - dragboard: same pattern in apps/dragboard/settings.js:29
  - drained: delayed settings redraw in apps/drained/settings.js:60
  - fastload: delayed main-menu rebuild after linked setting changes in apps/fastload/settings.js:30
  - rebble: delayed showMenu() after changing menu contents in apps/rebble/rebble.settings.js:70
  - hadash: delayed submenu transitions in apps/hadash/hadash.app.js:202
  - flightdash: delayed submenu open in apps/flightdash/flightdash.settings.js:195
  - calendar: delayed edit submenu open in apps/calendar/settings.js:123
  - grocery: delayed edit submenu open in apps/grocery/settings.js:106
  - sleeplog: delayed debug submenu open in apps/sleeplog/settings.js:550


gfwilliams:
Thanks! I'm just trying to understand this a bit though - I get the issue caused by code that tries to change the menu in the `onchange` handler, but we do already have various mitigations for that in showMenu, for example: https://github.com/espruino/Espruino/blob/master/libs/js/banglejs/E_showMenu_Q3.js#L155

I just looked at the first few uses you linked and they all seem to be to do with onchange handlers in `E.showMenu` which is unrelated to the `remove` handler?

Are there any apps where setUI is called from within a remove handler, apart from your test case? And `uiRedraw` is not guaranteed to be set anyway (a lot of setUI calls do not set `redraw`).

I'd actually be inclinded to just throw an error if setUI is called from within a `remove` handler, because that seems like a bug.

`remove` should be called when the UI is removed because a new UI is being put up, and in it you should be freeing anything that was allocated for what was displayed - so the last thing you should be doing is trying to display something else.

stweedo:
Hi Gordon, I think you are probably right and it's likely a bug, and an error message would be better in that case. I may have hastily gathered those examples, and really only had the issue with the new color picker menu of rescalc during development and I thought I had to do that in the past as well. So I got curious and looked for apps that appeared to use set timeout to avoid the same bug. 

Basically without the uiRedraw check in uiRemove, then we have to use the setTimeout(0) in our back handler of the color picker module.

Example from https://github.com/espruino/BangleApps/pull/4208/changes:

  function closeColorBandPicker() {
    // Wait until the picker touch handler unwinds before rebuilding the menu.
    setTimeout(showColorEntryMenu, 0);
  }

function showColorBandPicker(bandNumber) {
    let pickerColorsData = getPickerColors(bandNumber);
    pickerBandNumber = bandNumber;
    pickerNames = pickerColorsData.names;
    pickerColors = pickerColorsData.colors;
    require("colorpicker").show({
      colors: pickerColors,
      emptyColor: 'None',
      showPreview: false,
      back: closeColorBandPicker,
      onSelect: setBandColorFromPicker
    });
  }

stweedo:
You were right, this was a bug within the new color picker module. The colorpicker module registered the same callback for both its back and remove handlers. This caused setUI/showMenu being triggered from inside a remove handler. 

I've made changes to fix this here: https://github.com/espruino/BangleApps/pull/4208/changes#diff-5b0b35c6b9c6bf08fa7b7e2fe2004004b1001cbfb0565150c889ab10bc358346L1-R123

gfwilliams:
Great, thanks for spotting that and getting a fix in!

While I could add an error check, I wonder how often this kind of problem actually occurs, or whether you were just unlucky?

stweedo:
Probably just coincidence and a wrong assumption on my part. Thanks for chiming in, your insight is always valuable.

PR #2706, opened 2026-04-18
fix errors in NRF_GPIO_PIN_WRITE_FAST and NRF_GPIO_PIN_READ_FAST on NRF52840 / port P1

arglurgl:
The #define macros NRF_GPIO_PIN_WRITE_FAST and NRF_GPIO_PIN_READ_FAST are not correct on port P1/ if pin number is >= 32. This affects e.g. an external flash chip on NRF52840 connected to pins on port P1.

This occurs for both NRF_GPIO_PIN_WRITE_FAST and NRF_GPIO_PIN_READ_FAST in both targets/nrf5x/nrf5x_utils.h‎ and targets/nrf5x_dfu/hardware.h. The cause is that the write macro omitted the correction for making the P1 port number zero-based and the read macro read from the wrong port and additionally omitted the port number correction.

This PR corrects the port and adds the calculations.


gfwilliams:
This looks great - thanks!

PR #2709, opened 2026-05-24
Fix incorrect meridian during the 1200 hour

jessegonyou:
This is a minor bugfix that came about when I was following the bangle.js clockface tutorial, and, since it happened to be around 12:05pm my time while I was following it, the watchface incorrectly displayed the time as 12:05am. I looked into it, and found that the issue was within the espruino builtin libraries themselves, where the `require("locale").meridian` function erroneously checked whether the 24-hour time was less than or equal to 12 in determining the meridian.

This fix simply replaces the less-than or equal (`<=`) operator with a less-than (`<`) operator, fixing this bug.


gfwilliams:
Looks great, thanks!

Just FYI for stuff in Espruino (not BangleApps) we now pretokenise by passing `--pretokenise` to the minifier as well (https://github.com/espruino/Espruino/blob/master/libs/js/update_js.sh) which helps to shave a few bytes and improve execution speed - but I'll dump that in when I update the ChangeLog.

When you install the 'Languages' app it overwrites this with a country-specific version, and interestingly in that version it's fixed, so I guess the change never got pulled back into the default Espruino one: https://github.com/espruino/BangleApps/blob/master/apps/locale/locale.html#L214

jessegonyou:
Understood, thanks!

PR #2715, opened 2026-07-03
Document E.getBattery when CUSTOM_GETBATTERY is defined

yngv27:
Boards can wire a custom battery reader via makefile DEFINES without adding board headers to core; include the API in generated docs too.


gfwilliams:
thanks!

SimonGAndrews:
Hi @yngv27 great to see practical work in the esp32 idf4 espruino port.  I will make sure we take account of these in the testing of IDF 5 port also. Keep up the good work.

PR #2716, opened 2026-07-03
Add ESP32.lightSleep and fix USB-JTAG console for light sleep

yngv27:
ESP32-C3 boards using USB Serial/JTAG for the console cannot light-sleep while the UART0 driver is installed; skip UART0 init when USJ is enabled and tear down/reinstall the USB-JTAG driver around light sleep. Pause ConsoleTask during sleep to avoid faults from reads on gated pads.

Also fix deepSleepExt0/Ext1 wakeup pin validation and GPIO masks on ESP32-C3, and advance jsiLastIdleTime after light sleep so timers do not fire in a tight loop.

## Summary
- Add `ESP32.lightSleep(ms)` — timer wakeup, RAM/JS state preserved; fixes `jsiLastIdleTime` after sleep so `setInterval`/`setTimeout` don't storm
- USB Serial/JTAG console: skip UART0 driver when `ESPR_USE_USB_SERIAL_JTAG` (UART0 blocks light sleep on ESP32-C3); tear down/reinstall USJ driver around sleep; pause `ConsoleTask` to avoid faults on gated pads
- Fix `deepSleepExt0`/`deepSleepExt1` wakeup pin validation and GPIO bitmask on ESP32-C3 (`esp_sleep_is_valid_wakeup_gpio`, `pinToESP32Pin`)
## Files
- `targets/esp32/jswrap_esp32.c` / `.h`
- `targets/esp32/jshardwareUart.c`
## Test plan
- [ ] ESP32-C3 board with `ESPR_USE_USB_SERIAL_JTAG` (e.g. Xteink X4): `ESP32.lightSleep(2000)` returns without watchdog reset; `var x=1; ESP32.lightSleep(2000); x` still `1`
- [ ] `setInterval` callback count sane after light sleep (no tight re-fire loop)
- [ ] `ESP32.deepSleepExt0(D3, 0)` accepts valid C3 GPIO on power button pin
- [ ] Classic UART-console ESP32 board still builds and serial works

PR #2717, opened 2026-07-03
ESP32: improve ADC reliability and support JSHPINSTATE_ADC_IN

yngv27:
Configure ADC channel attenuation once per channel (reconfigure only when range changes) instead of on every readADC call, and remove debug printf from rangeADC. Add JSHPINSTATE_ADC_IN handling in jshPinSetState so board definitions can use pinstate ADC_IN for analog inputs without triggering "Unexpected state" errors.

## Summary

- Add `JSHPINSTATE_ADC_IN` handling in `jshPinSetState()` so board definitions can use `pinstate: ADC_IN` without hitting "Unexpected state" errors (GPIO input, floating pull).
- Improve ESP32 ADC reliability by configuring channel attenuation once per channel via `adcEnsureChannelConfigured()`, reconfiguring only when `rangeADC()` changes attenuation, instead of calling `adc1_config_channel_atten()` on every `readADC()`.
- Remove debug `printf` calls left in `rangeADC()`.

These changes are generic ESP32 improvements (tested in the context of ESP32-C3 resistor-ladder buttons and battery ADC). No board-specific code.

## Test plan

- [ ] Build `BOARD=ESP32C3_IDF4 RELEASE=1` (or any ESP32 IDF4 board with ADC)
- [ ] Flash and connect via Web IDE (USB serial)
- [ ] `analogRead(D1)` / `E.getAnalogVRef()` — values stable across repeated reads
- [ ] Board with `pinstate: ADC_IN` in `.py` (e.g. resistor-ladder buttons) — no "Unexpected state" on boot
- [ ] `ESP32.setAnalog()` / changing attenuation — readings update correctly after range change
- [ ] Classic ESP32 (non-C3) — ADC still works (no regression)

## Related

- Enables reliable ADC for boards using analog buttons / battery monitoring
- Complements #2715 (CUSTOM_GETBATTERY) and #2716 (ESP32 lightSleep)

### DIFF
--- a/targets/esp32/jshardware.c
+++ b/targets/esp32/jshardware.c
@@ -326,6 +326,10 @@ void jshPinSetState(
     pull_mode= GPIO_PULLUP_ONLY;
     if (negated) jsError( "jshPinSetState: can't do Open Drain on negated pin");
     break;
+  case JSHPINSTATE_ADC_IN:
+    mode = GPIO_MODE_INPUT;
+    pull_mode = GPIO_FLOATING;
+    break;
   default:
     jsError( "jshPinSetState: Unexpected state: %d", state);
   return;
--- a/targets/esp32/jshardwareAnalog.c
+++ b/targets/esp32/jshardwareAnalog.c
@@ -31,6 +31,17 @@
 #define adc_channel_max 8
 
 adc_atten_t adc_channel[8];
+static bool adc_channel_setup[adc_channel_max];
+static adc_atten_t adc_channel_configured_atten[adc_channel_max];
+
+static void adcEnsureChannelConfigured(adc1_channel_t channel, int idx) {
+  if (idx < 0) return;
+  if (!adc_channel_setup[idx] || adc_channel_configured_atten[idx] != adc_channel[idx]) {
+    adc1_config_channel_atten(channel, adc_channel[idx]);
+    adc_channel_setup[idx] = true;
+    adc_channel_configured_atten[idx] = adc_channel[idx];
+  }
+}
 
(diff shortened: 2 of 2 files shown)


gfwilliams:
Thanks! Out of interest, was there any problem caused by calling `adc1_config_channel_atten` so often?

I see a heading `Test Plan` on your PRs. Is the idea that you have already done these tests yourself? Or that you need them doing before you feel confident merging?

yngv27:
full disclosure.. edits were generated by Cursor. My knowledge of the ESP32 is not this deep. If you would prefer that i do NOT use AI for these updates, LMK and I'll stop. These just came out of my exploration into getting a working Espruino on the Xteink X4 (ESP32C3) and it seemed some of these changes looked helpful to the main branch. 
As for the test plan, it's just a sanity check on both our sides. Something for me to test locally where appropriate, and notes for you.

gfwilliams:
A guess when asking about `adc1_config_channel_atten` I'm wondering whether it was changed for a specific reason? It says at the head of the bug it's "ADC reliability" - but I just wondered what the issue was before...

I'm totally fine with AI submissions as long as they're properly tested and are fixing stuff :) Regardless of who did it, if there's a bunch of random refactoring which doesn't actually do anything useful I'll probably push back, but if there's a real problem that's getting fixed in a sensible way I'm very happy for PRs and I don't care how the code was made :)

yngv27:
Looks like it was a possible cause of flaky readings on the C3 specifically (I was having issues with readings on a resistor-ladder). The IDF usage pattern says it must be set before the first read, not every read.

gfwilliams:
Ok - thanks!

Issue #565, opened 2015-08-21
I2C/SPI slave


gfwilliams:
JS execution speed is really too slow to properly do something via an IRQ, but there could be an array of data that can be read out...

```
write X, read N bytes -> get Array[X .. X+N-1]
write X,Y,Z -> Array[X]=Y, Array[X+1]=Z
```

http://forum.espruino.com/conversations/273112/

gfwilliams:
@MaBecker: I2C Slave support has just been added to nRF52 builds with 77a60d935908c8837465a59cbb905407ec2e47b4

Use as follows:

Add this to the `BOARD.py` `makefile` section:

```
'DEFINES += -DI2C_SLAVE -DTWIS_ENABLED=1 -DTWIS1_ENABLED=1' # enable I2C slave support
```

Then on the device:

```
I2C1.setup({scl:D22,sda:D23,addr:20});
I2C1.on('read',x=>print("read",x));
I2C1.on('write',x=>print("write",x));

print(I2C1.buffer)
```

So there's a 64 byte Uint8Array called `buffer` added to I2C1 when `addr:xyz` is supplied. Then you get events when a read or write has happened, and you can read/write what is in the array (which will have already changed after the write has happened).

On the I2C master it's like every other I2C device:

```
var i = new I2C();
i.setup({scl:D14,sda:D15});
function writeData(addr, data) {
  i.writeTo(20,addr,data);
}
function readData(addr) {
  i.writeTo(20,addr);
  return i.readFrom(20,addr);
}
```

MaBecker:
This what has been tested with a  Pixel badge 2018 and a BBB as master.

```js
I2C1.setup({scl:D10,sda:D11,addr:20});
I2C1.buffer[0]=255;
I2C1.on('read',x=>print("read",x));
I2C1.on('write',x=>print("write",x));
//I2C1.on('write',()=>g.drawString(I2C1.buffer[0].toString(),0,0,true).flip());

print(I2C1.buffer);
```
* [x] i2cdetect finds 0x14 which is 20 decimal - ok
* [x] i2cset and i2cget tested - ok
* [x] i2cdump -y -r 0-15 1 0x14 - ok
* [ ] upload code for a second time cause internal error, after E.reboot() and reconnect it works again one time
```js
>Uncaught InternalError: I2C Initialisation Error 8
 at line 14 col 37
I2C1.setup({scl:D10,sda:D11,addr:20});
``` 
[...]

PR #2719, opened 2026-07-09
nRF5x: add support for using second I2C peripheral

ssievert42:
This allows using the second I2C peripheral on nRF52840.

While there's `I2C1` and `I2C2`, only `I2C1` actually did something; calling `I2C2.setup()` just returned early without an error message.
I noticed this on a Seeed Xiao BLE, but this should apply to at least JoltJS as well.
(I wanted to use the accelerometer built into the XiaoBLE and connect an I2C display on another pair of pins. Software I2C was too slow and switching `I2C1` between pins didn't seem like a good idea.)

Should work but I'm not sure about the way I did the I2C device specific stuff in `jshI2CSetup()`.
Seems sort of silly (and error prone) to have that switch statement in three different places, but then again I didn't want to change the  signature of `jshGetTWI()` since it's used in other places as well and the other two switch statements are probably sufficiently separated by a `#if TWI_ENABLED` block.

Or should `I2C2` actually work and I've just missed something? 🙈

btw: My Bangle.js 2 still works nicely (well, except for the barometer / temperature sensor that refuses to talk to me) and I'm really looking forward to the Bangle.js 3 which will probably be an insta-buy from me :)

### DIFF
--- a/targets/nrf5x/app_config.h
+++ b/targets/nrf5x/app_config.h
@@ -74,8 +74,12 @@
 
 #if ESPR_I2C_COUNT>0
 #define TWI_ENABLED 1
+#define TWI0_ENABLED 1
+#define TWI0_USE_EASY_DMA 0
+#if ESPR_I2C_COUNT>1
 #define TWI1_ENABLED 1
 #define TWI1_USE_EASY_DMA 0
+#endif
 #else
 #define TWI_ENABLED 0
 #endif // ESPR_I2C_COUNT
--- a/targets/nrf5x/jshardware.c
+++ b/targets/nrf5x/jshardware.c
@@ -398,8 +398,12 @@ void spi0EvtHandler(nrf_drv_spi_evt_t const * p_event
 }
 #endif
 #if TWI_ENABLED
+static const nrf_drv_twi_t TWI0 = NRF_DRV_TWI_INSTANCE(0);
+bool twi0Initialised = false;
+#if ESPR_I2C_COUNT > 1
 static const nrf_drv_twi_t TWI1 = NRF_DRV_TWI_INSTANCE(1);
 bool twi1Initialised = false;
+#endif
 #endif
 #ifdef I2C_SLAVE
 static const nrf_drv_twis_t TWIS1 = NRF_DRV_TWIS_INSTANCE(1);
@@ -625,7 +629,10 @@ void spiFlashSleep() {
 
 #if TWI_ENABLED
 const nrf_drv_twi_t *jshGetTWI(IOEventFlags device) {
-  if (device == EV_I2C1) return &TWI1;
+  if (device == EV_I2C1) return &TWI0;
+#if ESPR_I2C_COUNT > 1
+  if (device == EV_I2C2) return &TWI1;
+#endif
   return 0;
 }
 #endif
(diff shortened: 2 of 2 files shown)


gfwilliams:
Nice! Yes, this looks good. However just looking at this bit in your changes:

https://github.com/espruino/Espruino/pull/2719/changes/548b8d8ebbc0fa7effafdae5cb3aaef4464e1901#diff-73d23e6a2aa4de3ac698501e353d86e49c1cdfeaeafb5b025eca1e0dde2727fcR2316-R2347

Is it possible to get rid of the second `switch (device) {` and put `pinFuncDevice |= JSH_I2Cx` in the first one?

And please could you update `getTWIS`? I think it should work ok with 2 as well, but it'll be confusing having I2C1 in slave move overlapping I2C2.

ssievert42:
Thanks for taking a look at this!

> Is it possible to get rid of the second `switch (device) {`

I'd like to, but the first switch statement is inside a
```c
#if TWI_ENABLED
{
}
#endif
```
block and `twi0Initialised` only exists if `TWI_ENABLED` is defined.

I guess we could do something like this:
```c
JshPinFunction pinFuncDevice = 0;
#if TWI_ENABLED
{
  switch (device) {
    case EV_I2C1:
      twiInitialised = &twi0Initialised;
      pinFuncDevice |= JSH_I2C1;
      break;
#if ESPR_I2C_COUNT > 1
    case EV_I2C2:
      twiInitialised = &twi1Initialised;
      pinFuncDevice |= JSH_I2C2;
      break;
#endif
  }
}
#endif
```
where `pinFuncDevice` would just stay 0 if I2C is disabled.
That would reduce the number of duplicated switch statements, but probably break I2C slave mode.

> And please could you update `getTWIS`?

Can do; having I2C1 be different hardware depending on whether you're using it in master or slave mode definitely seems confusing :+1:

How would I go about testing I2C slave support though?
I found #565 and 77a60d935908c8837465a59cbb905407ec2e47b4 but adding
`'DEFINES += -DI2C_SLAVE -DTWIS_ENABLED=1 -DTWIS0_ENABLED=1 -DTWIS1_ENABLED=1' # enable I2C slave support`
to the board file makes the build fail, probably because there's no `$(NRF5X_SDK_PATH)/components/drivers_nrf/twis_slave/nrf_drv_twis.c` in v15 of the SDK.

gfwilliams:
Surely if `TWI_ENABLED` isn't set, I2C doesn't work at all (TWI is Nordic's name for the I2C peripheral) - so I'd just include all the code you can inside the TWI_ENABLED ifdef?

Argh, what a pain about the I2C slave. You can see it defined and commented out in `boards/NRF52832DK.py` (the code you used) but it was developed for SDK12 and NRF52832DK so I guess I never modified the SDK15 makefile to handle it. If you had an nRF52832 device that'd be easiest, but it might just be a case of adding `targetlibs/nrf5x_15/integration/nrfx/legacy/nrf_drv_twis.h`

ssievert42:
Alright, finally got around to giving this another go :)

I've now shuffled stuff around as you suggested (I hope) and rebased against master.  
But I2S slave mode can still only be used from I2S1 / use TWIS0, mainly because I don't have enough time to do the needed changes properly right now. (Way too busy with Uni stuff)

> If you had an nRF52832 device that'd be easiest

It just so happens that I've got a Puck.js sitting on my desk :)  
I2S slave mode still works on nRF52832 with these changes if there's the following in the board file
```python
'DEFINES += -DI2C_SLAVE -DTWIS_ENABLED=1 -DTWIS0_ENABLED=1', # enable I2C slave support
```
I sort of gave up on doing an nRF52840 build with I2S slave mode though.

Are there docs about I2C slave mode that would need updating?  
Apart from the nRF52832DK board file (which I currently still left as is) and #565, I couldn't find any mentions of I2C slave mode.

Oh, and this seems to be a tad too much for the Micro Bit 1 build (this adds 12 bytes of code)  
No idea what to do about that though :sweat_smile:

gfwilliams:
Thanks! This looks great! It's hard to see how it could have added much at all to the build size - I'll have to see if I can pull some code out somewhere else.

I made a quick change thinking I could save some code space but only got us down to 8 bytes on the micro:bit!

gfwilliams:
> Are there docs about I2C slave mode that would need updating?

No - but I just pushed a small addition that might help people find it a bit better

Issue #2723, opened 2026-07-20
Questions around adding support for STM32 F7


rikkuness:
Hey Gordon :)

I've got my hands on an STM32 F767 NUCLEO board and I've started looking into adding support for this MCU to Espruino, but rather than go down a path and then land a huge PR on you I thought it best to get some thoughts up front if possible?

From my initial investigation I'd looked to add the F7 the same way the F2/3/4 with all the HAL/CMSIS code in `targetlibs/` with code from https://github.com/STMicroelectronics/cmsis-device-f7 but it was looking like a pretty significant chunk of overhead.

I then had a look at the way the `STM_LL` stuff was being done for the L series boards that are in here, and for the F7 it seemed like it was mostly API complete to do the same thing here and just use the low level stuff.

Which would you prefer, and are there any gotchas that I may run into going down the LL path rather than HAL/CMSIS?

Initially my goal would just be to get across the board support for all the typical features, but I did then later down the road wanted to play around with the Chrom-ART accelerator in this MCU and if I can get that Graphics class in Espruino to leverage that for optimised 2D rendering.

gfwilliams:
Hi! The LL port was from ST themselves in order to target the L series boards - they literally just copy-pasted the whole project and changed it to work with LL.

So that wasn't great from the point of view of maintainability and unfortunately they have since diverged a lot, but the idea behind it (using ST's LL HAL code which I think works across a bunch of boards) was good, and I think that makes the most sense for targeting something like the F7.

So yes, I'd use the LL port and just ifdef it for F7 where needed... but I guess it might be worth running over the commits to targets/stm32/jshardware.c and seeing if any of them make sense to apply to the LL port too.

In an ideal world, maybe we'd get rid of stm32_ll and instead have `jshardware_ll.c` and `jshardware_legacy.c` in  `targets/stm32`, with `jshardware.c` containing all the common code - but I reckon that's quite a big bunch of work!

rikkuness:
Oh I see, thanks for the insight, I hadn't realised that was the origins of that port.

Alright, I'll pursue the LL stuff for now then and when I have something tested as working I'll draft a PR. Maybe then I'll take a look into what it would take to merge those two into one target with two branches.

I'll close this for now and get it off your backlog and I'll get a draft over once I've done some testing, thank you :)

PR #2724, opened 2026-07-20
Add support for STM32F7 (specifically NUCLEO-F767ZI)

rikkuness:
As discussed in #2723 

Add support for STM32F7 family using LL modules. Firmware retrieved from https://github.com/stmicroelectronics/Stm32CubeF7 and build files loosely based around STM32L4 family.

I've added all the CMSIS/HAL stuff for `targetlibs` in it's own commit to hopefully make it easier to review changes.

Changes that _might_ have wider impacts but as best I can tell, hopefully not:
- Increased the range of the pinutils.py UART/SPI devices, this thing has _a lot_ of interfaces
- Likewise for jspin.h I've added those extra UART/SPI device numbers
- A bunch of guards around functions in jshardware.c for the L4 series, it still all builds but I don't think I have an L4 to test against

In current state it builds, flashes to the device and I've tested:
- [x] Connect via UART (virtual com port via ST-LINK)
- [x] On board LEDs light
- [x] Test code running from RAM
- [x] Tested flash via `require("Storage")` can read/write/getStats

Still to implement/untested:

- [x] ADC and DAC
- [x] SPI
- [x] I2C (Tested but see note below about clocking issues)
- [x] USB support via the on board OTG FS & HS

Stretch goals:

~- [ ] ChromArt Accelerator use with Graphics maybe?~
~- [ ] The NUCLEO-F767 has an ethernet port, maybe could be used with the NET modules here but would need a drive writing/borrowing from elsewhere~

Edit: Stretch goals will be addressed in future


rikkuness:
Had some issues testing i2c, but I noticed that `jshI2CSetup` has a timing set as per https://github.com/espruino/Espruino/blob/c58b979200712fc8b5c164d5b42c2edc05bd5647/targets/stm32_ll/jshardware.c#L1767

but then uses SYSCLK as the clock source, which seemed to be a mismatch, at least for the F7:
https://github.com/espruino/Espruino/blob/c58b979200712fc8b5c164d5b42c2edc05bd5647/targets/stm32_ll/jshardware.c#L385-L390

If I change the clock source to be `LL_RCC_I2C1_CLKSOURCE_HSI` etc. to use HSI then the scaler in the initial setup correctly clocks i2c at around 100kHz (in my test rig 89kHz due to clock stretching) but it worked, where it wasn't working before.

This also flagged another point, which is that these devices are currently ignoring the `bitrate` flag in `jshI2CSetup` anyway, which it feels like it'd be nice to fix.

Either way not sure what the preferred approach is, I can push the clock change where I changed from using SYSCLK to using HSI to get 100kHz, which works in my setup but I've no way right now of testing on an L4 (I do have an L0). Or is it better to make a separate change to change this whole STM_LL platform to support bitrate and pick an appropriate clock source, then work this into it after?

rikkuness:
I've added the I2C clock correction as mentioned above, I _believe_ it shouldn't change anything for L4 but it's worth bearing in mind in review.

USB is now also in and tested, testing was using the on NUCLEO USB connection (CN13) also worth noting this connection does not supply power to the board, so testing was conducted with a USB micro in each end, one for the CDC and one providing power via the ST-LINK.

ADC/DAC were also added and tested just by setting output levels on the DAC and testing it with a scope, and for the ADC providing voltages from a bench power supply into those pins, all appears to be working correctly.

gfwilliams:
Thanks! Sorry for the delay on this - I was away most of last week.

Looks good - I won't assume that I2C ever worked 100% right on the LL port. It was probably just enough to communicate with whatever the devs had connected. The LL_RCC_I2C1_CLKSOURCE_HSI change sounds good? I don't see it in the diff yet though?

Obviously as you say using bitrate would be better though and might be a nice addition - but maybe that's something for another PR to minimise this one? Same for ethernet/graphics too really.

So do you want to wait until you've checked SPI? Ignoring clock I tend to juct connect MOSI/MISO pins and see if I get back what I write.

rikkuness:
> Thanks! Sorry for the delay on this - I was away most of last week.

There's absolutely no rush at all, none of this is urgent, just felt worthy of contributing back :)

> The LL_RCC_I2C1_CLKSOURCE_HSI change sounds good? I don't see it in the diff yet though?

I was confused what happened here but it looks like I'd `git stash`'d it and for some reason assumed I'd commit it 😄 added it back in https://github.com/espruino/Espruino/pull/2724/changes/14f535b853b6d37317c744661e03a5e99a3a2bba

> Obviously as you say using bitrate would be better though and might be a nice addition - but maybe that's something for another PR to minimise this one? Same for ethernet/graphics too really.

Yeah 100% I've held off on all that stuff for now so I'll follow up with those in other PRs as and when.

> So do you want to wait until you've checked SPI? Ignoring clock I tend to juct connect MOSI/MISO pins and see if I get back what I write.

I'll do some quick tests this evening, at the least I can scope out those pins and make sure it looks like it's sending good data and have _some_ degree of confidence that it's working.

rikkuness:
I've just checked SPI, it was functional but did flag up another bug where subsequent calls to `SPI.setup` to change the baud wasn't taking effect as it wasn't disabling the device before trying to reconfigure in `jshSPISetup`, so I just pushed a small commit to correct that https://github.com/espruino/Espruino/pull/2724/changes/1f961579a26b472dc5496dac3e73e0d8841a2ace

Other than that I'd say SPI works, I can see good MOSI and CLK at least

gfwilliams:
Great! Ok, let's merge then!

PR #2728, opened 2026-07-30
Add support for Raspberry Pi Pico's 1 & 2 (RP2040, RP2350)

rikkuness:
Pulling in @SimonGAndrews 's RP2040 port and moving it to a general RP2XXX target family to support all current RP2 chips (RP2040, RP2350).

Current board definitions add Raspberry Pi Pico 1 & 2 (I have just ordered all other variants Pico 1 W, Pico 2 W etc. so will follow up with testing all of those once they arrive in the mail)

Changed the `jsvFastPrefixEqual` to support unaligned reads on Cortex M0(+). I've tested resulting assembled code against existing M0 boards and this doesn't appear to change any of the code paths in `jsvFindChildFromString` for existing boards.

`provision.sh` updated to support pulling Pico SDK 2.1.1


rikkuness:
@gfwilliams are you able to also add @SimonGAndrews as a reviewer on this one so he can tests against his test harness and give his sign off? 🙏🏻

gfwilliams:
Annoyingly I don't seem to be able to do it because he's not yet got access on this repo, but I'm happy this isn't going to break anything on existing builds and @SimonGAndrews seems happy, so I'll merge now.

If there are issues for RP2040 we can just push those changes in - but a IMO a slightly broken build is better than no build at all :)

rikkuness:
Awesome!

I'm sure some things will come up but I can follow up with those once my additional boards arrive and I've done some more testing.

Thanks!

SimonGAndrews:
If its not too late was thinking perhaps we should be merging this lot in a branch.  And test it there.
  Im pretty sure there are changes to core files as well as the new target files.  We dont want to break Espruino (Obviously(

gfwilliams:
Thanks! It's currently in `master` - I'm pretty confident that nothing you did will break existing builds though. Far less so than most stuff I've been doing recently!

SimonGAndrews:
Yes I saw that, ok great, of course nothing to stop us creating a branch for testing any time if we propose a big core change.  All good , exciting times.

PR #2738, opened 2026-09-01
ESP32: support undefined pin state during cleanup

SimonGAndrews:
### Issue

Espruino uses `JSHPINSTATE_UNDEFINED` when a peripheral releases its pins.

The ESP32 implementation of `jshPinSetState()` did not handle that state.
Cleanup operations such as `Serial.unsetup()` therefore fell through to the
default case and reported:

`jshPinSetState: Unexpected state: 0`

The message was caused by a missing valid cleanup-state mapping, not by an
invalid JavaScript API request.

### Fix

Map `JSHPINSTATE_UNDEFINED` to ESP-IDF `GPIO_MODE_DISABLE`.

This releases the pin without selecting an input or output function and avoids
treating normal peripheral cleanup as an error.

The correction adds one three-line switch case.

### Validation

`BOARD=ESP32_IDF5 RELEASE=1`

and tested on an ESP32 DevKitC V4:

- raw `Serial.setup()`/`Serial.unsetup()` output contained no
  `jshPinSetState: Unexpected state: 0` diagnostic;
- `Serial.isConnected()` returned `false`, `true`, then `false` before setup,
  after setup and after `unsetup()` respectively;
- UART data transfer, reconfiguration, listener and cleanup tests completed
  without the previous pin-state diagnostic.

Current ESP32 master emitted the diagnostic during the same cleanup path.

`BOARD=ESP32 RELEASE=1` configuration. `Serial.unsetup()` completed without
the diagnostic, providing compatibility evidence for the existing classic
ESP32 build.

`1bac1b958d41f35d9400863513ed8da37a6996a3`.

### DIFF
--- a/targets/esp32/jshardware.c
+++ b/targets/esp32/jshardware.c
@@ -344,6 +344,9 @@ void jshPinSetState(
   gpio_pull_mode_t pull_mode=GPIO_FLOATING;
   bool negated = pinInfo[pin].port & JSH_PIN_NEGATED;
   switch(state) {
+  case JSHPINSTATE_UNDEFINED:
+    mode = GPIO_MODE_DISABLE;
+    break;
   case JSHPINSTATE_GPIO_OUT:
     mode = GPIO_MODE_INPUT_OUTPUT;
     break;


gfwilliams:
Looks great - thanks!

PR #2749, opened 2026-09-13
ESP32: wait for GATT service teardown before replacement

### LINKED ISSUE #2748: ESP32C3_IDF4: repeated NRF.setServices() calls can assert and reboot
## Problem

Identified during post-IDF5-merge BLE regression testing:

On an ESP32-C3 IDF4 build, repeatedly replacing a GATT service using NRF.setServices() can cause a JavaScript variable-lock assertion followed by a RISC-V store/AMO access fault and reboot.

The first NRF.setServices() call returns to JavaScript successfully, but the failure occurs shortly afterwards during asynchronous ESP-IDF GATT processing.   Observed on:
- Board: ESP32C3_IDF4
- Espruino: 2v29.392
- Source commit: d8322cec9
- ESP-IDF: 4.4.8

The failure reproduced on both attempts. The same test completed three 50-iteration runs on ESP32C3_IDF5 from the same Espruino source without an assertion or reboot.

Expected behavior is that each service replacement completes safely and the final service remains available, without warnings, assertions or a reboot.

## Minimal reproduction
No BLE peer, network connection or external wiring is required. Use a wired Espruino console, reset the board and upload the following:
``` JS
// Repeatedly replace one small GATT service while disconnected.
// A healthy run prints DONE=PASS after 50 calls.

(function () {
  var iteration = 0;
  var iterationLimit = 50;

  function servicesFor(value) {
    var services = {};
    services[0xFFF0] = {};
    services[0xFFF0][0xFFF1] = {
      value: "VALUE_" + value,
      readable: true
    };
    return services;
  }

  function next() {
    iteration++;
    print("ITERATION=" + iteration);

    NRF.setServices(servicesFor(iteration));

    print("RETURNED=" + iteration);

    if (iteration < iterationLimit) {
      setTimeout(next, 500);
    } else {
      print("PASS iterations=" + iteration);
      print("DONE=PASS");
    }
  }

  echo(false);
  print("TEST=esp32_gatt_setservices_lifecycle");
  print("TARGET=" + process.env.BOARD);
  print("VERSION=" + process.version);
  print("GIT_COMMIT=" + process.env.GIT_COMMIT);

  NRF.disconnect();
  setTimeout(next, 500);
}());
```
Typical failing output is:
```JS
TEST=esp32_gatt_setservices_lifecycle
TARGET=ESP32C3_IDF4
ITERATION=1
RETURNED=1
assert failed: jsvUnLockInline jsvar.c:876 (jsvGetLocks(var)>0)

MCAUSE : 0x00000007
MTVAL  : 0x00000000

Rebooting...
```
MCAUSE=7 is a RISC-V store/AMO access fault. The presence of RETURNED=1 shows that the synchronous JavaScript call returned before the later failure.

## Source indication
ESP-IDF GATT service stop, deletion and application unregistration are callback-driven operations. The failure is consistent with Espruino replacing or releasing its shared service state before that asynchronous teardown has completed.
Different callback timing between ESP-IDF 4 and ESP-IDF 5 would explain why the defect reproduces reliably on the tested IDF4 build but has not reproduced on the corresponding IDF5 build.

## Validation required for a correction
A correction should demonstrate:
- repeated completion of all 50 calls on ESP32C3_IDF4, without an assertion, access fault or reboot;
- the same test remaining clean on ESP32C3_IDF5;
- no BT_GATT: Active Service Found or teardown-timeout warnings;
- after the 50 replacements, a second board can discover the final service, read its final characteristic value, write to it and disconnect normally;
- relevant ESP32-family builds using the shared GATT implementation still compile successfully.
Duplicate peripheral connect and disconnect JavaScript events are a previously identified, separate ESP32 issue and are outside the scope of this defect.

## Next Steps
A Candidate PR has been successfully tested and will be raised.

SimonGAndrews:
## Summary

Fixes #2748.

Repeated calls to `NRF.setServices()` could replace and release Espruino's
shared GATT service state while the ESP-IDF stop, delete and application
unregister sequence was still completing asynchronously.

On `ESP32C3_IDF4`, this reproducibly caused a `jsvUnLockInline` assertion,
followed by a RISC-V store/AMO access fault and reboot after the first
`NRF.setServices()` call had returned.

This change:

- waits for the final GATT application-unregister callback before freeing and
  rebuilding the service arrays;
- retains the JavaScript service definition while ESP-IDF is still using the
  existing registrations;
- checks GATT interface lookups before using them as array indexes;
- checks the actual result returned by `esp_ble_gatts_stop_service()`;
- falls back to application unregistration when service stopping fails;
- uses a five-second bounded wait, reporting a warning and retaining the
  existing state if teardown does not complete.

The correction is in the shared ESP32 GATT implementation and is not guarded
to IDF4. The source ordering is common, while the observed IDF4/IDF5 difference
is consistent with different callback timing.

## Validation of the fix
Issue #2748 presents a minimal Espruino JS REPL  to reproduce the issue. 

Current source base: `26fff60f9`  
PR head: `7bcdeeed5`

### Test method

The lifecycle test script is  the minimal JavaScript REPL reproducer.

It installs one GATT service containing one readable characteristic, then
calls `NRF.setServices()` 50 times at 500 ms intervals. The characteristic
value is changed on every call.

One test execution therefore represents 50 successive service replacements.

A successful execution requires:

- all 50 calls to `NRF.setServices()` to return;
- `PASS iterations=50` and `DONE=PASS` to be printed;
- no Espruino assertion;
- no RISC-V access fault, panic or reboot; and
- no ESP-IDF GATT lifecycle warning.

No BLE peer or external wiring is required for this test.

### Baseline reproduction

| Unpatched build | Complete script executions | Result |
|---|---:|---|
| `ESP32C3_IDF4`, current-master source `d8322cec9`, ESP-IDF 4.4.8 | 0 of 2 | Both executions returned from the first `NRF.setServices()` call and then failed before iteration 2 with a `jsvUnLockInline` assertion followed by a RISC-V store/AMO access fault |
| `ESP32C3_IDF5`, current-master source `d8322cec9`, ESP-IDF 5.5.3 | 3 of 3 | All 150 service replacements completed without a warning, assertion or reboot |

This establishes a reproducible failure on the C3 IDF4 build. The failure was
not reproduced on C3 IDF5, but the affected GATT lifecycle implementation is
shared between the ESP32 targets.

### Patched lifecycle test

| Patched build | Complete script executions | Successful `NRF.setServices()` calls | Result |
|---|---:|---:|---|
| `ESP32C3_IDF4`, ESP-IDF 4.4.8 | 4 of 4 | 200 of 200 | Pass; final two executions used exact PR head `7bcdeeed5` |
| `ESP32C3_IDF5`, ESP-IDF 5.5.3 | 2 of 2 | 100 of 100 | Pass; tested before rebase with the identical source correction |

The first IDF4 and IDF5 tests were performed before the branch was rebased, to include the recent 70 KB heap change. The lifecycle correction itself was unchanged by the rebase, 

Across the completed patched executions:

- every `NRF.setServices()` call returned;
- no `jsvUnLockInline` assertion occurred;
- no RISC-V access fault occurred;
- no GATT lifecycle warning was reported; and
- no panic or reboot occurred.

### Over-the-air functional validation

The C3 was also tested as a BLE GATT peripheral after 50 timed service
replacements. A classic ESP32 acted as the controlled BLE central.

The final service used a run-specific value so that the central could prove
that it had discovered the service created by the final replacement rather
than stale state from an earlier iteration.

| C3 peripheral build | Result |
|---|---|
| Patched `ESP32C3_IDF4`, ESP-IDF 4.4.8 | Final service discovered; expected characteristic value read; two characteristics written successfully; both devices disconnected cleanly |
| Patched `ESP32C3_IDF5`, ESP-IDF 5.5.3 | The same replacement, discovery, read, write and disconnect checks passed |

The strict over-the-air validation confirmed:

- exactly 50 service replacements;
- discovery of service `0xFFF0`;
- discovery and read of characteristic `0xFFF1`;
- receipt of the exact final run-specific value;
- successful writes to characteristics `0xFFF2` and `0xFFF3`; and
- disconnected state on both devices after completion.

This demonstrates that the correction does more than prevent the crash: the
GATT service remaining after repeated replacement is operational over the air.

### Build validation

The exact PR head `7bcdeeed5` was clean-built for:

| Board definition | ESP-IDF version | Result |
|---|---|---|
| `ESP32C3_IDF4` | 4.4.8 | Build completed successfully |
| `ESP32C3_IDF5` | 5.5.3 | Build completed successfully |

Only existing compiler warnings were observed.

Ive manual reviewed the actual code (CODEX produced) and seems to make sense, but i admit to not fully understanding the changes with regard to the IDF Bluetooth functionality,  but im confident on the testing,  @gfwilliams  if you want to review and change then please feel free ( including the Additional warnings have been introduced in the change.) 

## Conclusion
The patch prevents the reproducible C3 IDF4 service-lifecycle failure and
preserves the working C3 IDF5 behaviour. Repeated API calls and subsequent
over-the-air access to the final service both pass.

The correction is intentionally applied to the shared ESP32 GATT
implementation rather than guarded for IDF4. The unsafe asynchronous teardown
ordering exists in the shared code; the different observed IDF4 and IDF5
outcomes are consistent with differences in callback timing.
```

### CODE REVIEW

**Comment by gfwilliams** [comment]
Thanks for looking into this!

Do you think it might make more sense to just use the semaphone in gatts_reset after each call to esp_ble_gatts_stop_service so we don't exit that function until all services are removed?

I can imagine there might be some circumstances where the current fix doesn't quite work:

* There's >1 service, in which case `xSemaphoreGive(gatts_reset_complete);` is called after the first service is unregistered, but the second one ends up being removed in the background
* (maybe) `NRF.updateServices` gets called after a reset - this uses the `gatts_service` list, which would be in the middle of being freed after a reset with `gatts_reset`, but we're not waiting for the semephore.

But it should be pretty easy to just move the xSemaphoreTake lines down into `gatts_reset`

**Comment by SimonGAndrews** [comment]
> But it should be pretty easy to just move the xSemaphoreTake lines down into `gatts_reset`

Thanks @gfwilliams ,  Ill try that and test it.

**Comment by SimonGAndrews** [comment]
Thanks Gordon. We have moved the bounded semaphore wait into `gatts_reset()`.

Here is the Codex feedback on your direction and resulting change:

One detail is that the code starts teardown for every active service and then
waits once for the final application-unregister callback, rather than waiting
after each individual `esp_ble_gatts_stop_service()` call.

`gatts_unreg_app()` checks all registered services and only signals the
semaphore after every service has been unregistered. Therefore, successful
return from `gatts_reset()` means that all the old service arrays have been
released and it is safe for the caller to continue.

The revised implementation also:

- makes `gatts_reset()` return success or failure;
- prevents `gatts_set_services()` from rebuilding the service state if
  semaphore allocation fails or teardown times out; and
- prevents `gatts_update_service()` from accessing the service arrays if a
  reset remains in progress, for example following a timeout.

A multi-service validation test specifically for the concerns raised
in the review was added.  It installs two custom services while Espruino's BLE UART
service remains enabled, so three registered services are replaced. After
each replacement it calls `NRF.updateServices()` for characteristics in both
custom services.

Validation results:

| Hardware and build | Original lifecycle test | Multi-service replacement/update test |
|---|---:|---:|
| Classic ESP32, `BOARD=ESP32` | 50 replacements completed | 50 replacements and 50 updates completed |
| Classic ESP32, `BOARD=ESP32_IDF4` | 50 replacements completed | 50 replacements and 50 updates completed |
| Classic ESP32, `BOARD=ESP32_IDF5` | 50 replacements completed | 50 replacements and 50 updates completed |
| ESP32-C3, `BOARD=ESP32C3_IDF4` | 50 replacements completed | 50 replacements and 50 updates completed |
| ESP32-C3, `BOARD=ESP32C3_IDF5` | 50 replacements completed | 50 replacements and 50 updates completed |

All five firmware configurations also completed clean builds. No GATT teardown
timeout, active-service warning, Espruino assertion, processor access fault,
panic or reboot was observed.

*****

What do you think?  Does this cover your suggested fix ?

**Comment by SimonGAndrews** [comment]
Sorry @gfwilliams its a bit tricky to see the resulting final change across the two commits in the this one PR,  if there is a better way to do this eg submit one new PR let me know. 

### Resulting code in the PR
The final combined implementation:
- starts teardown for every active GATT service;
- waits inside gatts_reset() for the callback confirming that all services are unregistered;
- retains the old JavaScript service definition until teardown finishes;
- frees the old arrays only after the final unregister callback;
- returns success or failure from gatts_reset();
- prevents NRF.setServices() from rebuilding the arrays following a timeout or semaphore-allocation failure;
- prevents NRF.updateServices() from accessing arrays while teardown remains incomplete;
- validates GATT interface indexes before using them;
- checks the actual result from esp_ble_gatts_stop_service();
- falls back to application unregistration when service stopping fails.
In GitHub:

File: targets/esp32/BLE/esp32_gatts_func.c [hunk 1]
```diff
@@ -697,25 +720,60 @@ void gatts_set_services(JsVar *data){
...
+        r = esp_ble_gatts_stop_service(gatts_service[i].service_handle);
+        if (r) {
+          jsWarn("stop service error:%d\n",r);
+          r = esp_ble_gatts_app_unregister(gatts_service[i].gatts_if);
+          if (r) jsWarn("app_unregister after stop error:%d\n", r);
+        }
         // ESP_GATTS_STOP_EVT should now be fired, and we do esp_ble_gatts_delete_service in there
       }
     }
+    if (xSemaphoreTake(gatts_reset_complete, pdMS_TO_TICKS(GATTS_RESET_TIMEOUT_MS)) != pdTRUE) {
```
gfwilliams: Do we want to put this (and `xSemaphoreTake`) inside the FOR loop? As I understand it, each separate esp_ble_gatts_stop_service will fire its own ESP_GATTS_STOP_EVT, so if we do it here we're going to complete when only the first gatts_service is handled, not when they're all done?

gfwilliams:
I just made a quick change - hopefully it'll be ok, but please could you try and re-run your tests when you get a moment?

We now wait for each service in turn to be removed. And because we do it on the `for` we no longer have to check for activeServices and return, because if none are active we won't wait.

SimonGAndrews:
Thanks Gordon,  Ive checked this with the REPL test (follows) but Im getting a failure.  Using:
``` JS
// Repeatedly replace one small GATT service while no peer is connected.
// No external wiring or second BLE device is required. Use a wired console.
// A healthy run prints the completion marker after 50 returned API calls.
(function () {
  var iteration = 0;
  var iterationLimit = 50;

  function servicesFor(value) {
    var services = {};
    services[0xFFF0] = {};
    services[0xFFF0][0xFFF1] = {
      value : "VALUE_" + value,
      readable : true
    };
    return services;
  }

  function next() {
    iteration++;
    print("ITERATION=" + iteration);
    NRF.setServices(servicesFor(iteration));
    print("RETURNED=" + iteration);

    if (iteration < iterationLimit) {
      setTimeout(next, 500);
    } else {
      print("PASS iterations=" + iteration);
      print("DONE" + "=PASS");
    }
  }

  echo(false);
  print("TEST=esp32_gatt_setservices_lifecycle");
  print("TARGET=" + process.env.BOARD);
  print("VERSION=" + process.version);
  print("GIT_COMMIT=" + process.env.GIT_COMMIT);
  NRF.disconnect();
  setTimeout(next, 500);
}());
```
and it is failing with 
```
>
TEST=esp32_gatt_setservices_lifecycle
TARGET=ESP32_IDF5
VERSION=2v29.397
GIT_COMMIT=74409df84
ITERATION=1
RETURNED=1
ITERATION=2
Prompt not detected - upload failed. Trying to recover...
WARNING: Timed out waiting for GATT services to reset
RETURNED=2
=undefined
ITERATION=3
WARNING: Timed out waiting for GATT services to reset
RETURNED=3
```

Codex feedback is 
- Gordon’s loop stops one service and immediately waits for the semaphore.
- gatts_unreg_app() currently gives that semaphore only after it finds that all registered services have been unregistered.
- The remaining services have not yet been stopped because the loop is waiting.
- The first wait therefore times out.


If i can provide further Info , just shout,

gfwilliams:
Argh, sorry. That was me misunderstanding the patch then - I didn't realise that the semaphore was only called back when all services were gone.

I've just reverted my commit and merged :)

targets/esp32/BLE/esp32_gatts_func.c
gfwilliams: No: the semaphore is only completed when all services are gone

PR #2752, opened 2026-09-16
ESP32: connect after station start only when requested

### LINKED ISSUE #2750: ESP32_IDF5: Wifi.scan() returns an empty access-point list when starting station mode
## Summary

On the classic ESP32 `BOARD=ESP32_IDF5` build, `Wifi.scan()` returns an empty
access-point array when it has to start the Wi-Fi station interface.

This occurs even when multiple 2.4 GHz access points are visible. A separately
controlled access point was not returned by the scan, although connecting
directly to that same access point subsequently passed association, DHCP and
UDP communication.

The issue reproduces on current Espruino master `26fff60f9`, using ESP-IDF
5.5.3.

## Simple reproduction

Run this on an ESP32 within range of at least one known 2.4 GHz access point:

```javascript
echo(false);

var wifi = require("Wifi");

print("TEST=esp32_wifi_scan_empty");
print("BOARD=" + process.env.BOARD);
print("VERSION=" + process.version);
print("GIT_COMMIT=" + process.env.GIT_COMMIT);

wifi.disconnect();
wifi.stopAP();

setTimeout(function () {
  wifi.scan(function (accessPoints) {
    print("SCAN_COUNT=" + accessPoints.length);
    print("SCAN_RESULTS=" + JSON.stringify(accessPoints));
    print("DONE=esp32_wifi_scan_empty");
  });
}, 500);
```
On BOARD=ESP32_IDF5 the result is:
```JS
TEST=esp32_wifi_scan_empty
BOARD=ESP32_IDF5
VERSION=2v29.394
GIT_COMMIT=26fff60f9
SCAN_COUNT=0
SCAN_RESULTS=[]
DONE=esp32_wifi_scan_empty
```
## Controlled reproduction

A separate ESP32-C3 provided a WPA2 access point on channel 6. This allowed its availability and station activity to be observed independently.  The empty result was reproduced with:
Tested source | AP settling time | Scan result
-- | -- | --
IDF5 development image 20302d7f9 | 1.5 seconds | zero access points
IDF5 development image 20302d7f9 | 5 seconds | zero access points
Exact current master 26fff60f9 | 5 seconds | zero access points

Bypassing Wifi.scan() and connecting directly to the same access point passed:association; DHCP address acquisition; UDP request and response; and clean disconnect. This rules out the access point, credentials, channel and basic radio operation.

SimonGAndrews:
## Summary

Fixes #2750.

`Wifi.scan()` and `Wifi.connect()` both use the ESP32 station interface, but
the shared `WIFI_EVENT_STA_START` handler currently calls
`esp_wifi_connect()` unconditionally.

When `Wifi.scan()` starts a stopped station, that event therefore starts an
unwanted connection attempt while the scan is running. ESP-IDF can abort the
scan in this situation, causing Espruino to return an empty access-point list.

This change separates starting the station interface from requesting a
connection.

## Changes

The change adds two Boolean variables to Espruino's ESP32 wrapper: one tracks
the station-interface lifecycle and the other tracks connection intent.

- `g_isStaStarted` records whether ESP-IDF has reported the station interface
  as started. It is set by `WIFI_EVENT_STA_START` and cleared by
  `WIFI_EVENT_STA_STOP`.
- `g_connectAfterStaStart` records that an explicit `Wifi.connect()` request is
  waiting for a stopped station to start. The station-start handler consumes
  and clears this one-shot request before calling `esp_wifi_connect()`.

When `Wifi.scan()` starts a stopped station, it does not set
`g_connectAfterStaStart`. The resulting `WIFI_EVENT_STA_START` therefore
records that the station has started without initiating a connection.

When `Wifi.connect()` starts a stopped station, it sets
`g_connectAfterStaStart` before calling `esp_wifi_start()`. The subsequent
station-start event then completes the requested connection.

If the station is already running, `Wifi.connect()` calls
`esp_wifi_connect()` directly because ESP-IDF will not emit another
station-start event.

`Wifi.disconnect()` and a failed `esp_wifi_start()` clear
`g_connectAfterStaStart`, preventing an old connection request from affecting
a later operation.

The existing idle-radio shutdown and power-saving behaviour is retained.

## Validation

Clean release builds passed for:

- `ESP32`
- `ESP32_IDF4`
- `ESP32_IDF5`
- `ESP32C3_IDF4`
- `ESP32C3_IDF5`
- `ESP32S3_IDF4`
- `ESP32S3_IDF5`

A Linux host build also passed, and host tests matched the upstream baseline.

Runtime validation included:

- Classic `ESP32_IDF5`, ESP-IDF 5.5.3:
  - exact clean master returned zero access points in two controlled runs;
  - the candidate returned 15 and 13 access points;
  - the expected controlled AP was found once in each run;
  - neither scan caused an unsolicited connection.
- Classic legacy `ESP32`:
  - 15 station-lifecycle checks passed;
  - four independently observed UDP exchanges passed.
- `ESP32C3_IDF4`:
  - the same 15 station-lifecycle checks passed;
  - four independently observed UDP exchanges passed.
- `ESP32S3_IDF5`:
  - focused clean-base test returned zero access points;
  - the candidate returned 10 access points and found the expected AP;
  - no unsolicited connection was observed.

The classic IDF5 lifecycle checks also covered:

- direct connection from a stopped station;
- scanning and reconnecting while connected;
- disconnect, idle shutdown and connection after scanning;
- AP+station transition;
- cancelled connection followed by repeated connect/disconnect cycles;
- saved-configuration restoration after hardware reboot;
- explicit recovery after controlled AP loss.

A separate S3 reconnect stall was reproduced during testing on both the clean base and the
candidate and is not part of this change.


MaBecker:
Hmm, let me check the difference again between idf3, idf4 and idf5 builds before we merge this.

ˋˋˋ
wifi=require(“Wifi“);
wifi.stopAP();
wifi. getStatus();
ˋˋˋ

And compare the modes (none, sta, ap, sta+ap).

Ps: here was a change that might explain the different behaviour: wifi start was disabled in this function.

https://github.com/espruino/Espruino/blob/1c96229d08efd3df6894d21e01435fe09270842c/libs/network/esp32/jswrap_esp32_network.c#L608-L610

SimonGAndrews:
Good catch. I agree that the change you identified in `b4e8a4b5a`, which removed the early
AP-mode start from `esp32_wifi_init()`, helps explain why the problem became
easier to reproduce. 

_This change seems to be part of @gfwilliams change to stop the radio until needed to save power._

Before that change, the IDF5 port started Wi-Fi in AP mode during
initialization. The driver would  therefore already be running when a scan was
requested. After the change, an initial scan from the stopped state has to call
`esp_wifi_start()`, which emits `WIFI_EVENT_STA_START` and exposes the shared
handler’s unconditional `esp_wifi_connect()` call.  

However, the unconditional connection from STA_START predates the IDF5 port; it was introduced in 2018. The IDF5 port later added a separate behaviour in April 2026: starting Wi-Fi in AP mode during initialization. Earlier legacy builds did not start Wi-Fi this way. Re-enabling it would therefore mask the stopped-state case rather than correct the shared scan/connect lifecycle and effect the power saving changes. 

Can you try the new code “state-controlled version”  in https://github.com/espruino/Espruino/issues/2750 that creates the issue , if you get a chance please.

gfwilliams:
This looks good to me - thanks! It makes sense that my change broke this - as @SimonGAndrews says it was to save power when WiFi wasn't needed. I guess the problem was there before (eg if you did stopAP and then scan?) it's just that because the AP was up by default it wasn't obvious.

@MaBecker let me know what you're done checking and we can merge.

MaBecker:
created a new pr #2755

gfwilliams:
Just chatted to @MaBecker - am merging this, and pulling in some changes from #2755 too

PR #2755, opened 2026-09-18
ESP32 wifi fixes

MaBecker:
new -DESPR_WIFI_DELAY_START to delay start
fix wifi scan issues, add hidden
add wifi reason UNINITIALIZED


libs/network/esp32/jswrap_esp32_network.c
gfwilliams: you could just move this line down below `esp_wifi_set_mode` and then you don't have to manually free it if it errors?

libs/network/esp32/jswrap_esp32_network.c
```diff
@@ -601,13 +601,17 @@ void esp32_wifi_init() {
...
   //ESP_ERROR_CHECK(esp_event_handler_instance_register(IP_EVENT, IP_EVENT_AP_STAIPASSIGNED, &wifi_event_handler, NULL, &instance_ap_ip));
+
 #else
   ESP_ERROR_CHECK(esp_wifi_set_storage(WIFI_STORAGE_RAM));
 #endif
-
   // Don't init wifi yet - wait until wifi.connect/startAP is called
   //ESP_ERROR_CHECK(esp_wifi_set_mode(WIFI_MODE_AP)); // WIFI_MODE_STA/WIFI_MODE_APSTA
   //ESP_ERROR_CHECK(esp_wifi_start());
+  #ifndef ESPR_ESP32_WIFI_DELAY_START
```
gfwilliams: I really don't like this. What's the reason for re-adding it? All this has done is fragment the code, so with  ESPR_ESP32_WIFI_DELAY_START set, setScan doesn't work first time, and without it wifi draws a bunch of power all the time, and if you stop it, setScan then doesn't work!

@SimonGAndrews PR (https://github.com/espruino/Espruino/pull/2752) actually fixed the underlying issue so setScan always worked. What did you think was wrong with it?

SimonGAndrews:
see testing feedback  https://github.com/espruino/Espruino/issues/2750#issuecomment-5727605079

MaBecker:
thanks for checking, just posted a fix for the type

